PR #22, opened 2020-09-25
readme with input/output binding tutorial

cachai2:


cachai2:
Got sql trigger binding working so will complete documentation surrounding that as well before making PR.

PR #23, opened 2020-09-28
update README and triggerREADME

cachai2:


README.md
```diff
@@ -1,12 +1,89 @@
...
-This repository contains extension code for SQL Server input and output bindings. Trigger bindings are coming soon.
+    ```sql
+    ALTER TABLE [dbo].[Employees] ADD CONSTRAINT PKey PRIMARY KEY CLUSTERED (EmployeeId);
+    ```
 
-# Input and Output Binding
-## Quick Start
+### Set Up Local .NET Function App ###
+
+**NOTE: THE MYGET PACKAGE DOES NOT CURRENTLY CONTAIN TRIGGER BINDING FUNCTIONALITY**
```
cgillum, line 84: This is no longer accurate, right?

README.md
```diff
@@ -28,25 +105,124 @@ This repository contains extension code for SQL Server input and output bindings
     dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --version 1.0.0-preview1
```
cgillum, line 105: Should we point people to 1.0.0-preview2?

README.md
cachai2: I had to modify .csproj in order to get the SQL trigger working. I tried it out again this morning and just using the myget package resulted in some namespaces not being found for the trigger. I've attached a screenshot below. Is the myget supposed to include full functionality for the trigger now? I might have missed something or taken a weird step in setup, so if someone can verify, that would be great. 

Capture

README.md
cachai2: What is the difference between preview1 and preview2? @anthonychu probably has a better idea of this.

cgillum: preview2 is the package I uploaded that had the trigger binding.

cachai2: Oh ok. That makes sense. I missed that I wasn't using preview2. I was still using preview1 as I was following the README and thought it was updated. I just tried it, and trigger works. That is definitely a lot simpler. Thanks for pointing this out Chris. I'm going to close the PR for now and update the docs to reflect using preview2.

README.md
cachai2: Fixed with preview2

cachai2:
Closing PR for now. Updating docs to reflect using preview2.

PR #24, opened 2020-09-29
Trigger read me

cachai2:
updated to reflect preview2


README.md
```diff
@@ -1,12 +1,87 @@
-SQL Server Extension for Azure Functions
-===
+# SQL Server Extension for Azure Functions #
+
+## Introduction ##
+
+This repository contains extension code for SQL Server input and output bindings as well as a getting started guide and samples of how to use them. The getting started guide details how to setup your environment and provides a basic tutorial. A high level explanation of each binding is provided below. More in depth descriptions of each binding are in their respective sample sections. The trigger binding is currently not complete. However, if you would like to try it out and learn more about its current state, feel free to do so in the triggerREADME.
```
cgillum, line 5: Why are we saying that the trigger binding is incomplete?

cgillum (commented):
A couple high-level questions/comments before I dive too deep into this:

* Why are we saying that the trigger binding is not yet complete? Are we intentionally trying to discourage folks from using it? I was expecting we'd try to promote it so we can get feedback.
* I worry about having two separate README files for trigger and non-triggers. I think it would be better to either have a single README or break it up in a way such that we have a general README, and then one for each binding type (input, output, and trigger), preferably in a `/docs` folder.

README.md
cachai2: Valid points. Thanks for bringing them up. Please let me know your thoughts on the below Chris.

- Anthony and I were hesitant on pushing it as complete while knowing that there needs to be dev work. There was also the uncertainty of whether or not going forward we would be showing both the input/output and trigger bindings for the private preview. The question now is are we comfortable calling it complete/do we need to inform people about the current state it's in? What are your thoughts on this? 

A big concern was making sure people aren't expecting something fully polished. I'm not entirely familiar with how the private previews work yet, so I was under the impression that we would still get feedback even with the notes. However, if we are looking to promote it more, I can change the tone. In terms of the specific wording of 'not complete,' I'm rereading it with fresh eyes now, and I agree it could at least be rephrased if not removed.

- Originally, it started as a single README. However, through discussions with Anthony, we decided to split it into two parts. One with the input/output bindings which seems to be almost in a state that could be shipped, and one as a sort of try this with a note that it is still being developed. Again, a large part of this was due to uncertainty about the future of the SQL bindings. Thinking back, I don't think we brought this up in the weekly sync. Although, I definitely should have for the additional perspective and will try my best to do so going forward. I do agree that it is a bit weird to have two README files though. It would probably be better to have either the single README or a /docs folder. @anthonychu thoughts?

anthonychu: I don't think we need to call out how ready we think the trigger is. I agree we want people to try it. I was thinking of breaking it up into 2 because it felt like it was going to be very long with them together. We can put the on one page if we have links up top to show that there are 2 sections and give the ability for someone to go directly to each one of them.

cachai2: Ok, that makes sense. I'll make some edits to not call out the state of the trigger and consolidate to one README with links.

cachai2: Updated to one README and changed call outs to trigger. @cgillum and @anthonychu thoughts?

README.md
```diff
@@ -1,14 +1,71 @@
...
+### SQL Setup ###
+
+This requires already having a SQL database. If you need to create a SQL database, please refer to [Create Azure SQL Database](#Create-Azure-SQL-Database) in the tutorials section.
+
+A primary key must be set in your SQL table before using the bindings. To do this, run the below SQL commands in the SQL query editor.
+
+1. Ensure there are no NULL values in the primary key column. The primary key will usually be an ID column.
+
+    ```sql
+    ALTER TABLE ['your table name'] alter column ['column to be primary key'] int NOT NULL
```
glennamanns, line 37: I don't know that Azure SQL (or whatever) supports tables without Primary Keys and/or null Primary Keys. IMO, it's unnecessary to include these instructions but I don't really have a strong opinion on it

glennamanns: (referring to both this instruction and the `Set primary key column` instruction)

README.md
```diff
@@ -1,14 +1,71 @@
...
+    ```
+
+1. Set primary key column.
+
+    ```sql
+    ALTER TABLE ['your table name'] ADD CONSTRAINT PKey PRIMARY KEY CLUSTERED (['column to be primary key']);
+    ```
+
+1. Change tracking must be enabled on the database to use the trigger. If you do not plan on using the trigger, you can skip this step. To enable change tracking on the database, run:
```
glennamanns, line 46: I would include an FYI that Change Tracking costs money :) (or a link to their overview docs)

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
 1. Install the extension.
 
     ```bash
-    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --version 1.0.0-preview1
+    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --version 1.0.0-preview2
```
glennamanns, line 85: once I upload it, this will be `1.0.0-preview3`

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+            public string Company { get; set; }
+            public string Team { get; set; }
+        }
+    }
+    ```
+
+- Navigate back to your HttpTrigger file. We can ignore the 'Run' warning for now.
+- Open the local.settings.json file, and in the brackets for "Values," verify there is a 'SqlConnectionString.' If not, add it.
+- Hit 'F5' to run your code. Both an HttpTrigger and SQLBinding will run.
+- Click the link that appears in your terminal. If you get a 'HTTP ERROR 500' message, replace the browser url after the last '/' with 1. This is the id of the employee to be outputted. The url should look like 'localhost:7071/api/employees/1'
```
glennamanns, line 179: lol, in the future we should include a sample that has a happy landing page and not a 500 error, but probably ok for now

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+        IEnumerable<Employee> employee)
+    {
+        return new OkObjectResult(employee);
+    }
+    ```
+
+    *In the above, "select from Employees where EmployeeId = @EmployeeId" is the SQL script run by the input binding. The CommandType on the line below specifies whether the first line is a query or a stored procedure. The line below that determines the parameters. On the next line, the ConnectionStringSetting specifies that the app setting that contains the SQL connection string used to connect to the database is "SqlConnectionString." For more information on this, see the [Input Binding Samples](#Input-Binding-Samples) section*
+
+- Add 'using System.Collections.Generic;' to the namespaces list at the top of the page.
+- Currently, there is an error for the Ienumerable. We'll fix this by creating an Employee class.
```
glennamanns, line 160: nit: `IEnumerable`

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+            public string FirstName { get; set; }
+            public string Company { get; set; }
+            public string Team { get; set; }
+        }
+    }
+    ```
+
+- Navigate back to your HttpTrigger file. We can ignore the 'Run' warning for now.
+- Open the local.settings.json file, and in the brackets for "Values," verify there is a 'SqlConnectionString.' If not, add it.
+- Hit 'F5' to run your code. Both an HttpTrigger and SQLBinding will run.
```
glennamanns, line 178: I would instead say something like, "This will start up the Functions Host with a local HTTP Trigger and SQL Input Binding"

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+        employee.Company = "Microsoft";
+        employee.Team = "Functions";
+        output[0] = employee;
+        employee = new Employee();
+        employee.EmployeeId = 2;
+        employee.FirstName = "Hi";
+        employee.LastName = "SQLupdate";
+        employee.Company = "Microsoft";
+        employee.Team = "Functions";
+        output[1] = employee;
```
glennamanns, line 214: nit:
```suggestion
         output = new Employee[]
            {
                new Employee
                {
                    EmployeeId = 1,
                    FirstName = "Hello",
                    LastName = "World",
                    Company = "Microsoft",
                    Team = "Functions"
                },
                new Employee
                {
                    EmployeeId = 12,
                    FirstName = "Hi",
                    LastName = "SQLupdate",
                    Company = "Microsoft",
                    Team = "Functions"
                },
            };
```

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+            }
+        }
+    }
     ```
 
-1. See samples below for information on how to use the binding.
+- Add the below to 'local.settings.json' in "Values"
+
+    ```json
+    "AzureWebJobsDashboard": "UseDevelopmentStorage=true"
```
glennamanns, line 266: do they need this for all of them, or just the trigger?


We should

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
 1. Install the extension.
 
     ```bash
-    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --version 1.0.0-preview1
+    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --version 1.0.0-preview2
+    ```
+
+1. Ensure you have Azure Storage Emulator running. For information on the Azure Storage Emulator, refer [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator#get-the-storage-emulator)
+
+1. Set the SqlConnectionString. Your connection string can be found in your SQL database resource by going to the left blade and clicking 'Connection strings'. Copy the Connection String to 'local.settings.json', so '"SqlConnectionString": [Connection String]' is in 'Values:'.
```
glennamanns, line 90: I would include what their final host.json needs to look like (like what you did for AzureWebJobsDashboard, except show the full host.json w/ both settings required)

README.md
```diff
@@ -25,28 +82,218 @@ This repository contains extension code for SQL Server input and output bindings
...
+    ```json
+    "AzureWebJobsDashboard": "UseDevelopmentStorage=true"
+    ```
+
+- (*Skip this step if you have not completed the output binding tutorial*) Open your output binding file and modify some of the values (e.g. Change Team from 'Functions' to 'Azure SQL'). This will update the row when the code is run.
+- Hit 'F5' to run your code. Click the link for the http trigger to the output binding. You should see the log in VS Code update and tell you which row changed and what the data in the row is now.
+- Update, insert, or delete additional rows in your SQL table using the SQL query editor while the function app is running and observe the log updates.
+- Congratulations! You have successfully created your first SQL trigger! Checkout [Trigger Samples](#Trigger-Samples) for more information on how to use the trigger and explore on your own!
+
+## Samples ##
```
glennamanns, line 274: nit: I would name this "More Samples" (since we already have some earlier, in the tutorial)

glennamanns (approved):
thanks for doing this, Cary! Just a handful of nits

README.md
cachai2: I believe when I was initially testing out the SQL bindings the input binding didn't require me to set the primary keys/null keys for the table. However, enabling change tracking required it. I could be misremembering though. I think we can leave it in at least for the Bug Bash tomorrow and get feedback. I can test to double check this next week.

README.md
cachai2: Good point. I didn't know this was actually the case. I'll update this

README.md
cachai2: Will update and test

README.md
cachai2: Haha, very fair. When I was originally putting this in, I wanted to include the parameters in the tutorial. Will see if there's a quick update to make a happy landing. If not, we can leave as is for tomorrow

README.md
cachai2: This is perfect. I remember thinking about how to word that better earlier. Thanks Glenna

README.md
cachai2: I mainly ran into this when setting up the trigger. I'll add this to the quickstart setup though

README.md
cachai2: That makes sense. I'll make the update

README.md
cachai2: Made the input binding tutorial not take a parameter. Still walks them through the input binding but provides a happy landing page. There are lots of examples with parameters still in 'More Samples'

README.md
cachai2: Let me know if the update is what you were thinking

glennamanns:
I read over the latest changes and they look good to me. I'll run through the setup/samples one more time tomorrow morning before the bug bash.

PR #25, opened 2020-10-06
Performance Improvements

yorek:
Using JSON instead of Row Constructors
Reduced active connection usage 
Using more realistic test table

### DIFF
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -18,6 +18,7 @@ using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Tokens;
 using MoreLinq;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 
 namespace Microsoft.Azure.WebJobs.Extensions.Sql
@@ -91,7 +92,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private async Task<bool> OldImplementation(IEnumerable<T> rows, SqlAttribute attribute, IConfiguration configuration)
         {
             Stopwatch st = new Stopwatch();
-            
+
             int batchsize = 1000;
 
             using (SqlConnection connection = SqlBindingUtilities.BuildConnection(attribute.ConnectionStringSetting, configuration))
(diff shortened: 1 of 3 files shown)


src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -379,8 +381,55 @@ public static TableInformation RetrieveTableInformation(SqlConnection sqlConnect
                     primaryKeyMatchingQuery += $" AND ExistingData.{primaryKey} = NewData.{primaryKey}";
                 }
 
+                // 2. Get Columns Definition
+                string columnDefinitionQuery = $@"
```
glennamanns, line 385: what benefits does grabbing the column types offer? More performance? or is it just required to use the JSON functionality?

glennamanns (approved):
thanks for making these changes, Davide! I will merge them into my branch and work on the additional changes necessary (refactoring, support different orders of columns, etc.)

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -319,44 +300,61 @@ public class TableInformation
...
-                if (!string.IsNullOrEmpty(schema))
-                {
-                    primaryKeyQuery += $"AND Col.Constraint_Schema = '{schema}'";
-                }
+                IEnumerable<string> columnNames = typeof(T).GetMembers().Where(f => f.MemberType == MemberTypes.Property || f.MemberType == MemberTypes.Field).Select(f => f.Name);
 
                 var primaryKeys = new List<string>();
+                var columnsDefinition = new List<string>();
+
+                // 1. Query SQL to get primary keys
```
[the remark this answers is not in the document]
yorek: 300 ms are needed just the first time, as connection needs to be established...that's what is taking so much time. After that, connection is pooled and time will be something like 2 msec.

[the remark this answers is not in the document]
yorek: Btw, query is fine, I just cleaned up it a bit

src/SqlBinding/SqlAsyncCollector.cs
yorek: It is required for using OPENJSON

yorek: https://docs.microsoft.com/en-us/sql/relational-databases/json/use-openjson-with-an-explicit-schema-sql-server?view=sql-server-ver15

PR #26, opened 2020-10-07
Output Binding: Switch to MERGE + JSON query model

glennamanns:
`SqlDataAdapter` requires a lot of client-side work. It has to pull the whole SQL table into memory in order to distinguish between INSERTs and UPDATEs. As a result, we've now switched to generating `MERGE` T-SQL queries. This allows us to offload the work to SQL, who can then optimize for upserts. Thanks @yorek for the suggestion & assistance!

Assumptions made:
1. Column Names in SQL are Case-Insensitive (cannot have 'Cost' and 'cost')
2. The set of C# class field/property names must be equivalent to the set of your SQL column names (case insensitive)
3. I cache the PKs and column names and types for 10 minutes. If a user somehow manages to change their primary keys or column types without needing to recompile (and thus restart) their Functions code+host, then they may experience issues for up to 10 minutes. 

TODOs/Open Questions:
1. Is assumption 2 above an acceptable restriction? Do we want to allow extra SQL/C# class columns/properties? 
2. If we cannot find the primary keys/column names, we fail the operation. Do we need a backoff period?

### DIFF
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -5,10 +5,18 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Data;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Caching;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Configuration;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Tokens;
+using MoreLinq;
 using Newtonsoft.Json;
 
 
@@ -20,10 +28,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private readonly IConfiguration _configuration;
         private readonly SqlAttribute _attribute;
         private readonly List<T> _rows;
-        // Maps from database name + table name to SqlCommandBuilders
-        private static ConcurrentDictionary<string, SqlCommandBuilder> _commandBuilders = new ConcurrentDictionary<string, SqlCommandBuilder>();
-        // Maps from database name + table name to a byte array which contains information about that table's schema
-        private static ConcurrentDictionary<string, byte[]> _schemas = new ConcurrentDictionary<string, byte[]>();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqlAsyncCollector<typeparamref name="T"/>"/> class.
@@ -74,32 +78,34 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         {
             if (_rows.Count != 0)
             {
-                string rows = JsonConvert.SerializeObject(_rows);
-                await UpsertRowsAsync(rows, _attribute, _configuration);
+                // string rows = JsonConvert.SerializeObject(_rows);
+                // todo: take a lock on _rows
+                await UpsertRowsAsync(_rows, _attribute, _configuration);
                 _rows.Clear();
             }
         }
 
+        // Maps from database name + table name to SqlCommandBuilders
+        private static ConcurrentDictionary<string, SqlCommandBuilder> _commandBuilders = new ConcurrentDictionary<string, SqlCommandBuilder>();
 
-        /// <summary>
-        /// Upserts the rows specified in "rows" to the table specified in "attribute"
-        /// If a primary key in "rows" already exists in the table, the row is interpreted as an update rather than an insert.
-        /// The column values associated with that primary key in the table are updated to have the values specified in "rows".
-        /// If a new primary key is encountered in "rows", the row is simply inserted into the table. 
-        /// </summary>
-        /// <param name="rows"> The rows to be upserted </param>
-        /// <param name="attribute"> Contains the name of the table to be modified and SQL connection information </param>
-        /// <param name="configuration"> Used to build up the connection </param>
-        private async Task UpsertRowsAsync(string rows, SqlAttribute attribute, IConfiguration configuration)
+        private async Task<bool> OldImplementation(IEnumerable<T> rows, SqlAttribute attribute, IConfiguration configuration)
         {
+            Stopwatch st = new Stopwatch();
+            
+            int batchsize = 1000;
+
             using (SqlConnection connection = SqlBindingUtilities.BuildConnection(attribute.ConnectionStringSetting, configuration))
             {
                 string tableName = SqlBindingUtilities.NormalizeTableName(attribute.CommandText);
 
                 DataSet dataSet = new DataSet();
-                DataTable newData = (DataTable)JsonConvert.DeserializeObject(rows, typeof(DataTable));
+
+                string stringRows = JsonConvert.SerializeObject(_rows);
+                DataTable newData = (DataTable)JsonConvert.DeserializeObject(stringRows, typeof(DataTable));
 
                 await connection.OpenAsync();
+
+                st.Start();
                 using (SqlTransaction transaction = connection.BeginTransaction(IsolationLevel.RepeatableRead))
                 {
                     SqlDataAdapter dataAdapter = new SqlDataAdapter(new SqlCommand($"SELECT * FROM {tableName};", connection, transaction));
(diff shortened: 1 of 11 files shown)


src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+                    string message = $"Encountered exception while retrieving columns definition for table '{tableName}.' Cannot generate upsert command without them.";
+                    throw new InvalidOperationException(message, ex);
+                }
+                finally
+                {
+                    await sqlConnection.CloseAsync();
+                }
+
+                // 2. Ensure the POCO fields match the SQL column names. If not, throw. (TODO: is this expected behavior?)
+                IEnumerable<string> columnNamesFromPOCO = typeof(T).GetMembers().Where(f => f.MemberType == MemberTypes.Property || f.MemberType == MemberTypes.Field).Select(f => f.Name);
```
cgillum, line 317: For properties, do we need to make sure they are settable?

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
-                        // discover the table schema
-                        SqlCommand insertCommand = builder.GetInsertCommand();
-                        insertCommand.Connection = connection;
-                        insertCommand.Transaction = transaction;
-                        dataAdapter.InsertCommand = insertCommand;
-                        SqlCommand updateCommand = builder.GetUpdateCommand();
-                        updateCommand.Connection = connection;
-                        updateCommand.Transaction = transaction;
-                        dataAdapter.UpdateCommand = updateCommand;
+                        while (rdr.Read())
```
cgillum, line 346: Consider using `ReadAsync()` here too.

src/SqlBinding/SqlBinding.csproj
```diff
@@ -21,6 +21,8 @@
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.*" />
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.*" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
+    <PackageReference Include="morelinq" Version="3.3.2" />
```
cgillum, line 24: Do we have a hard requirement on this nuget package? I'm a little wary of adding 3rd party libraries to our list of dependencies if we don't have a strong reason for it.

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+                    primaryKeyMatchingQuery += $" AND ExistingData.{primaryKey} = NewData.{primaryKey}";
+                }
+
+                // Generate the UPDATE part of the merge query (all columns that should be updated)
+                string columnMatchingQuery = string.Empty;
+                foreach (var column in columnNames)
+                {
+                    columnMatchingQuery += $" ExistingData.{column} = NewData.{column},";
+                }
+                columnMatchingQuery = columnMatchingQuery.TrimEnd(',');
```
cgillum, line 264: Consider using `StringBuilder` instead of relying on string concatenation here and above. That would be more efficient  compared to relying on string concatenation.

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+			                    when DATA_TYPE in ('decimal', 'numeric') then '(' + cast(NUMERIC_PRECISION as varchar(9)) + ',' + + cast(NUMERIC_SCALE as varchar(9)) + ')'
+			                    else ''
+		                    end as {ColumnDefinition}	
+                    from 
+	                    INFORMATION_SCHEMA.COLUMNS c
+                    where
+	                    c.TABLE_NAME = '{tableName}'         
+                    and
+                        c.TABLE_SCHEMA = {schema}";
+            }
```
cgillum, line 244: Have we done a scan of this code to check for possible SQL injection vulnerabilities? There is a lot of manual query building that is making me a bit nervous.

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
             using (SqlConnection connection = SqlBindingUtilities.BuildConnection(attribute.ConnectionStringSetting, configuration))
             {
-                string tableName = SqlBindingUtilities.NormalizeTableName(attribute.CommandText);
+                string fullDatabaseAndTableName = attribute.CommandText;
+                ObjectCache cachedTables = MemoryCache.Default;
+                TableInformation tableInfo = cachedTables[fullDatabaseAndTableName] as TableInformation;
+
+                if (tableInfo == null)
+                {
+                    Console.WriteLine("lookin");
```
cgillum, line 105: :)

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+                    tableInfo = await TableInformation.RetrieveTableInformationAsync(connection, fullDatabaseAndTableName);
+
+                    // If we were unable to look up the primary keys, for whatever reason, return early. 
+                    // We'll try again next time. TODO: do we need to have a max # of retries / backoff plan?
+                    if (tableInfo == null)
+                    {
+                        return;
```
cgillum, line 112: Would this look like a silent failure to the customer? Would an exception be a better experience?

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -74,13 +77,12 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
         {
             if (_rows.Count != 0)
             {
-                string rows = JsonConvert.SerializeObject(_rows);
-                await UpsertRowsAsync(rows, _attribute, _configuration);
+                // TODO: do we need to take a lock on _rows?
```
cgillum, line 80: Only if we want instances of `SqlAsyncCollector<T>` to be thread-safe, I think.

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+                }
+
+                if (!primaryKeys.Any())
+                {
+                    string message = $"Did not retrieve any primary keys for '{tableName}.' Cannot generate upsert command without them.";
+                    throw new InvalidOperationException(message);
+                }
+
+                // Match column names to POCO field/property objects
+                IEnumerable<MemberInfo> primaryKeyFields = typeof(T).GetMembers().Where(f => primaryKeys.Contains(f.Name, StringComparer.OrdinalIgnoreCase));
```
cgillum, line 370: We might want to consider logging the results of the schema lookup we do to make debugging easier if something doesn't behave the way we expect.

src/SqlBinding/SqlAsyncCollector.cs
glennamanns: I agree, the manual query generation is definitely not ideal. Table name and schema come from the function `Attribute,` which the Function writer determines, so I think that is trustworthy. However, the data portion is generated on the fly, and could potentially come from user input. I'll double check if `OPENJSON` is vulnerable to injection.

src/SqlBinding/SqlAsyncCollector.cs
glennamanns: We do throw exceptions in `RetrieveTableInformationAsync` if we run into any issues, so this line of code is probably dead code.

src/SqlBinding/SqlBinding.csproj
glennamanns: I use it to batch the `IEnumerable<T>`. This isn't a hard requirement and I could implement my own version of it if it's not worth the dependency

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -90,55 +92,284 @@ public async Task FlushAsync(CancellationToken cancellationToken = default)
...
+        /// Generates T-SQL for data to be upserted using Merge. 
+        /// This needs to be regenerated for every batch to upsert.
+        /// </summary>
+        /// <param name="table">Information about the table we will be upserting into</param>
+        /// <param name="rows">Rows to be upserted</param>
+        /// <returns>T-SQL containing data for merge</returns>
+        private static void GenerateDataQueryForMerge(TableInformation table, IEnumerable<T> rows, out string newDataQuery, out string rowData)
+        {
+            IList<T> rowsToUpsert = new List<T>();
+            HashSet<string> uniqueUpdatedPrimaryKeys = new HashSet<string>();
```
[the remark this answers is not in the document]
glennamanns: SQL server defaults to case-insensitive PKs, although they do allow case-sensitive ones if explicitly set. I'll assume the default (case-insensitive) for now, and if we get an ask later, we could make it configurable

src/SqlBinding/SqlAsyncCollector.cs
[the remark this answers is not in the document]
glennamanns: I agree, this implementation feels somewhat ridiculous to do manually, rather than some existing library. However, in my limited experience with EF, I haven't been very impressed by its performance. From what I understand, its version of merge (`AddOrUpdate`) requires two round-trips to the database, rather than just one. I'm not sure if there are better alternatives to EF (@yorek - do you know of any?)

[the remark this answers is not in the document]
yorek: I would suggest to use Dapper and something like Automapper (https://automapper.org/).

src/SqlBinding/SqlAsyncCollector.cs
yorek: Correct, instead of using string interpolation, parameters should be used. I recommend using Dapper again as it will simplify code a lot, but not sure if you have any rule regarding taking external dependencies.

yorek:
> 2. If we cannot find the primary keys/column names, we fail the operation. Do we need a backoff period?

Primary Key are definitely needed otherwise updating the correct row will be challenging. (Technical any UNIQUE index will be enough, bet let's start by assuming a PRIMARY KEY needs to be there. Let's see the feedback on this)

For what concern properties and columns: I will not raise any error if a property of the POCO object is not mappable to any columns: simply that property will not be persisted. This approach is ignoring additional properties is used also when deserializing JSON into an object and I think it is coherent to follow the same approach also for "serializing" and object into the database.

src/SqlBinding/SqlAsyncCollector.cs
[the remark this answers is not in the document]
cgillum: If we support case insensitivity, I'm wondering if we should initialize this as:
```suggestion
            HashSet<string> uniqueUpdatedPrimaryKeys = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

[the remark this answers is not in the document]
glennamanns: yes, sorry. I was agreeing with you that it should be a case-insensitive `HashSet` and I've made that change now

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -362,14 +375,52 @@ public async static Task<TableInformation> RetrieveTableInformationAsync(SqlConn
...
+                {
+                    string message = $"All primary keys for SQL table '{tableName}' and schema '{schema}' need to be found in '{typeof(T)}.' Missing primary keys: [{string.Join(",", missingFromPOCO)}]";
+                    throw new InvalidOperationException(message);
+                }
+
+                return new TableInformation(primaryKeyFields, columnDefinitionsFromSQL, GetMergeQuery(primaryKeys, columnDefinitionsFromSQL, fullName));
+            }
+        }
+
+        public class DynamicPOCOContractResolver : DefaultContractResolver
```
cgillum, line 396: Can this be `private` instead of `public`?

cgillum (approved):

PR #35, opened 2020-10-16
Lower-case all SQL and POCO fields

### LINKED ISSUE #34: Allow mismatch of casing between POCO and SQL primary keys
Today, we fail the UPSERT if the POCO Primary Key casing is different than that of the SQL column casing.

glennamanns:
To address the casing issue, I just convert both the POCO fields/properties and the SQL column names to lowercase, so they always match.

closes #34

### DIFF
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -333,7 +333,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     {
                         while (await rdr.ReadAsync())
                         {
-                            columnDefinitionsFromSQL.Add(rdr[ColumnName].ToString(), rdr[ColumnDefinition].ToString());
+                            columnDefinitionsFromSQL.Add(rdr[ColumnName].ToString().ToLowerInvariant(), rdr[ColumnDefinition].ToString());
                         }
                     }
                 }
@@ -358,7 +358,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     {
                         while (await rdr.ReadAsync())
                         {
-                            primaryKeys.Add(rdr[ColumnName].ToString());
+                            primaryKeys.Add(rdr[ColumnName].ToString().ToLowerInvariant());
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -358,7 +358,7 @@ public async static Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                     {
                         while (await rdr.ReadAsync())
                         {
-                            primaryKeys.Add(rdr[ColumnName].ToString());
+                            primaryKeys.Add(rdr[ColumnName].ToString().ToLowerInvariant());
```
cgillum, line 361: I don't quite understand the details of the bug, but by do we need to lowercase all the column names? Is there some comparison we're doing that doesn't allow us to use case-insensitivity?

glennamanns: Our C# comparisons are case-insensitive, but some of the ones SQL is doing are not. 

Say your C# object data serializes into this JSON parameter, based on the casing of the POCO object:
`"[{"productid":5, "cost":0}]"`

but our T-SQL query looks like this, based on the casing of the SQL columns:
`WHEN MATCHED THEN UPDATE SET  ExistingData.ProductID = NewData.ProductID, ExistingData.Cost = NewData.Cost 
WHEN NOT MATCHED THEN INSERT (ProductID,Cost) VALUES (ProductID,Cost);`

the upsert fails, as SQL cannot find "ProductID" and "Cost" in the serialized JSON parameter

glennamanns: Whereas if you have consistent casing in both the JSON parameter and the T-SQL merge command, SQL can properly match them, despite the potential mismatch between that casing ("productid") and the column name casing ("ProductID").

This may be a bug on their side, as the columns themselves are case-insensitive, but this was my way of getting around it for now

cgillum: I see. BTW, case-insensitivity in SQL actually depends on the collation of the database, so not all databases are case sensitive.

cgillum (approved):

PR #45, opened 2021-06-16
Set up Azure pipeline

zijchen:
Set up build pipeline in ADO
- PR validation runs build and test on Windows, Mac, and Linux
- Release pipeline currently scheduled to run at midnight UTC daily
- .NET Build task builds nuget package directly
- ESRP Code Signing for nupkg only
- Also fixed a bug in Linux build failing to find the test csproj due to case sensitive file path


zijchen:
I actually have no idea why this PR kicked off the pipeline... Earlier commits had a trigger so wondering if it's cached.

zijchen:
> I actually have no idea why this PR kicked off the pipeline... Earlier commits had a trigger so wondering if it's cached.

Nevermind I think I figured it out, had to disable those explicitly. https://github.com/Azure/azure-functions-sql-extension/pull/45/commits/a1ba3a78776957a366ed76737fe0126fefea49cb

kisantia:
should the pipeline also get run for PRs so that the tests get run for each PR?

zijchen:
> should the pipeline also get run for PRs so that the tests get run for each PR?

Yeah that's a good idea, although I don't think we need a full build for PR validation. I'm thinking we can have 2 pipelines:
- PR Validation: Build + Test
- Release: template from PR Validation + Publish + Code Sign

Thoughts?

zijchen:
Linux build was failing due to case sensitive path to the test csproj. Fixed it and verified that PR validation is hooked up and working :)

WebJobs.Extensions.Sql.sln
```diff
@@ -7,7 +7,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlExtensionSamples", "samp
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlBinding", "src\SqlBinding\SqlBinding.csproj", "{4453B407-2CA3-4011-BDE5-247FCE5C1974}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlExtension.Tests", "test\SQLExtension.tests\SqlExtension.Tests.csproj", "{A5B55530-71C8-4A9A-92AD-1D33A5E80FC1}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlExtension.Tests", "test\SQLExtension.tests\SQLExtension.tests.csproj", "{A5B55530-71C8-4A9A-92AD-1D33A5E80FC1}"
```
udeeshagautam, line 10: intentional change?

builds/azure-pipelines/build-release.yml
```diff
@@ -0,0 +1,40 @@
...
+stages:
+- stage: BuildPublish
+  displayName: 'Release Build and Publish'
+
+  jobs:
+  - job: BuildTestPublish
+    displayName: 'Release: Build, Test and Publish on Windows'
+
+    pool:
+      vmImage: 'windows-latest'
```
udeeshagautam, line 26: So am I understanding correctly that for build and test - we are running on all three plaforms and running tests on all three but for nuget publish we only need one since the end result does not dependend on build platform - right?

WebJobs.Extensions.Sql.sln
zijchen: Yes without this change Linux build fails (due to case sensitive file path). I can separate this out in its own PR if you want

builds/azure-pipelines/build-release.yml
zijchen: Correct :)

udeeshagautam: Just reading its trying to call the tests too right - then may be add the matrix here too...??

udeeshagautam (approved):
Looks good to me, just few clrifications/suggetsions. 

@glennamanns @fabiocav please let us know if you would be able to take a look or Zi can go ahead with checkin?

builds/azure-pipelines/build-release.yml
zijchen: I copied the template from the OpenAPI extension and they don't have it. My understanding is since we already have PR validation running on all 3 platforms, changes that are checked in are gated in a sense so we don't need to run all tests again in the nightly.

udeeshagautam: Hmm... I was thinking for our pielines they all run for dalily. I am okay starting wih just this and then moving to more as needed.

builds/azure-pipelines/build-release.yml
```diff
@@ -0,0 +1,40 @@
+trigger: none
+
+pr: none
+
+schedules:
+  - cron: "0 0 * * Mon-Fri"
```
glennamanns, line 6: consider doing Sun-Thurs since it's running at night & we will likely look at it the next day

WebJobs.Extensions.Sql.sln
```diff
@@ -7,7 +7,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlExtensionSamples", "samp
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlBinding", "src\SqlBinding\SqlBinding.csproj", "{4453B407-2CA3-4011-BDE5-247FCE5C1974}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlExtension.Tests", "test\SQLExtension.tests\SqlExtension.Tests.csproj", "{A5B55530-71C8-4A9A-92AD-1D33A5E80FC1}"
```
glennamanns, line 10: nit: rather than changing the csproj to match the folder name & ending up with weird casing, we should fix the folder name (SqlExtension.Tests)

glennamanns (approved):
Thanks, Zi! LGTM, just two minor nits

builds/azure-pipelines/build-release.yml
zijchen: Changed to run everyday, pipeline runs in under 5 minutes. We can change this to a lower frequency if it becomes too taxing in the future.

WebJobs.Extensions.Sql.sln
zijchen: Changed paths and files to all be "SqlExtension.Tests"

PR #59, opened 2021-09-17
Add code coverage to build pipeline

lucyzhang929:


builds/azure-pipelines/template-steps-publish.yml
```diff
@@ -40,4 +40,10 @@ steps:
...
     targetPath: '$(Build.ArtifactStagingDirectory)' 
-    artifactName: 'drop'
\ No newline at end of file
+    artifactName: 'drop'
+
+- task: PublishCodeCoverageResults@1
+  displayName: 'Publish Code Coverage Results'
+  inputs:
+    codeCoverageTool: Cobertura
+    summaryFileLocation: '$(Build.SourcesDirectory)\TestResults\coverage\coverage.cobertura.xml'
```
chlafreniere, line 49: Is there an advantage to using sources dir instead of artifact staging dir?

lucyzhang929: Sources directory was from an example I was following but I believe both would work the same. I can change it to artifact staging dir for consistency and run a build to double check.

chlafreniere (approved):

PR #62, opened 2021-09-21
Add test for GetTableAndSchema

lucyzhang929:
Made a small fix in GetTableAndSchema where the brackets would not be removed if the fullName was not in schema.table format.

### DIFF
--- a/src/SqlBinding/SqlBindingUtilities.cs
+++ b/src/SqlBinding/SqlBindingUtilities.cs
@@ -207,12 +207,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
         public static void GetTableAndSchema(string fullName, out string schema, out string tableName)
         {
-            // defaults
-            tableName = fullName;
-            schema = "SCHEMA_NAME()"; // default to user schema
-
             // remove [ ] from name if necessary
             string cleanName = fullName.Replace("]", string.Empty).Replace("[", string.Empty);
+
+            // defaults
+            tableName = cleanName;
+            schema = "SCHEMA_NAME()"; // default to user schema
 
             // if in format schema.table, split into two parts for query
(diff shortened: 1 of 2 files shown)


src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -207,13 +207,13 @@ public static string NormalizeTableName(string tableName)
 
         public static void GetTableAndSchema(string fullName, out string schema, out string tableName)
         {
-            // defaults
-            tableName = fullName;
-            schema = "SCHEMA_NAME()"; // default to user schema
-
             // remove [ ] from name if necessary
             string cleanName = fullName.Replace("]", string.Empty).Replace("[", string.Empty);
```
Charles-Gagnon, line 211: We should look into how this is used - table/schema names can have ['s and ]'s in them and so this might be incorrectly stripping those. If this is just supposed to be removing brackets from a quoted identifier then this logic would be incorrect.

lucyzhang929: The schema and tableName are used in these two query statements:
https://github.com/Azure/azure-functions-sql-extension/blob/b25cdd50e7e489f51e5590e43b7bf034f55ebe58/src/SqlBinding/SqlAsyncCollector.cs#L252-L254
https://github.com/Azure/azure-functions-sql-extension/blob/b25cdd50e7e489f51e5590e43b7bf034f55ebe58/src/SqlBinding/SqlAsyncCollector.cs#L275-L277
The tableName has quotes around it and the schema is returned with quotes as well
https://github.com/Azure/azure-functions-sql-extension/blob/147986998f2440f917ca67410c37a880a882f5a5/src/SqlBinding/SqlBindingUtilities.cs#L222
so I think the brackets are expected to be removed.

PR #65, opened 2021-09-21
Add integration tests

zijchen:
This PR adds integration testing based on the Azure Functions in the samples project.

To run the tests, you need to have these installed
- [Azure Functions CLI](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
- [SQL Server Express LocalDB](https://docs.microsoft.com/sql/database-engine/configure-windows/sql-server-express-localdb)

Most of the test infrastructure is in the IntegrationTestBase class. The constructor sets up a test database on LocalDb then starts the Functions Host via Functions CLI. Tests run from there.

TODO: Azure pipelines changes to install the new required tools (unsure if pre-installed), and testing

References
- [Guide for adding Functions integration tests](https://microsoft.github.io/AzureTipsAndTricks/blog/tip196.html)
- [Reference for Functions CLI](https://docs.microsoft.com/en-us/azure/azure-functions/functions-core-tools-reference?tabs=v2#func-start)


test/SqlExtension.IntegrationTests/TestUtils.cs
```diff
@@ -0,0 +1,126 @@
...
+                if (catchException == null || !catchException(ex))
+                {
+                    throw;
+                }
+            }
+            finally
+            {
+                if (cmd != null)
+                {
+                    cmd.Dispose();
```
Charles-Gagnon, line 120: Typically it's easier to put the command creation in a using block which will auto-dispose the command after it's done

test/SqlExtension.IntegrationTests/SqlExtension.IntegrationTests.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IsPackable>false</IsPackable>
```
Charles-Gagnon, line 5: I think it would be better to have all the tests - both unit and integration - in a single project. Splitting them up doesn't seem like something we should need to do and will make code coverage stuff harder.

Charles-Gagnon (commented):
There seems to be a lot of other changes in this PR - the samples changes for example. What are those for?

Also - all CS files should have the copyright header at the top

Charles-Gagnon:
That's fine for this time, but in the future yes please try to split stuff up into separate PRs if you can 😄

test/SqlExtension.IntegrationTests/SqlExtension.IntegrationTests.csproj
Charles-Gagnon: Didn't realize you were running the tests on the solution, so the code coverage stuff shouldn't be a problem in that case. But I still think having all the tests in one place makes it easier to work with. 

(although the tests are failing now so you'll need to still do some work there to get them running in the pipeline first. Were you planning on doing that yourself? Maybe we can do that in a follow up PR and just disable them for now in this PR)

zijchen: Moved IntegrationTests to inside the Tests project, disabled all but the Ubuntu pipeline for now

test/SqlExtension.IntegrationTests/TestUtils.cs
zijchen: Switched to a using statement

builds/azure-pipelines/build-pr.yml
```diff
@@ -22,10 +22,11 @@ stages:
       matrix:
         linux:
           imageName: 'ubuntu-latest'
-        mac:
-          imageName: 'macos-latest'
-        windows:
-          imageName: 'windows-latest'
+        # Disabling other platform builds temporarily until we can get integration tests working there
```
Charles-Gagnon, line 25: I would rather we do something else for the platforms that don't support integration tests currently such that the unit tests still run. I'd suggest looking into the filter option to dotnet test - adding a category seems like a decent idea (Categories being Unit/Integration)

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -9,6 +9,23 @@ steps:
   inputs:
     version: '3.1.x'
 
+- script: npm install -g azure-functions-core-tools@3 --unsafe-perm true
```
Charles-Gagnon, line 12: Why the unsafe perm param?

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -9,6 +9,23 @@ steps:
...
+- task: DockerInstaller@0
+  displayName: Docker Installer
+  inputs:
+    dockerVersion: 17.09.0-ce
+    releaseType: stable
+        
+- script: docker pull mcr.microsoft.com/mssql/server:2019-latest
+  displayName: Pull MSSQL Docker Image
+
+- script: 'docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=Placeholder1" -e "MSSQL_PID=Express"
```
Charles-Gagnon, line 24: We should generally be avoiding hardcoding passwords. I'd suggest having the pipeline randomly generate one and pass that in instead.

samples/SqlExtensionSamples/local.settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
Charles-Gagnon, line 1: We shouldn't be checking this in - especially for samples that people might be looking at. Something like this should probably be generated by the pipeline itself.

test/SqlExtension.Tests/IntegrationTestBase.cs
```diff
@@ -0,0 +1,252 @@
+﻿using Microsoft.Data.SqlClient;
```
Charles-Gagnon, line 1: All new files should have copyright header

test/SqlExtension.Tests/IntegrationTestBase.cs
```diff
@@ -0,0 +1,252 @@
...
+        /// </summary>
+        protected string DatabaseName { get; private set; }
+
+        /// <summary>
+        /// Output redirect for XUnit tests.
+        /// Please use TestOutput.WriteLine() instead of Console or Debug.
+        /// </summary>
+        protected ITestOutputHelper TestOutput { get; private set; }
+
+        protected int Port { get; private set; }
```
Charles-Gagnon, line 40: doc comment

test/SqlExtension.Tests/IntegrationTestBase.cs
```diff
@@ -0,0 +1,252 @@
...
+                ExecuteNonQuery(File.ReadAllText(file));
+            }
+
+            // Set SqlConnectionString env var for the Function to use
+            Environment.SetEnvironmentVariable("SqlConnectionString", connectionStringBuilder.ToString());
+        }
+
+        private void StartFunctionHost()
+        {
+            Port = 7071 + TestUtils.ThreadId;
```
Charles-Gagnon, line 105: Why do you need to increase this? Shouldn't there only be one function host running at a time?

test/SqlExtension.Tests/IntegrationTestBase.cs
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        public void Dispose()
+        {
+            Connection.Close();
+
+            try
+            {
+                // Drop the test database
+                using (SqlConnection masterConnection = new SqlConnection("Data Source=(LocalDb)\\MSSQLLocalDB;Initial Catalog=master;Integrated Security=True;Pooling=False;"))
```
Charles-Gagnon, line 233: This is still trying to connect to localdb

test/SqlExtension.Tests/SqlInputBindingIntegrationTests.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using SqlExtensionSamples;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace SqlExtension.Tests
```
Charles-Gagnon, line 10: It would be convenient to put all the tests into separate folders and under a different namespace based on their type (`SqlExtension.Tests.Integration` and `SqlExtension.Tests.Unit` for example)

samples/SqlExtensionSamples/OutputBindingSamples/QueueTriggerProducts.cs
```diff
@@ -5,22 +5,22 @@
 using Microsoft.Extensions.Logging;
 using static SqlExtensionSamples.ProductUtilities;
 
-namespace SqlExtensionSamples.OutputBindingSamples
+namespace SqlExtensionSamples
```
Charles-Gagnon, line 8: Why did you remove the namespace?

builds/azure-pipelines/build-pr.yml
zijchen: Added filter to exclude integration tests on Mac and Windows

builds/azure-pipelines/template-steps-build-test.yml
zijchen: Removed (copied from the Functions CLI docs, not sure why it's set there)

builds/azure-pipelines/template-steps-build-test.yml
zijchen: Added a step to generate password using the BuildId

test/SqlExtension.Tests/IntegrationTestBase.cs
zijchen: Removed. Left over code from when I attempted to multithread it

samples/SqlExtensionSamples/OutputBindingSamples/QueueTriggerProducts.cs
zijchen: Removed for consistency with the other sample functions, none of the other ones have it

samples/SqlExtensionSamples/local.settings.json
zijchen: Wouldn't we want to include this for people to look at? They will need these settings to run the sample functions (especially the SqlConnectionString which is specific to our samples)

samples/SqlExtensionSamples/OutputBindingSamples/AddProduct.cs
```diff
@@ -9,20 +9,19 @@
...
-            [Sql("Products", ConnectionStringSetting = "SqlConnectionString")] out Product product)
+            [Sql("dbo.Products", ConnectionStringSetting = "SqlConnectionString")] out Product product)
         {
             product = new Product
             {
-                Name = req.Query["name"],
-                ProductID = int.Parse(req.Query["id"]),
-                Cost = int.Parse(req.Query["cost"])
+                Name = req.Query["Name"],
+                ProductID = int.Parse(req.Query["ProductID"]),
```
chlafreniere, line 23: Why are these being changed? Isn't it easier to understand when the params passed in aren't exactly the same as the column names?

zijchen: Changed back

samples/SqlExtensionSamples/local.settings.json
Charles-Gagnon: I was basing it off of the guidance from the AF docs about never checking this in. But I guess we have the .gitignore set up properly so maybe it's not the worst idea...seems like other sample guides have it e.g. https://github.com/yokawasa/azure-functions-python-samples/tree/master/v2functions...

I guess as long as it's empty like this it's fine

builds/azure-pipelines/template-steps-build-test-ubuntu.yml
```diff
@@ -0,0 +1,69 @@
...
+- task: DockerInstaller@0
+  displayName: Docker Installer
+  inputs:
+    dockerVersion: 17.09.0-ce
+    releaseType: stable
+        
+- script: docker pull mcr.microsoft.com/mssql/server:2019-latest
+  displayName: Pull MSSQL Docker Image
+
+- bash: echo "##vso[task.setvariable variable=serverPassword]Test-$(Build.BuildNumber)"
```
Charles-Gagnon, line 25: Since we're doing this already let's at least add in a timestamp or something to randomize it a bit more

builds/azure-pipelines/build-pr.yml
```diff
@@ -14,18 +14,33 @@ stages:
   displayName: 'PR Validation'
 
   jobs:
+  - job: BuildTestOnUbuntu
```
Charles-Gagnon, line 17: Let's avoid referencing ubuntu specifically for this stuff - just linux is fine

samples/SqlExtensionSamples/Product.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace SqlExtensionSamples
```
Charles-Gagnon, line 1: All new files should have copyright header

Charles-Gagnon (approved):

chlafreniere:
🔥 🔥 🔥

PR #77, opened 2021-09-24
Update readme with known issues section

### LINKED ISSUE #72: ntext, text, image SQL Data Types Not Supported
When I tested these data types that [will be removed in the future](https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver15), I got the following error:

```
TEXT, NTEXT, SQL_VARIANT and IMAGE types cannot be used as column types in OPENJSON function with explicit schema.
```

We should document this in relnotes at a minimum.

chlafreniere:
Fixes #72.

Would love feedback on the string 😄.

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,7 +502,7 @@ public static IActionResult Run(
 ```
 ## Known Issues
 
-- Attempting to use output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` is not supported, as these types are deprecated and do not work with the `OPENJSON` function. Upserting data will fail in this case.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -499,6 +500,9 @@ public static IActionResult Run(
     return new CreatedResult($"/api/addproduct", product);
 }
 ```
+## Known Issues
+
+- Attempting to use output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` is not supported, as these types are deprecated and do not work with the `OPENJSON` function. Upserting data will fail in this case.
```
dzsquared, line 505: Output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` are not supported and data upserts will fail. These types [will be removed](https://docs.microsoft.com/sql/t-sql/data-types/ntext-text-and-image-transact-sql) in a future version of SQL Server and are not compatible with the `OPENJSON` function used by this Azure Functions binding.

dzsquared (approved):
proposed wording in review

README.md
chlafreniere: This is great. Thank you!

Charles-Gagnon (approved):

Issue #32, opened 2020-10-16
Error message for output binding doesn't explain the issue when table name is wrong


jwendl:
### Description
When pointing the binding to a table that isn't valid it gives an error message saying it can't find the primary key, when the real issue is that the binding wasn't able to connect to the table itself.

### Expected behavior
An error message that says something to the effect of "Wasn't able to find table x"

### Actual behavior
System.Private.CoreLib: Exception while executing function: TimerTriggerProducts. Microsoft.Azure.WebJobs.Host: Error while handling parameter products after function returned:. SqlBinding: Did not retrieve any primary keys for 'Products' in schema 'SCHEMA_NAME().' Cannot generate upsert command without them.

### Repo Steps
Create a table called [dbo].[Product] in the database and run the sample (instead of [dbo].[Products])

Charles-Gagnon:
Closing trigger binding related issues until we revisit support for that

Charles-Gagnon:
Look into creating the table if it doesn't exist

chlafreniere:
> Look into creating the table if it doesn't exist

I've thought about this a bit more, and I certainly think that this part of it is lower pri, but if we did do this, I'm concerned that it may be a slippery slope/feature creep (e.g. could this further be expanded where a db needs to be created as well?)

My hypothesis from my super limited experience is that the vast majority of the time, users will hit this when they've misspelled the table name and/or schema. In this case, they wouldn't want the table created to begin with.

@dzsquared happy to hear your thoughts as well.

PR #78, opened 2021-09-24
Check if table exists before getting table info

lucyzhang929:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/32

### DIFF
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -237,7 +237,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <summary>
             /// Generates SQL query that can be used to retrieve the Primary Keys of a table
             /// </summary>
-            public static string GetPrimaryKeysQuery(string schema, string tableName) 
+            public static string GetPrimaryKeysQuery(string quotedSchema, string quotedTableName) 
             {
                 return $@"
                     select 
@@ -249,15 +249,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     where
                         tc.CONSTRAINT_TYPE = 'PRIMARY KEY'
                     and
-                        tc.TABLE_NAME = {tableName}
+                        tc.TABLE_NAME = {quotedTableName}
                     and
-                        tc.TABLE_SCHEMA = {schema}";
+                        tc.TABLE_SCHEMA = {quotedSchema}";
             }
 
             /// <summary>
             /// Generates SQL query that can be used to retrieve column names & types of a table
             /// </summary>
-            public static string GetColumnDefinitionsQuery(string schema, string tableName)
+            public static string GetColumnDefinitionsQuery(string quotedSchema, string quotedTableName)
             {
                 return $@"
                     select	
@@ -272,9 +272,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     from 
 	                    INFORMATION_SCHEMA.COLUMNS c
                     where
-	                    c.TABLE_NAME = {tableName}
+	                    c.TABLE_NAME = {quotedTableName}
                     and
-                        c.TABLE_SCHEMA = {schema}";
+                        c.TABLE_SCHEMA = {quotedSchema}";
             }
 
             /// <summary>
@@ -321,14 +321,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public async static Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName)
             {
-                SqlBindingUtilities.GetTableAndSchema(fullName, out string schema, out string tableName);
+                SqlBindingUtilities.GetTableAndSchema(fullName, out string quotedSchema, out string quotedTableName);
 
                 // 1. Get all column names and types
                 var columnDefinitionsFromSQL = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
                 try
                 {
                     await sqlConnection.OpenAsync();
-                    SqlCommand cmdColDef = new SqlCommand(GetColumnDefinitionsQuery(schema, tableName), sqlConnection);
+                    SqlCommand cmdColDef = new SqlCommand(GetColumnDefinitionsQuery(quotedSchema, quotedTableName), sqlConnection);
                     using (SqlDataReader rdr = await cmdColDef.ExecuteReaderAsync())
                     {
                         while (await rdr.ReadAsync())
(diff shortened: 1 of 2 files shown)


src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -340,15 +380,15 @@ public async static Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                 catch (Exception ex)
                 {
                     // Throw a custom error so that it's easier to decipher.
-                    string message = $"Encountered exception while retrieving column names and types for table '{tableName}' in schema '{schema}.' Cannot generate upsert command without them.";
+                    string message = $"Encountered exception while retrieving column names and types for table '{tableName}' in schema {schema}. Cannot generate upsert command without them.";
```
lucyzhang929, line 383: Removing the extra quotes around schema since GetTableAndSchema returns schema with quotes around it.

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -234,6 +234,18 @@ public TableInformation(IEnumerable<MemberInfo> primaryKeys, IDictionary<string,
                 };
             }
 
+            /// <summary>
+            /// Generates SQL query that can be used to check if a table exists
+            /// </summary>
+            public static string GetTableExistsQuery(string schema, string tableName)
+            {
+                return $@"
+                    if (exists (select * from INFORMATION_SCHEMA.TABLES where TABLE_NAME = '{tableName}' and TABLE_SCHEMA = {schema}))
```
chlafreniere, line 243: Let's chat about this a bit. I'd like to minimize the number of calls to the database (while maintaining a nice customer experience with human-readable error messages 😄).

Currently, what's the error that a user sees? Is it around not being able to find a primary key, or is it some giant error message from sqlclient?

lucyzhang929: This is the current error:
System.Private.CoreLib: Exception while executing function: TimerTriggerProducts. Microsoft.Azure.WebJobs.Host: Error while handling parameter products after function returned:. **SqlBinding: Did not retrieve any primary keys for 'Products' in schema 'SCHEMA_NAME().' Cannot generate upsert command without them.**

Maybe instead of having a new query, we can add the check to the GetColumnsDefinitionsQuery? If the table exists then return the column definitions, else return false.

chlafreniere: @lucyzhang929 if we can work it out where we can distinguish between:
- table doesn't exist and
- table exists but no primary key exists on table

and still use the existing query, then that'd be ideal. lemme know if you need any help!

src/SqlBinding/SqlAsyncCollector.cs
chlafreniere: If the schema is always returned with quotes around it, then at the very least we should include that in the comments above the `GetTableAndSchema` method.

However, I think we should have consistency here. Either both table name + schema should have quotes around them, or neither should from `GetTableAndSchema`. Let's just be consistent here.

lucyzhang929: I think this is because schema can also be USER_SCHEMA(), which is returned without quotes. For consistency, we can return both tableName and schema with quotes.

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -159,10 +159,17 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
+        /// If there is no schema in fullName, SCHEMA_NAME is returned as schema.
+        /// </summary>
+        /// <param name="fullName">
+        /// Full name of table, including schema (if exists).
+        /// </param>
         public static void GetTableAndSchema(string fullName, out string schema, out string tableName)
         {
             // defaults
-            tableName = fullName;
+            tableName = $"'{fullName}'";
```
Charles-Gagnon, line 172: Make sure to escape the name properly - `$"'{fullName.replace("'", "''")}'"`

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -268,11 +268,11 @@ public static string GetColumnDefinitionsQuery(string schema, string tableName)
                                 when DATETIME_PRECISION is not null and DATA_TYPE not in ('datetime', 'date', 'smalldatetime') then '(' + cast(DATETIME_PRECISION as varchar(1)) + ')'
 			                    when DATA_TYPE in ('decimal', 'numeric') then '(' + cast(NUMERIC_PRECISION as varchar(9)) + ',' + + cast(NUMERIC_SCALE as varchar(9)) + ')'
 			                    else ''
-		                    end as {ColumnDefinition}	
+		                    end as {ColumnDefinition}
                     from 
 	                    INFORMATION_SCHEMA.COLUMNS c
                     where
-	                    c.TABLE_NAME = '{tableName}'         
+	                    c.TABLE_NAME = {tableName}
```
Charles-Gagnon, line 275: I'd suggest also updating the param names to be `quotedSchema` and `quotedTableName`

Charles-Gagnon (changes requested):

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -159,11 +159,18 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
+        /// </summary>
+        /// <param name="fullName">
+        /// Full name of table, including schema (if exists).
+        /// </param>
+        public static void GetTableAndSchema(string fullName, out string quotedSchema, out string quotedTableName)
         {
             // defaults
-            tableName = fullName;
-            schema = "SCHEMA_NAME()"; // default to user schema
+            quotedTableName = $"'{fullName.Replace("'", "''")}'";
```
Charles-Gagnon, line 172: comment here explaining why this is necessary

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -173,8 +180,8 @@ public static void GetTableAndSchema(string fullName, out string schema, out str
 
             if (pieces.Length == 2)
             {
-                schema = $"'{pieces[0]}'";
-                tableName = pieces[1];
+                quotedSchema = $"'{pieces[0]}'";
+                quotedTableName = $"'{pieces[1]}'";
```
Charles-Gagnon, line 184: You need to escape the ' in this as well

src/SqlBinding/SqlAsyncCollector.cs
Charles-Gagnon: ping on updating these

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -173,8 +180,8 @@ public static void GetTableAndSchema(string fullName, out string schema, out str
 
             if (pieces.Length == 2)
             {
-                schema = $"'{pieces[0]}'";
-                tableName = pieces[1];
+                quotedSchema = $"'{pieces[0].Replace("'", "''")}'";
```
Charles-Gagnon, line 183: 1. Would suggest adding a comment here too about the replacement
2. Or probably useful to just make a helper function to do the escaping (reduce change of typos and the name makes it clearer what it's doing)
3. You can also just escape the fullName once at the beginning instead of escaping 3 separate times.

Charles-Gagnon (approved):

chlafreniere (approved):

PR #82, opened 2021-09-30
Add CODEOWNERS

Charles-Gagnon:
Will help make sure we get added - plus then we don't have to wait for Github to load up the names when adding reviewers 😄

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+# Syntax can be found here: https://docs.github.com/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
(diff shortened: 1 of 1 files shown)


chlafreniere (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+# Syntax can be found here: https://docs.github.com/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+@Charles-Gagnon @lucyzhang929 @chlafreniere
```
chlafreniere, line 5: It's interesting how syntax highlighting only highlights the first line. This is the right format for this file, right? (just making sure 😄)

Charles-Gagnon: Should be based on the doc - guess we'll find out in the next PR 😁

PR #84, opened 2021-09-30
Fix IDE0019 pattern matching issue

chlafreniere:
<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>

wasn't included in all of the .csproj files.

Happy to split that out if we want into a separate PR. Could also put this in a .props file, but that seemed a bit silly for one property. Happy to hear opinions here.

Finally, I'm unclear why it shows every line as changed. Is that because line endings are defined as clrf? It only shows a couple of lines as changed for me in VS Code:

image

### DIFF
--- a/test/SqlExtension.Tests/TestData.cs
+++ b/test/SqlExtension.Tests/TestData.cs
@@ -2,7 +2,9 @@
 
 namespace SqlExtension.Tests
 {
+    #pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
     public class TestData
+    #pragma warning restore CS0659
     {
         public int ID { get; set; }
 
@@ -20,11 +22,6 @@ namespace SqlExtension.Tests
             }
             return this.ID == otherData.ID && this.Cost == otherData.Cost && ((this.Name == null && otherData.Name == null) ||
                 string.Equals(this.Name, otherData.Name, StringComparison.OrdinalIgnoreCase)) && this.Timestamp.Equals(otherData.Timestamp);
-        }
-
-        public override int GetHashCode()
-        {
-            return HashCode.Combine(ID, Name, Cost, Timestamp);
         }
     }
 }
--- a/test/SqlExtension.Tests/TestData.cs
+++ b/test/SqlExtension.Tests/TestData.cs
@@ -2,9 +2,7 @@
 
 namespace SqlExtension.Tests
 {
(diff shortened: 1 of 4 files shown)


test/SqlExtension.Tests/TestData.cs
```diff
@@ -1,28 +1,27 @@
...
-                string.Equals(this.Name, otherData.Name, StringComparison.OrdinalIgnoreCase)) && this.Timestamp.Equals(otherData.Timestamp);
-        }
-    }
-}
+﻿using System;
+
+namespace SqlExtension.Tests
+{
+    #pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
+    public class TestData
```
chlafreniere, line 6: Could all of these changes be related to line endings?

test/SqlExtension.Tests/TestData.cs
```diff
@@ -1,28 +1,27 @@
...
-            return this.ID == otherData.ID && this.Cost == otherData.Cost && ((this.Name == null && otherData.Name == null) ||
-                string.Equals(this.Name, otherData.Name, StringComparison.OrdinalIgnoreCase)) && this.Timestamp.Equals(otherData.Timestamp);
-        }
-    }
-}
+﻿using System;
+
+namespace SqlExtension.Tests
+{
+    #pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
```
Charles-Gagnon, line 5: We should be implementing GetHashCode as well https://stackoverflow.com/questions/371328/why-is-it-important-to-override-gethashcode-when-equals-method-is-overridden

test/SqlExtension.Tests/TestData.cs
Charles-Gagnon: Seems likely...I assume you see them as `\n` locally?

chlafreniere: https://github.com/Azure/azure-functions-sql-extension/compare/chlafren/testeroni

Definitely a line ending thing. Personally, I don't care either way. Any strong opinions on how we should enforce this (if at all?)?

Charles-Gagnon: I think we can remove that - github takes care of that stuff for us generally (and if there are files that are for specific OSs such as .bat files then we can control that separately through gitattributes like vs code does https://github.com/microsoft/vscode/blob/main/.gitattributes)

Charles-Gagnon (approved):

PR #85, opened 2021-09-30
Fix IDE0060

Charles-Gagnon:
Removed EnableNETAnalyzers from the other project either - it seems that isn't needed as long as we're building with .NET 5+


Charles-Gagnon:
Ok, verified that the build fails as expected with the changes I made in https://github.com/Azure/azure-functions-sql-extension/pull/87. So that's good to know 😄

chlafreniere (approved):
Apologies for the merge conflict 😄

PR #86, opened 2021-09-30
Fix CODEOWNERS

Charles-Gagnon:


chlafreniere (approved):
Ahh, now all of the names are highlighted in the same way 😄

Charles-Gagnon:
Guess I should have looked into that a bit more 😛

PR #93, opened 2021-10-01
Fix IDE0063

Charles-Gagnon:
Use simplified using statements - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0063

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,6 @@ dotnet_analyzer_diagnostic.category-Style.severity = error
 dotnet_analyzer_diagnostic.category-CodeQuality.severity = error
 # Disable warnings from existing issues until we can go through and fix them
 dotnet_diagnostic.IDE0130.severity = silent
-dotnet_diagnostic.IDE0063.severity = silent
 dotnet_diagnostic.IDE0007.severity = silent
 dotnet_diagnostic.IDE0008.severity = silent
 dotnet_diagnostic.IDE0047.severity = silent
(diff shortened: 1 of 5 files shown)


chlafreniere:
This is interesting, new syntax in C# 8.0. Can we just make sure that dispose is called when we expect and that this doesn't introduce behavior changes?

Charles-Gagnon:
The docs say it does and I'm inclined to believe they continue to work as expected : https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

But I guess I can test it locally if you think there might be an issue.

chlafreniere:
> The docs say it does and I'm inclined to believe they continue to work as expected : https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement
> 
> But I guess I can test it locally if you think there might be an issue.
> 
> ```
> The using statement ensures that Dispose (or DisposeAsync) is called even if an exception occurs within the using block. You can achieve the same result by putting the object inside a try block and then calling Dispose (or DisposeAsync) in a finally block; in fact, this is how the using statement is translated by the compiler. The code example earlier expands to the following code at compile time (note the extra curly braces to create the limited scope for the object):
> ...
> The newer using statement syntax translates to similar code. The try block opens where the variable is declared. The finally block is added at the close of the enclosing block, typically at the end of a method.
> ```
> 
> It's all just syntactic sugar in the end - the generated IL should be the same.

Docs are good enough for me. No need 😄

chlafreniere (approved):

PR #97, opened 2021-10-01
Fix IDE0036

Charles-Gagnon:
Order modifiers - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036

### DIFF
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -23,10 +23,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
     /// <typeparam name="T">A user-defined POCO that represents a row of the user's table</typeparam>
     internal class SqlAsyncCollector<T> : IAsyncCollector<T>
     {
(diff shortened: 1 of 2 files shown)


lucyzhang929 (approved):

src/SqlBinding/SqlAsyncCollector.cs
```diff
@@ -23,10 +23,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
     /// <typeparam name="T">A user-defined POCO that represents a row of the user's table</typeparam>
     internal class SqlAsyncCollector<T> : IAsyncCollector<T>
     {
-        private readonly static string RowDataParameter = "@rowData";
-        private readonly static string ColumnName = "COLUMN_NAME";
-        private readonly static string ColumnDefinition = "COLUMN_DEFINITION";
-        private readonly static string NewDataParameter = "cte";
+        private static readonly string RowDataParameter = "@rowData";
```
chlafreniere, line 26: Should these be readonly or const?

chlafreniere (approved):

PR #98, opened 2021-10-01
Enable IDE0017 - Use object initializers

chlafreniere:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0017

Totally get that lines are being added, but I find this much clearer to inspect initial property values.


Charles-Gagnon (approved):
Yeah, initializers are great.

PR #99, opened 2021-10-01
Fix IDE0058

Charles-Gagnon:
Fix unused expression errors - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0058

I have mixed feelings about this one...none of these were actually issues and some of them are pretty weird to be returning values anyways (why does the query builder return itself when appending?)

But not using return values can sometimes indicate a bug (such as not updating a string when doing replace) so I think it's better to have this enabled and deal with a little bit of extra discard code.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,6 @@ dotnet_diagnostic.IDE0008.severity = silent
 dotnet_diagnostic.IDE0047.severity = silent
 dotnet_diagnostic.IDE0036.severity = silent
 dotnet_diagnostic.IDE0130.severity = silent
-dotnet_diagnostic.IDE0058.severity = silent
 dotnet_diagnostic.IDE0003.severity = silent
 dotnet_diagnostic.IDE0032.severity = silent
 dotnet_diagnostic.IDE0046.severity = silent
--- a/src/SqlBinding/SqlAsyncCollector.cs
+++ b/src/SqlBinding/SqlAsyncCollector.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 }
                 finally
                 {
-                    _rowLock.Release();
+                    _ = _rowLock.Release();
                 }
             }
         }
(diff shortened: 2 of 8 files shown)


chlafreniere:
Hmm, I'm not so sure about this one. Is there some finer-tuned control that we can use to still prevent some class of issues but not have to put `_ = ` everywhere?

Charles-Gagnon:
I guess we could look into something like what Zi did in https://github.com/Azure/azure-functions-sql-extension/pull/100#pullrequestreview-769403712

But honestly I don't really think it's that huge of a deal. That suppressions file is kind of messy and this kind of stuff isn't going to cause any huge amounts of pain in the end.

zijchen:
> I guess we could look into something like what Zi did in [#100 (review)](https://github.com/Azure/azure-functions-sql-extension/pull/100#pullrequestreview-769403712)
> 
> But honestly I don't really think it's that huge of a deal. That suppressions file is kind of messy and this kind of stuff isn't going to cause any huge amounts of pain in the end.

I auto-generated the suppressions file from Visual Studio. The default is more granular and you need to suppress each occurrence, but I believe you can suppress it at a broader level, like everything under a namespace: https://stackoverflow.com/questions/11359652/suppressmessage-for-a-whole-namespace

Charles-Gagnon:
@chlafreniere @lucyzhang929 What do you think about just not enabling this rule? I'm not sure it's worth the trouble to suppress it or add the extra code to avoid it - the impact is pretty easy to catch and either way we approach the fix (suppress or add discards) results in a bunch of extra stuff. And it's pretty likely more would keep needing to be suppressed as we keep adding stuff which is just going to get even more annoying.

chlafreniere:
> @chlafreniere @lucyzhang929 What do you think about just not enabling this rule? I'm not sure it's worth the trouble to suppress it or add the extra code to avoid it - the impact is pretty easy to catch and either way we approach the fix (suppress or add discards) results in a bunch of extra stuff. And it's pretty likely more would keep needing to be suppressed as we keep adding stuff which is just going to get even more annoying. 

I'm onboard. Let's not enable this one.

PR #103, opened 2021-10-03
Fix IDE0003

Charles-Gagnon:
Qualify member properties/methods/etc. with this - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009

Note that this (and a few other rules) currently can't fail the build by design. 

See https://github.com/dotnet/roslyn/issues/53215 for the reason why and https://github.com/dotnet/roslyn/blob/main/src/Analyzers/Core/Analyzers/EnforceOnBuildValues.cs#L106 for the list of rules that this happens for.


chlafreniere:
I'm curious about this one. For properties (just taking an example), there seems like there's a benefit to being explicit here, and not hunting for a variable that could be in a couple of scopes (e.g. during PR time). What do you two think?

Charles-Gagnon:
The general idea here is that it removes the qualifier if it's in the same scope that you're in. So this is unnecessary in these cases since you know that you're already working in this object so it couldn't be anything else. (for properties especially)

But honestly this is one of those things where I don't have a preference either way. It doesn't impact anything code wise so as long as we're consistent I'm fine with whatever. I just went with this since it was the default VS generated so I figured that's more likely to be consistent across other projects. (although I know SMO greatly prefers using this for everything)

Charles-Gagnon:
Offline discussion, `this` is good 😄

test/SqlExtension.Tests/SqlInputBindingTests.cs
```diff
@@ -114,7 +114,7 @@ public void TestValidCommandType()
             Assert.Equal(System.Data.CommandType.Text, command.CommandType);
             Assert.Equal(query, command.CommandText);
 
-            string procedure = "StoredProceudre";
+            string procedure = "StoredProcedure";
```
Charles-Gagnon, line 117: Fixed small spelling error too

chlafreniere (approved):

PR #105, opened 2021-10-03
Fix IDE0046

Charles-Gagnon:
Use conditional expression for return - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0046

I decided to change this to false since having this be true just resulted in some messy looking conditional return statements and don't really seem like something that would actually help avoid any errors.


chlafreniere (approved):
Agreed. This would be quite messy in some situations.

PR #112, opened 2021-10-05
Enable integration test with timer trigger

zijchen:
I had to exclude the timer trigger sample function in my previous PR because it conflicted with the rest of the function runs and gave me inconsistent results. This PR adds it back and adds support for timer triggers in integration tests by enabling Azurite. I had to change the function and some other tests a bit so they don't conflict with each other.


test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
```diff
@@ -85,8 +87,23 @@ public void QueueTriggerProductsTest()
...
+            // Function should add 100 rows (excluding ProductId >= 10000 which could be added by the timer trigger function)
+            Assert.Equal(100, this.ExecuteScalar("SELECT COUNT(1) FROM Products WHERE ProductId < 10000"));
+        }
+
+        [Fact]
+        public void TimerTriggerProductsTest()
+        {
+            // Since this function runs on a schedule (every 30 seconds), we don't need to invoke it.
+            // However the timer starts as soon as the Functions host starts, so we need to take that into account when calculating how long to wait.
+            // In total we will wait 60 seconds + 1 ms which should allow the function to run twice, generating 2000 rows of data.
```
Charles-Gagnon: Can we guarantee that it'll only ever have 2000 rows though? If it starts as soon as the host starts then what if this test isn't ran until much later?

I think it would be fine to just refactor this test to wait for at least 2 ticks and then just verify that it has inserted at least 2 ticks worth of rows. (I don't think we should be checking they're exactly equal - since again that's non-deterministic depending on how fast the SELECT COUNT query executes)

test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
```diff
@@ -85,8 +87,23 @@ public void QueueTriggerProductsTest()
 
             Thread.Sleep(5000);
 
-            // Function should add 100 rows
-            Assert.Equal(100, this.ExecuteScalar("SELECT COUNT(1) FROM Products"));
+            // Function should add 100 rows (excluding ProductId >= 10000 which could be added by the timer trigger function)
```
Charles-Gagnon, line 90: I would rather we make a separate table for each function to use - easier to debug and keep track of state

Charles-Gagnon (changes requested):

test/SqlExtension.Tests/Integration Tests/IntegrationTestBase.cs
```diff
@@ -22,7 +22,12 @@ public class IntegrationTestBase : IDisposable
         /// <summary>
         /// Host process for Azure Function CLI
         /// </summary>
-        private Process FunctionHost;
+        protected Process FunctionHost;
+
+        /// <summary>
+        /// Host process for Azurite emulator for local Azure Storage, required by some trigger functions.
```
Charles-Gagnon, line 28: Why is this necessary just for triggers?

test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
zijchen: Does SQL bindings support different tables? I see the table defined in the function attributes but if we use different tables, we would need to create them dynamically when we set up the database.

test/SqlExtension.Tests/Integration Tests/IntegrationTestBase.cs
zijchen: The triggers need something to bind to at function startup, and the local storage emulator provides that.

test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
Charles-Gagnon: The table name is the first param. I'd suggest making a new table and add a new binding just to use for the trigger though since we still want the samples to demonstrate both input and output bindings for the same table 

https://github.com/Azure/azure-functions-sql-extension/blob/main/samples/SqlExtensionSamples/OutputBindingSamples/AddProduct.cs#L17

test/SqlExtension.Tests/Integration Tests/IntegrationTestBase.cs
Charles-Gagnon: What is this "something"? And why are trigger bindings the only one that need this? Do you have docs you can link to for this?

zijchen: https://docs.microsoft.com/en-us/azure/azure-functions/functions-develop-local

If you look at the doc for "AzureWebJobsStorage", it says 
`Contains the connection string for an Azure storage account. Required when using triggers other than HTTP. For more information, see the AzureWebJobsStorage reference.
When you have the Azure Storage Emulator installed locally and you set AzureWebJobsStorage to UseDevelopmentStorage=true...`

Charles-Gagnon: Please add this link to the docs in the comment

test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
zijchen: With the redesign this is not needed any more

test/SqlExtension.Tests/Integration Tests/SqlOutputBindingIntegrationTests.cs
zijchen: Changed it so we just check that data has been added. Also changed the Azure function to fire more frequently so we don't have to wait as long.

samples/SqlExtensionSamples/OutputBindingSamples/TimerTriggerProducts.cs
```diff
@@ -8,24 +8,28 @@
...
 
-namespace SqlExtensionSamples.TriggerBindingSamples
+namespace SqlExtensionSamples
 {
     public static class TimerTriggerProducts
     {
         private static int _executionNumber = 0;
+
+        /// <summary>
+        /// This timer function runs every 5 seconds. Each time it generates 1000 rows of data, starting at a ProductID of 10000.
```
Charles-Gagnon, line 18: Do we need to start with the product ID at 10000 anymore?

Charles-Gagnon (approved):

samples/SqlExtensionSamples/OutputBindingSamples/TimerTriggerProducts.cs
zijchen: We don't any more, forgot to update comment

chlafreniere (approved):

Issue #64, opened 2021-09-21
table and schema names improperly parsed


dzsquared:
https://github.com/Azure/azure-functions-sql-extension/blob/ad352485aad33a10d457d550fc80a0475d3a93fb/src/SqlBinding/SqlBindingUtilities.cs#L162

Utility function splits schema and table name by period after removing square brackets, resulting in broken functionality if either table or schema name have a period in them.

if there are square brackets, use them to delimit the schema/table names, then do any further parsing

PR #114, opened 2021-10-05
Use ScriptDom to parse schema and table name

lucyzhang929:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/64

### DIFF
--- a/src/SqlBinding/SqlBindingUtilities.cs
+++ b/src/SqlBinding/SqlBindingUtilities.cs
@@ -164,6 +164,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         }
 
         /// <summary>
+        /// Escape any existing quotes and add quotes around the identifier.
+        /// </summary>
+        /// <param name="identifier">The string to quote.</param>
+        /// <returns>The escaped and quoted string.</returns>
+        public static string QuoteIdentifier(string identifier)
+        {
+            return $"'{identifier.Replace("'", "''")}'";
+        }
+
+        /// <summary>
         /// Use ScriptDom to parse schema and tableName and return them with quotes.
         /// If there is no schema in fullName, SCHEMA_NAME() is returned as schema.
         /// </summary>
@@ -202,9 +212,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
             public override void Visit(SchemaObjectName node)
             {
-                // ensure names with quotes are properly escaped
-                this.quotedSchema = node.SchemaIdentifier != null ? $"'{node.SchemaIdentifier.Value.Replace("'", "''")}'" : "SCHEMA_NAME()";
-                this.quotedTableName = $"'{node.BaseIdentifier.Value.Replace("'", "''")}'";
+                this.quotedSchema = node.SchemaIdentifier != null ? QuoteIdentifier(node.SchemaIdentifier.Value) : "SCHEMA_NAME()";
+                this.quotedTableName = QuoteIdentifier(node.BaseIdentifier.Value);
             }
         }
     }


chlafreniere:
Please add unit tests 😄

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -173,21 +175,31 @@ public static void GetTableAndSchema(string fullName, out string quotedSchema, o
...
-
-            // if in format schema.table, split into two parts for query
-            string[] pieces = cleanName.Split('.');
-
-            if (pieces.Length == 2)
+            foreach (ParseError err in errors)
             {
-                quotedSchema = $"'{pieces[0]}'";
-                quotedTableName = $"'{pieces[1]}'";
+                Console.WriteLine(err.Message);
```
Charles-Gagnon, line 184: If there's errors I'm not sure we should be continuing

(and even if we were this should be `Console.WriteError`)

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -173,21 +175,31 @@ public static void GetTableAndSchema(string fullName, out string quotedSchema, o
...
+    }
+
+    public class Vistor : TSqlFragmentVisitor
+    {
+        internal string quotedSchema;
+        internal string quotedTableName;
+
+        public override void Visit(SchemaObjectName node)
+        {
+            this.quotedSchema = node.SchemaIdentifier != null ? $"'{node.SchemaIdentifier.Value}'" : "SCHEMA_NAME()";
```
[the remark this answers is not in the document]
lucyzhang929: I am escaping the fullName before passing it into the visitor.

test/SqlExtension.Tests/SqlOutputBindingTests.cs
```diff
@@ -40,5 +40,41 @@ public async Task TestAddAsync()
             };
             await collector.AddAsync(data);
         }
+
+        [Fact]
+        public void TestGetTableAndSchema()
```
Charles-Gagnon, line 45: Would suggest combining these all into a single test using theories and inline data

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -173,21 +175,39 @@ public static void GetTableAndSchema(string fullName, out string quotedSchema, o
...
+            tree.Accept(visitor);
+            quotedSchema = visitor.quotedSchema;
+            quotedTableName = visitor.quotedTableName;
+        }
+    }
+
+    /// <summary>
+    /// Get the schema and table name with quotes from the SchemaObjectName.  
+    /// </summary>
+    internal class QuotedTSqlFragmentVisitor : TSqlFragmentVisitor
```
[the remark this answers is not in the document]
lucyzhang929: When I change this to private, I get the following error: `Elements defined in a namespace cannot be explicitly declared as private, protected, protected internal, or private protected [SqlBinding]`.
I think classes can only be public or internal?

test/SqlExtension.Tests/SqlOutputBindingTests.cs
```diff
@@ -40,5 +40,41 @@ public async Task TestAddAsync()
...
+            string fullName = "Products";
+            SqlBindingUtilities.GetTableAndSchema(fullName, out string quotedSchema, out string quotedTableName);
+            Assert.Equal("SCHEMA_NAME()", quotedSchema);
+            Assert.Equal("'Products'", quotedTableName);
+        }
+
+        [Fact]
+        public void TestGetTableAndSchemaWithBrackets()
+        {
+            string fullName = "[dbo].[Products]";
```
chlafreniere, line 65: Just for sanity purposes, please add a test case where only the schema is square bracketed and another where only the table name is bracketed

src/SqlBinding/SqlBindingUtilities.cs
```diff
@@ -162,31 +164,47 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
+            internal string quotedSchema;
+            internal string quotedTableName;
 
-            if (pieces.Length == 2)
+            public override void Visit(SchemaObjectName node)
             {
-                quotedSchema = $"'{pieces[0]}'";
-                quotedTableName = $"'{pieces[1]}'";
+                // ensure names with quotes are properly escaped
+                this.quotedSchema = node.SchemaIdentifier != null ? $"'{node.SchemaIdentifier.Value.Replace("'", "''")}'" : "SCHEMA_NAME()";
```
Charles-Gagnon, line 206: Let's make a helper method in the utils to do the actual quoting to reduce chance of copy/paste errors (and we'll likely be wanting to re-use that in other contexts)

```
public static string QuoteIdentifier(string identifier)
{
    return $"'{identifier.Replace("'", "''")}'";
}
```

Charles-Gagnon (approved):

test/SqlExtension.Tests/SqlOutputBindingTests.cs
```diff
@@ -40,5 +40,19 @@ public async Task TestAddAsync()
...
             await collector.AddAsync(data);
         }
+
+        [Theory]
+        [InlineData("dbo.Products", "'dbo'", "'Products'")]
+        [InlineData("Products", "SCHEMA_NAME()", "'Products'")]
+        [InlineData("[dbo].[Products]", "'dbo'", "'Products'")]
+        [InlineData("[dbo].Products", "'dbo'", "'Products'")]
+        [InlineData("dbo.[Products]", "'dbo'", "'Products'")]
+        [InlineData("[Prod'ucts]", "SCHEMA_NAME()", "'Prod''ucts'")]
```
Charles-Gagnon, line 50: Would suggest adding error cases (for example things that would make the parser fail). I assume having the fullname be something like `dbo.my'table` would be one of those (since to work the table name should be quoted)

Charles-Gagnon: These can be a separate test if that's easier since it's slightly different logic to test that scenario

PR #116, opened 2021-10-06
Refactor folder structure

Charles-Gagnon:


chlafreniere:
@Charles-Gagnon can you please provide a description on what the motivation behind the refactor is?

In addition, looks like the integration tests failed.

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -12,20 +12,27 @@
...
     <DebugType>embedded</DebugType>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.Sql</PackageId>
+    <RootNamespace></RootNamespace>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
   </PropertyGroup>
 
+  <ItemGroup>
```
chlafreniere, line 22: what's this newfolder stuff?

Charles-Gagnon: hah, oops. Was messing around with stuff and apparently VS didn't clean this up as nicely as I thought

chlafreniere: just saw the property above too. coul dyou take a look?

Charles-Gagnon: Yeah the root namespace? I added a comment explaining that. I think it's a bug with the analyzer or something...for files in the root of the project folder it always gives an error about the namespace.

I'll probably restructure them to be in sub-folders (which fixes that issue) in the future but for now this seems fine enough.

PR #119, opened 2021-10-07
correct licensing on files and editorconfig

dzsquared:
Correcting file licenses to match the repository's license


dzsquared:
I'm a dummy that left this open too long, will close and redo to avoid fighting through all the drift.

PR #121, opened 2021-10-07
Add test README

zijchen:
Instructions on how to run integration tests

### DIFF
--- a/test/SqlExtension.Tests/README.md
+++ b/test/SqlExtension.Tests/README.md
@@ -1,0 +1,25 @@
+# Running Tests
+
+## Running Integration Tests
+Our integration tests are based on functions from the samples project. To run integration tests, you will need
+1. [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local#install-the-azure-functions-core-tools) - This is used to start the functions runtime.
+
+   Installation with npm:
+   ```
+   npm install -g azure-functions-core-tools@3
+   ```
+2. [Azurite Emulator for Local Azure Storage](https://docs.microsoft.com/azure/storage/common/storage-use-azurite?tabs=npm#install-and-run-azurite) - This is required to run non-HTTP binding functions.
+
+   Installation with npm:
+   ```
+   npm install -g azurite
+   ```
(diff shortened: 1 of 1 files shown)


test/SqlExtension.Tests/README.md
```diff
@@ -0,0 +1,25 @@
+# Running Tests
+
+## Running Integration Tests
+Our integration tests are based on functions from the samples project. To run integration tests, you will need
+1. [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local#install-the-azure-functions-core-tools) - This is used to start the functions runtime.
+
+   Installation with npm:
+   ```
+   npm install -g azure-functions-core-tools@3
```
Charles-Gagnon, line 9: Do we actually need to use v3 or can we just have it pull the latest?

test/SqlExtension.Tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+   ```
+3. A local SQL Server instance - This is used by tests to verify that data is correctly added/fetched from the database when a test Function is run. You just need the server to be up and running, the tests will create the database and tables which will be cleaned up afterwards.
+
+   - You can either have a SQL Server installation with `localhost` available for connection via integrated security, or
+   - Start a SQL Server instance with Docker
+     ```
+     docker pull mcr.microsoft.com/mssql/server:2019-latest
+     docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD={your_password}" -e "MSSQL_PID=Express" -p 1433:1433 --name sql1 -h sql1 -d mcr.microsoft.com/mssql/server:2019-latest
+     ```
+     After the Docker image is running, you just need to set `SA_PASSWORD` environment variable to `{your_password}` and can run tests normally.
```
Charles-Gagnon, line 25: Would be good to note that you can also use integrated auth if `SA_PASSWORD` is empty

zijchen: I don't think integrated auth is enabled by default when starting the Docker image, which is why you need to set the SA_PASSWORD when starting the image
https://hub.docker.com/_/microsoft-mssql-server

Charles-Gagnon: Yes, this would be for a local server install

zijchen: Added a note

lucyzhang929 (approved):

PR #122, opened 2021-10-11
Enable integration tests on Windows using localdb

zijchen:
- Integration tests on Windows will use localdb as test server as it comes pre-installed on agent machines. 
- Needed to add a step to set npm installation path as on Windows agents npm installs to a non-traditional path.


builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -15,9 +16,16 @@ steps:
   inputs:
     version: '3.1.x'
 
-- script: npm install -g azure-functions-core-tools@3
+- script: npm install -g azure-functions-core-tools
```
Charles-Gagnon, line 19: This is fine for now, but we should be aware of major changes potentially breaking us. But for now I think it's best to take the latest and handle it later if something does end up breaking because of that.

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -15,9 +16,16 @@ steps:
...
   displayName: 'Install Azure Functions Core Tools'
 
+- script: npm install -g azurite
+  displayName: 'Install Azurite Local Storage Emulator'
+
+- bash: echo "##vso[task.setvariable variable=nodeModulesPath]C:\npm\prefix\node_modules"
```
Charles-Gagnon, line 25: We generally shouldn't be hardcoding paths like this - especially when we don't control the machine. 

Where is this path coming from anyways? I'm not seeing this path on my machine...

test/Integration/IntegrationTestBase.cs
```diff
@@ -171,7 +182,7 @@ protected void StartFunctionHost(string functionName)
         private string GetFunctionsCoreToolsPath()
         {
             // Determine npm install path from either env var set by pipeline or OS defaults
-            string nodeModulesPath = Environment.GetEnvironmentVariable("node_modules_path");
+            string nodeModulesPath = Environment.GetEnvironmentVariable("NODE_MODULES_PATH");
```
Charles-Gagnon, line 185: We already have logic for determining the path below - why do you need to do this too?

builds/azure-pipelines/template-steps-build-test.yml
zijchen: Didn't know of this command, will try with this

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -15,9 +16,16 @@ steps:
...
     version: '3.1.x'
 
-- script: npm install -g azure-functions-core-tools@3
+- script: npm install -g azure-functions-core-tools
   displayName: 'Install Azure Functions Core Tools'
 
+- script: npm install -g azurite
+  displayName: 'Install Azurite Local Storage Emulator'
+
+- bash: echo "##vso[task.setvariable variable=nodeModulesPath]$(npm root -g)"
```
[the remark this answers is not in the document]
zijchen: If you can think of alternatives to this method I'd be glad to try it out

builds/azure-pipelines/template-steps-build-test.yml
zijchen: Updated yaml to set the path variable from the output of "npm root -g"

builds/azure-pipelines/template-steps-build-test.yml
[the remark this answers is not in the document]
zijchen: Added comments to both yaml and cs file

test/Integration/IntegrationTestBase.cs
```diff
@@ -149,9 +149,14 @@ protected void StartAzurite()
...
         /// <summary>
-        /// This starts the Functions runtime with the specified function.
-        /// Note the functionName is different than the route.
+        /// This starts the Functions runtime with the specified function(s).
         /// </summary>
+        /// <remarks>
+        /// - The functionName is different than its route.<br/>
+        /// - You can start multiple functions by passing in a space-separated list of function names.<br/>
+        /// - The full path to the Functions CLI is required in the ProcessStartInfo because UseShellExecute is set to false.
+        /// We cannot both use shell execute and redirect output at the same time: https://docs.microsoft.com//dotnet/api/system.diagnostics.processstartinfo.redirectstandardoutput#remarks
```
Charles-Gagnon, line 158: I'd just put this and the line above it down in the code - they're just for one part of what this method is doing and isn't going to be something that someone looking at the docs will care about typically.

chlafreniere (approved):

PR #128, opened 2021-10-14
Check for extra columns before upsert

lucyzhang929:
Throw an error if there are properties in the AF object that do not exist in the SQL table columns.

Example error that gets thrown:
`SqlBinding: The following properties in SqlExtensionSamples.Product do not exist in the table dbo.Products: ExtraInt, ExtraString.`


src/SqlAsyncCollector.cs
```diff
@@ -155,6 +162,27 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
+        /// <summary>
+        /// Checks if any properties in T do not exist as columns in the table
+        /// to upsert to and returns the extra property names in a List.
+        /// </summary>
+        /// <param name="columns"> The columns of the table to upsert to </param>
+        /// <returns>List of property names that don't exist in the table</returns>
+        private static List<string> GetExtraProperties(IDictionary<string, string> columns)
+        {
+            var properties = typeof(T).GetProperties().ToList();
+            var extraProperties = new List<string>();
```
Charles-Gagnon, line 175: `return typeof(T).GetProperties().ToList().Where(prop => !columns.ContainsKey(prop.Name));`

Charles-Gagnon (commented):
Tests please 😄

samples/OutputBindingSamples/AddProductExtraColumns.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.OutputBindingSamples
+{
+    public static class AddProductExtraColumns
+    {
+        [FunctionName("AddProductExtraColumns")]
```
Charles-Gagnon, line 13: I don't think we should be adding a binding that is wrong - even if that's on purpose. 

What if we added it to the Test project itself?

test/Integration/AddProductExtraColumns.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
+{
+    public static class AddProductExtraColumns
```
Charles-Gagnon, line 11: Add a doc comment explaining what this does

Charles-Gagnon (approved):

PR #130, opened 2021-10-14
Add first draft of CONTRIBUTING.md file

chlafreniere:
Adding first draft of CONTRIBUTING.md file. It is not done, and I will continue to tweak as time goes on, but it's a start, and I'd like to update this iteratively and get it in sooner rather than wait for it to be "done" and then check it in.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,60 @@
+# Contributing
+
+This project welcomes contributions and suggestions. Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.opensource.microsoft.com.
+
+When you submit a pull request, a CLA bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., status check, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the Microsoft Open Source Code of Conduct. For more information see the Code of Conduct FAQ or contact opencode@microsoft.com with any additional questions or comments.
+
+<br>
+
+## Contributer Getting Started
+
+### SQL Setup
+
+This requires already having a SQL database. If you need to create a SQL database, we recommend one of 2 options:
+- [Local SQL Server running in a Docker container](https://docs.microsoft.com/sql/linux/quickstart-install-connect-docker)
+- [Azure SQL Database](#Create-Azure-SQL-Database).
+
+A primary key must be set in your SQL table before using the bindings. To do this, run the below SQL commands in the SQL query editor.
+
+1. Ensure there are no NULL values in the primary key column. The primary key will usually be an ID column.
+
+    ```sql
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+This requires already having a SQL database. If you need to create a SQL database, we recommend one of 2 options:
+- [Local SQL Server running in a Docker container](https://docs.microsoft.com/sql/linux/quickstart-install-connect-docker)
+- [Azure SQL Database](#Create-Azure-SQL-Database).
+
+A primary key must be set in your SQL table before using the bindings. To do this, run the below SQL commands in the SQL query editor.
+
+1. Ensure there are no NULL values in the primary key column. The primary key will usually be an ID column.
+
+    ```sql
+    ALTER TABLE ['your table name'] alter column ['column to be primary key'] int NOT NULL
```
Charles-Gagnon, line 24: keywords should be uppercased typically, and I'd suggest getting rid of the single quotes since that might be confusing for people who aren't familiar with T-SQL syntax

`ALTER TABLE [TableName] ALTER COLUMN [PrimaryKeyColumnName] int NOT NULL`

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+    ```
+
+2. Set primary key column.
+
+    ```sql
+    ALTER TABLE ['your table name'] ADD CONSTRAINT PKey PRIMARY KEY CLUSTERED (['column to be primary key']);
```
Charles-Gagnon, line 30: `ALTER TABLE [TableName] ADD CONSTRAINT PKey PRIMARY KEY CLUSTERED ([PrimaryKeyColumn]);`

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Contributer Getting Started
+
+### SQL Setup
+
+This requires already having a SQL database. If you need to create a SQL database, we recommend one of 2 options:
+- [Local SQL Server running in a Docker container](https://docs.microsoft.com/sql/linux/quickstart-install-connect-docker)
+- [Azure SQL Database](#Create-Azure-SQL-Database).
+
+A primary key must be set in your SQL table before using the bindings. To do this, run the below SQL commands in the SQL query editor.
```
Charles-Gagnon, line 19: Would suggest mentioning that this is a one-time step, they can skip if they already have it done.

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+```bash
+git clone https://github.com/Azure/azure-functions-sql-extension
+cd azure-functions-sql-extension
+code .
+```
+3. Install extensions when prompted after VS Code opens
+   - Note: This includes the Azure Functions, C#, and editorconfig extensions
+
+4. Get your SqlConnectionString. Your connection string can be found in your SQL database resource by going to the left blade and clicking 'Connection strings'. Copy the Connection String.
```
Charles-Gagnon, line 49: This is only for Azure, it's probably fine just to skip this (or point to some generic documentation on connection strings)

PR #134, opened 2021-10-19
adjusting naming to cohesive pattern

dzsquared:


Charles-Gagnon (approved):
So it's not a problem that technically we support bindings to any MS SQL server - not just Azure?

Issue #75, opened 2021-09-23
Output Binding Always Passes NULL for Columns Not Defined


chlafreniere:
Let's take the following simple example, where one column has a default value:

```
-- Create the table in the specified schema
CREATE TABLE [dbo].[Products2]
(
    [ProductID] INT PRIMARY KEY, -- Primary Key column
    [Blah] NVARCHAR(2) NOT NULL DEFAULT 'hi'
);
GO
```

When I run the following query, everything inserts just fine:

```
-- Insert rows into table 'Products2' in schema '[dbo]'
INSERT INTO [dbo].[Products2]
( -- Columns to insert data into
 [ProductID]
)
VALUES
( 
 7
)
-- Add more rows here
GO
```

Using edit data, I can graphically see that data is inserted, and the column 'Blah' has the default value (which is expected):

![image](https://user-images.githubusercontent.com/40371649/134583411-7d649ffb-b317-4bed-ac12-7e21dbcdc6f4.png)

**Now**, let's get this repo in the picture. I used the following sample: https://github.com/Azure/azure-functions-sql-extension/blob/chlafren/fixDatetime/samples/SqlExtensionSamples/OutputBindingSamples/AddProduct.cs. Note that the 'Blah' column isn't part of this sample.

(note that the table has the same primary key column name out of convenience)

When I try to execute this function, I see the following in the console:

```
[2021-09-23T20:58:16.288Z] Executed 'AddProduct' (Failed, Id=3144545e-da87-4788-b466-8697fd19563e, Duration=511ms)
[2021-09-23T20:58:16.288Z] System.Private.CoreLib: Exception while executing function: AddProduct. Microsoft.Azure.WebJobs.Host: Error while handling parameter product after function returned:. Core Microsoft SqlClient Data Provider: Cannot insert the value NULL into column 'Blah', table 'foo.dbo.Products2'; column does not allow nulls. UPDATE fails.
```

The `Cannot insert the value NULL into column 'Blah'` part is key.

I would posit that we shouldn't try to insert `NULL` data for columns that aren't explicitly defined. I'm sure there are other cases beyond default constraints where this behavior can be troublesome.

chlafreniere:
Put the P0 label in to make sure that we chat about this behavior.

dzsquared:
I agree.  To restate what @chlafreniere mentioned - if a column in the AF object is undefined, the query should not upsert against that column.

[...]

Issue #67, opened 2021-09-21
Investigate error if schema of upsert doesn't match table schema


Charles-Gagnon:

chlafreniere:
Let's make sure that we define expected behavior here. What should happen when:

- A column doesn't exist in the table but data is provided
- A column does exist but data isn't provided (note that this one is trickier, since there could be default values, NULL could be acceptable, etc., so we certainly shouldn't block users in these cases)

lucyzhang929:
@dzsquared In the case that the row being upserted has a column that does not exist in the table, should we throw an error? show a warning? keep the current behavior (adds the column)?

Issue #38, opened 2021-01-08
Use POCO to filter columns in MERGE


stephlocke:
We have a temporal table that means we won't be inserting into all columns (this also applies for where data may be optional etc). The SQL MERGE generator uses the INFORMATION_SCHEMA to get columns to be used in the MERGE. This means that columns that aren't contained in our class will are included in the MERGE,c ausing it to fail.

The MERGE code generator should subset the list of columns to be MERGED based on the class definition to prevent excess columns being included in the SQL and causing faults.

stephlocke:
A SQL DB/ SQL Server specific solution, might be to use the sys views eg
```
select 
c.name as ColumnName,
types.name +
    case 
        when types.name in ('decimal', 'numeric') then '(' + cast(c.[precision] as varchar(9)) + ',' + + cast(c.scale as varchar(9)) + ')'
        when c.scale > 0 then '(' + cast(c.scale as varchar(1)) + ')'      
        when c.max_length = -1 then '(max)'
        when c.max_length <> -1 then '(' + cast(c.max_length as varchar(4)) + ')'  
        else ''
    end as ColumnDefinition
from sys.columns c
inner join sys.tables t on t.object_id=c.object_id
inner join sys.types on c.user_type_id=types.user_type_id
where t.name = 'load'
and c.generated_always_type = 0
and c.is_computed = 0
```

PR #135, opened 2021-10-20
Modify merge statement to insert columns based on AF Object instead of SQL table columns

lucyzhang929:
This PR fixes https://github.com/Azure/azure-functions-sql-extension/issues/75, fixes https://github.com/Azure/azure-functions-sql-extension/issues/67, fixes https://github.com/Azure/azure-functions-sql-extension/issues/38


src/SqlAsyncCollector.cs
```diff
@@ -212,7 +212,9 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
             }
 
             rowData = JsonConvert.SerializeObject(rowsToUpsert, table.JsonSerializerSettings);
-            newDataQuery = $"WITH {NewDataParameter} AS ( SELECT * FROM OPENJSON({RowDataParameter}) WITH ({string.Join(",", table.ColumnDefinitions)}) )";
+            IEnumerable<string> columnNamesFromPOCO = typeof(T).GetProperties().Select(prop => prop.Name);
```
lucyzhang929, line 215: @chlafreniere @Charles-Gagnon I know we chatted about the use of "POCO" before. I'm using it here to be consistent with other parts of the code but let me know if something else would be preferred.

PR #142, opened 2021-10-29
Add test for no partial upserts

lucyzhang929:


test/Integration/AddProductsNoPartialUpsert.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        // This output binding should throw an error since the ProductsNameNotNull table does not 
+        // allows rows without a Name value. No rows should be upserted to the Sql table.
+        [FunctionName("AddProductsNoPartialUpsert")]
+        public static IActionResult Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproducts-nopartialupsert")]
+            HttpRequest req,
+            [Sql("dbo.ProductsNameNotNull", ConnectionStringSetting = "SqlConnectionString")] ICollector<Product> products)
+        {
+            List<Product> newProducts = ProductUtilities.GetNewProducts(1000);
```
Charles-Gagnon, line 22: It would be nice to use the batch size as a constant in case we change that later

Charles-Gagnon (approved):

PR #144, opened 2021-11-02
Add some additional package props

Charles-Gagnon:
Based on https://github.com/Azure/azure-webjobs-sdk/tree/dev/src/Microsoft.Azure.WebJobs.Extensions.Storage

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -12,7 +12,7 @@
     <IncludeSymbols>false</IncludeSymbols>
     <DebugType>embedded</DebugType>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.Sql</PackageId>
-    <PackageTags>Microsoft Azure WebJobs AzureFunctions SQL</PackageTags>
+    <PackageTags>Microsoft Azure WebJobs AzureFunctions SQL AzureSQL</PackageTags>
     <PackageRequireLicenseAcceptance>True</PackageRequireLicenseAcceptance>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/Azure/azure-functions-sql-extension</PackageProjectUrl>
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -1,17 +1,21 @@
...
     <Authors>Microsoft</Authors>
     <Product>SQL Binding Extension</Product>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateAssemblyInfo>true</GenerateAssemblyInfo>
     <DebugSymbols>true</DebugSymbols>
     <IncludeSymbols>false</IncludeSymbols>
     <DebugType>embedded</DebugType>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.Sql</PackageId>
+    <PackageTags>Microsoft Azure WebJobs AzureFunctions SQL</PackageTags>
```
chlafreniere, line 15: Is there an "AzureSQL" package tag that we should add as well?

chlafreniere (approved):

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
Charles-Gagnon: I'm not seeing anything like that when I search https://www.nuget.org/packages?q=azure+sql

Is AzureSQL (all one word) something you think people would actually be searching? As long as you have azure or sql in it then it'll be fine and that seems to me to be the more common phrases to search

lucyzhang929 (approved):

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
chlafreniere: I have no data to support "AzureSQL". Just wondering if there was another tag beyond "SQL", that's all 😄 . Totally fine.

Charles-Gagnon: Added it, not like it's going to cause problems and might be useful to someone

PR #159, opened 2021-11-11
Cleanup docs

Charles-Gagnon:
Re-organizing and clarifying some parts. Also moved away from having Azure SQL be the only choice to reduce friction for new devs so they can get started ASAP using something like docker. 

Lots more work to do I'm sure but this at least fixes the majority of the things that I had noted could use with some polishing 😄 

Link to the README so you can check out what it actually looks like https://github.com/Azure/azure-functions-sql-extension/blob/chgagnon/cleanupDocs/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@
 
 This repository contains the Azure SQL binding for Azure Functions extension code as well as a quick start tutorial and samples illustrating how to use the binding in different ways.  A high level explanation of the bindings is provided below. Additional information for each is in their respective sample sections.
 
-- **input binding**: takes a SQL query to run and returns the output of the query in the function.
-- **output binding**: takes a list of rows and upserts them into the user table (i.e. If a row doesn't already exist, it is added. If it does, it is updated).
+- **Input Binding**: takes a SQL query to run and returns the output of the query in the function.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -113,47 +160,9 @@ These steps can be done in the Terminal/CLI or with PowerShell. Completing this
...
-    ```sql
-    INSERT INTO [dbo].[Employees] values (1, 'Hello', 'World', 'Microsoft', 'Functions')
-    ```
-
-- Congratulations! You have successfully created an Azure SQL Database! Make sure you complete [Quick Start](#Quick-Start) before continuing to the rest of the tutorial.
-
 ### Input Binding Tutorial
 
-Note: This tutorial requires that the Azure SQL database is setup as shown in [Create Azure SQL Database](#Create-Azure-SQL-Database).
+Note: This tutorial requires that the SQL database is setup as shown in [Create Azure SQL Database](#Create-a-SQL-Server).
```
lucyzhang929, line 165: Should this be [Create a SQL Database]?

lucyzhang929: Do we also want to update the note in Output Binding Tutorial (line 213)?

lucyzhang929 (approved):

Charles-Gagnon:
@dzsquared Merging this in now so people can get the updates, if you get a chance to look it over just leave any comments here and I'll fix them up later

PR #164, opened 2021-11-12
specifying version

dzsquared:
since --prerelease flag is only available .net core sdk 5 forward.  my bad.


chlafreniere (approved):

lucyzhang929 (approved):

Charles-Gagnon:
Well that's annoying. 

I feel like maybe we should tell users to use the .NET 5+ for development instead then. As long as the project targets netcoreapp3.1 that should be fine.

PR #170, opened 2021-11-17
Build and test on all three platforms for product build pipeline

lucyzhang929:
I also combined `template-steps-build-test-linux.yml` and `template-steps-build-test.yml` into one `template-steps-build-test.yml` file so that we don't have to update two places when making changes to build or test steps.

https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=129175&view=results


builds/azure-pipelines/build-release.yml
```diff
@@ -23,8 +23,39 @@ stages:
...
+
+    steps:
+      - template: 'template-steps-build-test.yml'
+        parameters:
+          solution: '$(solution)'
+          configuration: '$(configuration)'
+          nugetVersion: '$(nugetVersion)'
+          testFilter: '$(testFilter)'
+
+  - job: BuildTestPublishLinux
```
Charles-Gagnon, line 57: I'd just combine this into the job above and then have the publish step only run on linux

Charles-Gagnon (approved):

builds/azure-pipelines/build-release.yml
lucyzhang929: I tried to do that but couldn't figure out how to run the publish template only for linux. I tried adding
`condition: eq(variables['Agent.OS'], 'linux')`
and
`${{ if eq($(Agent.OS), 'linux') }}:`
but both did not work.

Charles-Gagnon: Hm, odd. Ok well go ahead and merge this in now then. It's not a huge deal and we can investigate later if needed

PR #177, opened 2021-11-19
Use global.json for defining build SDKs

Charles-Gagnon:
This way we're ensuring that the SDK used is consistent so everyone has the same environment.

Samples need the .NET 3 SDK since that's what we support running on currently.


Charles-Gagnon:
Sigh, got ahead of myself. Something weird with the signing task is blocking this...

Charles-Gagnon:
Alright, turns out that the signing task will use the repo global.json and was unhappy about that being set to .NET 6...so for now we'll have the top level one be .NET 5 until they can support .NET 6.

lucyzhang929 (approved):

PR #181, opened 2021-11-29
Minor fixes to output binding code

lucyzhang929:
A couple of small suggestions from Shay, an engineer from .NET, who Drew reached out to for advice on securely accessing database in our code.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
             int batchSize = 1000;
             await connection.OpenAsync();
-            SqlCommand command = connection.CreateCommand();
+            using SqlCommand command = connection.CreateCommand();
             SqlTransaction transaction = connection.BeginTransaction();
             command.Connection = connection;
             command.Transaction = transaction;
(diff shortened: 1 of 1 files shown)


src/SqlAsyncCollector.cs
```diff
@@ -151,7 +151,7 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
 
             int batchSize = 1000;
             await connection.OpenAsync();
-            SqlCommand command = connection.CreateCommand();
+            using SqlCommand command = connection.CreateCommand();
```
Charles-Gagnon, line 154: Could you see if there's a rule such as https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213 we could enable to flag these? Generally we want to make sure we're cleaning up disposables

src/SqlAsyncCollector.cs
```diff
@@ -151,7 +151,7 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
 
             int batchSize = 1000;
             await connection.OpenAsync();
```
Charles-Gagnon, line 153: All of the lines after this before the try should go into the try so that we're always sure to close the connection once it's been opened. Currently if an exception occurs while doing the other stuff (unlikely but you never know) then we'll still never close the connection.

Charles-Gagnon (approved):

src/SqlAsyncCollector.cs
Charles-Gagnon: Darn ☹️

PR #183, opened 2021-12-13
Add SqlObject helper class

Charles-Gagnon:
This will make it a lot easier/safer to work with the object names in the code.

It might be a good idea to even require that the parameter to the attribute is of this type - at least for output bindings. Something to keep an eye on as we continue to make improvements.

### DIFF
--- a/src/SqlObject.cs
+++ b/src/SqlObject.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         /// <summary>
         /// Get the schema and object name from the SchemaObjectName.
         /// </summary>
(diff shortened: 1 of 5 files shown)


src/SqlObject.cs
```diff
@@ -0,0 +1,82 @@
...
+    internal class SqlObject
+    {
+        private static readonly string SCHEMA_NAME_FUNCTION = "SCHEMA_NAME()";
+
+        /// <summary>
+        /// The name of the object
+        /// </summary>
+        public readonly string Name;
+        /// <summary>
+        /// The name of the object, quoted and escaped with single quotes if it's not the default SCHEMA_NAME() function
```
lucyzhang929, line 25: I think you got the summaries for QuotedName and QuotedSchema mixed up here

lucyzhang929 (approved):

Issue #37, opened 2020-11-14
Support output binding for auto-incremented primary key


shibayan:
Failed to add records using Output Binding to a table with auto-incrementing id, which is commonly used in SQL DB.

It's just an idea, but it might be a solution if we can tell the Binding side that the key is auto-incremented by adding the `DatabaseGenerated` attribute to the property.

### Repro code

**Table definition**

```sql
CREATE TABLE [dbo].[TodoItems](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[Title] [nvarchar](255) NOT NULL,
	[Body] [nvarchar](max) NOT NULL,
	CONSTRAINT [PK_TodoItems] PRIMARY KEY CLUSTERED ([Id] ASC)
)
```

**Function code**

```csharp
public class TodoItem
{
    // Temporary comment out
    //public int Id { get; set; }
    public string Title { get; set; }
    public string Body { get; set; }
}

public class Function1
{
    [FunctionName("Function1")]
    public async Task<IActionResult> Run(
        [HttpTrigger(AuthorizationLevel.Function, "get")] HttpRequest req,
        [Sql("dbo.TodoItems", ConnectionStringSetting = "SqlConnection")] IAsyncCollector<TodoItem> items,
        ILogger log)
    {
        await items.AddAsync(new TodoItem
        {
            Title = "foo",
            Body = "bar"
        });

        return new OkResult();
    }
}
```

### Expected

`Id` value is generated in an auto-increment and the record is successfully added to the DB.

### Actual

I tried every possible pattern, but the only way I could find was to disable the auto-incremented `Id`.

**Comment out the `Id` property**

> System.Private.CoreLib: Exception while executing function: Function1. Microsoft.Azure.WebJobs.Host: Error while handling parameter items after function returned:. SqlBinding: All primary keys for SQL table 'TodoItems' and schema ''dbo'' need to be found in 'FunctionApp25.TodoItem.' Missing primary keys: [Id].

**Uncommenting the `Id` property**

> System.Private.CoreLib: Exception while executing function: Function1. Microsoft.Azure.WebJobs.Host: Error while handling parameter items after function returned:. Core Microsoft SqlClient Data Provider: Cannot update identity column 'Id'.

mark-quickel:
Confirming this issue as well today after not being able to find a workaround for these identity fields. Auto-incremented identity fields are a common pattern and should be minimally updated as "not supported" in the documentation.

[...]

PR #187, opened 2021-12-28
Handle identity column

Charles-Gagnon:
This is the initial work for https://github.com/Azure/azure-functions-sql-extension/issues/37

With this we now handle an identity column being included in the primary key (either as the sole column or part of a composite primary key). If an identity column is present then we will always generate an insert statement instead - since the column value is generated by the engine and can't be inserted directly.

There's still some other cases I want to revisit for making the experience better. Specifically : 

1. If an identity column is present in the POCO object but it's null/empty we should treat it like it isn't there when constructing the query
2. If an identity column is specified then we should treat it as an update instead of an insert. We might even be able to use the merge statement for this.
2. Investigate if we can somehow return the ID of the created object back to the calling function so that it can be returned as part of the response

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -180,8 +180,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     GenerateDataQueryForMerge(tableInfo, batch, out string newDataQuery, out string rowData);
                     command.CommandText = $"{newDataQuery} {tableInfo.Query};";
                     par.Value = rowData;
-                    Console.WriteLine(command.CommandText); // TODO remove
-                    Console.WriteLine(rowData); // TODO remove
                     await command.ExecuteNonQueryAsync();
                 }
                 transaction.Commit();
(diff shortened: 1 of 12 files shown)


src/SqlAsyncCollector.cs
```diff
@@ -451,6 +478,10 @@ public DynamicPOCOContractResolver(IDictionary<string, string> sqlColumns)
             {
                 // we only want to serialize POCO properties that correspond to SQL columns
                 this._propertiesToSerialize = sqlColumns;
+                foreach (KeyValuePair<string, string> kv in this._propertiesToSerialize)
+                {
+                    Console.WriteLine($"{kv.Key}:{kv.Value}");
```
lucyzhang929, line 483: Was this for debugging?

src/SqlAsyncCollector.cs
```diff
@@ -423,23 +446,27 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                     await sqlConnection.CloseAsync();
                 }
 
-                if (!primaryKeys.Any())
+                if (!primaryKeys.Any(k => k.IsIdentity) && !primaryKeys.Any())
```
lucyzhang929, line 449: Isn't `!primaryKeys.Any()` condition enough to check for primary keys here? Do we need the `!primary.Any(k => k.IsIdentity)` check as well?

lucyzhang929 (approved):

Issue #73, opened 2021-09-23
Investigate SqlConnection Lifecycles


chlafreniere:
- [x] Ensure that we're not closing/reopening connections necessarily
- [x] Ensure that we're not opening multiple connections unnecessarily concurrentlly
- [x] Ensure that all connections that are opened are closed in all codepaths
- [x] Investigate connection pooling

PR #190, opened 2022-01-04
Open connection once for upserts

lucyzhang929:
This PR addresses the first task https://github.com/Azure/azure-functions-sql-extension/issues/73
Previously the connection is opened/closed three times for an upsert
1. Retrieving column information
2. Retrieving primary keys
3. Performing actual upsert

Now the connection will be opened once and used for all three queries. Since the connection is created in a `using` statement, it will be closed and disposed of correctly even if an exception is thrown.

I ran perf tests locally and there were no performance changes.


Charles-Gagnon (approved):
I'd suggest adding a comment to the param docs for the methods that now require the connection be opened already so callers are aware of that.

chlafreniere:
@lucyzhang929 out of curiosity how were perf tests executed?

lucyzhang929:
For the perf tests, I used the Stopwatch class to run the AddProduct output binding integration test and compared the average of 1000 iterations.

chlafreniere:
> For the perf tests, I used the Stopwatch class to time the AddProduct output binding integration test locally and compared the average of 1000 iterations.

Gotcha, thanks. I wonder if Benchmark.NET would be helpful here too. I've been doing some thinking over break about how we can use benchmarking in our tools to prevent perf regressions. I know other teams are looking at this too. Just thinking, that's all 😄

lucyzhang929:
Yep I spent some time looking at setting up Benchmark.NET for perf tests as well but I wasn't able to set it up quickly so I ended up using Stopwatch to get some quick local perf results first. But I do want to continue looking at Benchmark.NET

Issue #133, opened 2021-10-19
Logic doesn't use collation of database for string comparisons


Charles-Gagnon:
Currently all our logic assumes case insensitive names (such as checking column names exist during upsert).

Ideally we should be using the database collation to determine whether we do a case sensitive or case insensitive comparison.

PR #191, opened 2022-01-05
Check database collation for case sensitivity

lucyzhang929:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/133


src/SqlAsyncCollector.cs
```diff
@@ -394,20 +414,40 @@ AS NewData
...
+                    var cmdCollation = new SqlCommand(GetDatabaseCollationQuery(sqlConnection), sqlConnection);
+                    using SqlDataReader rdr = await cmdCollation.ExecuteReaderAsync();
+                    while (await rdr.ReadAsync())
+                    {
+                        caseSensitive = GetCaseSensitivityFromCollation(rdr[Collation].ToString());
+                    }
+                }
+                catch (Exception ex)
+                {
+                    logger.LogInformation($"Encountered exception while retrieving database collation: {ex}. Case insensitive behavior will be used by default.");
```
Charles-Gagnon, line 436: This should probably be a warning

src/SqlAsyncCollector.cs
```diff
@@ -220,12 +222,11 @@ private static IEnumerable<string> GetExtraProperties(IDictionary<string, string
         /// <param name="table">Information about the table we will be upserting into</param>
         /// <param name="rows">Rows to be upserted</param>
         /// <returns>T-SQL containing data for merge</returns>
-        private static void GenerateDataQueryForMerge(TableInformation table, IEnumerable<T> rows, out string newDataQuery, out string rowData)
+        private static void GenerateDataQueryForMerge(TableInformation table, IEnumerable<T> rows, StringComparer comparer, out string newDataQuery, out string rowData)
```
Charles-Gagnon, line 225: The table info already has the comparer so we don't need to pass it in separately right?

Charles-Gagnon (approved):
Please also update the README : https://github.com/Azure/azure-functions-sql-extension/blob/main/README.md?plain=1#L527

src/SqlAsyncCollector.cs
lucyzhang929: Ah yes I can use table.Comparer. Thanks!

lucyzhang929:
@Charles-Gagnon Should we leave the case-sensitive collation as a known issue in the README for now since we do not yet know when our next release will be?

Charles-Gagnon:
Good point. Can you make a follow up issue to remove that once the update is released?

lucyzhang929:
Sure thing - created an issue here https://github.com/Azure/azure-functions-sql-extension/issues/192

PR #196, opened 2022-01-21
Add AppInsights

Charles-Gagnon:
Add AppInsights package and get everything initialized for adding telemetry. No events added in this PR - those will be done in follow-up PRs.

Telemetry can be disabled in two ways : 

- Through environment variable `AZUREFUNCTIONS_SQLBINDINGS_TELEMETRY_OPTOUT`
- Through app setting `AzureFunctionsSqlBindingsTelemetryOptOut`

If either of these are set to "1", "true" or "yes" then telemetry is disabled and nothing is logged. 

We will also show a welcome message on startup (if logging is enabled) for users to see these options : 

image

(Note I disabled IDE0130 for now. It was complaining about the folder structure because of the RootNamespace hack I did when moving project stuff around. I'm going to come back later and see if I can get it fixed for real but for now I want to keep all this stuff in one folder so disabling the rule is the easiest way)


samples/local.settings.json
```diff
@@ -3,6 +3,7 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "SqlConnectionString": ""
-}
+    "SqlConnectionString": "",
+    "AzureFunctionsSqlBindingsTelemetryOptOut": true
```
chlafreniere, line 7: Are we setting this to true to avoid our tests from inflating numbers?

src/Utils.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
chlafreniere, line 2: ideally, we'd have some tests for this file 😄

src/Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// <summary>
+        /// The environment variable used for opting out of telemetry
+        /// </summary>
+        public const string TelemetryOptoutEnvVar = "AZUREFUNCTIONS_SQLBINDINGS_TELEMETRY_OPTOUT";
+        /// <summary>
+        /// The app setting used for opting out of telemetry
+        /// </summary>
+        public const string TelemetryOptoutSetting = "AzureFunctionsSqlBindingsTelemetryOptOut";
+
+        public const string WelcomeMessage = @"SQL Bindings for Azure Functions
```
chlafreniere, line 39: I believe this should be "Azure SQL binding for Azure Functions"

src/Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        private const string EventsNamespace = "azure-functions-sql-bindings";
+        internal static string CurrentSessionId;
+        private TelemetryClient _client;
+        private Dictionary<string, string> _commonProperties;
+        private Dictionary<string, double> _commonMeasurements;
+        private Task _trackEventTask;
+        private ILogger _logger;
+        private bool _initialized;
+        private const string InstrumentationKey = "98697a1c-1416-486a-99ac-c6c74ebe5ebd";
```
chlafreniere, line 29: Just a sanity check. Is this a new ikey?

src/Telemetry/Telemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+            catch (Exception e)
+            {
+                Debug.Fail(e.ToString());
+            }
+        }
+
+        private Dictionary<string, double> GetEventMeasures(IDictionary<string, double> measurements)
+        {
+            var eventMeasurements = new Dictionary<string, double>(this._commonMeasurements);
```
chlafreniere, line 128: Can we do any of these other solutions instead (to clean up this code)? https://stackoverflow.com/questions/6422091/convert-idictionary-to-dictionary

samples/local.settings.json
Charles-Gagnon: Correct

src/Telemetry/Telemetry.cs
Charles-Gagnon: Yes

src/Telemetry/Telemetry.cs
Charles-Gagnon: What are you suggesting? Because there's two things happening here : 

1. We create a clone of the commonMeasurements dictionary (we don't want to modify the original)
2. We then add all the passed in measurements to that dictionary, overwriting any properties that might exist

We want to create a clone of the dictionary regardless (using the dictionary passed in could be problematic, we can't trust the caller to be doing the right thing there). And as far as I know there isn't a way to create a "combined" dictionary from two separate sources. 

Unless you're just saying that the logic is fine you just think we could compress it - in which case maybe? I'd have to play around with it, but that doesn't really seem worth the effort here given how short they are already.

samples/local.settings.json
Charles-Gagnon: Moved this to just be set by the env var when running tests

src/Telemetry/Telemetry.cs
chlafreniere: Oh, I missed line 128 where we're creating a clone of commonMeasurements. I thought this was just cloning in quite a roundabout fashion.

And yeah was mainly around compressing this into fewer lines. It's fine.

chlafreniere (approved):

chlafreniere:
@Charles-Gagnon hopefully the other PR will help with the build failure 😄

PR #198, opened 2022-01-25
Add initial set of telemetry events

Charles-Gagnon:
Added initial set of events - focusing on a few key metrics : 

- Overall and feature-specific usage
- Query performance
- Errors*

*Note that for errors all we capture is the name of exception type (e.g. `InvalidOperationException`). No PII is gathered for any of the events.

List of the non-Error events added : 

- Creation of handlers (one is created once per binding)
- Input conversions being performed (maps to number of invocations of the binding)
- Output binding table info cache hit/miss
- Output binding execution trace and timings - one event for each sub-query (case sensitivity, column definitions, primary keys)
- Output binding upsert perf (including # of batches inserted)


chlafreniere (approved):
Please check the test failure. Looks good otherwise; I appreciate your thoroughness.

Issue #192, opened 2022-01-05
Update README: Remove Case-Sensitive collation from known issues after next release


lucyzhang929:

dzsquared:
(also need to update ms docs to remove the KI at that time)

dzsquared:
just merged in change to the ms docs too

PR #199, opened 2022-01-26
Remove case sensitive known issue

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/192


Charles-Gagnon:
@dzsquared Can you handle removing the MS docs article?

PR #202, opened 2022-01-28
Add machine ID and exception logging

Charles-Gagnon:
Copied and simplified from https://github.com/dotnet/interactive/tree/main/src/Microsoft.DotNet.Interactive.Telemetry

I decided just to generate a random GUID for each user instead of trying to do any sort of MAC hashing - I didn't see the point in doing all the extra work here for that.

Also added some exception logging so we can track if there's issues with initializing the telemetry client.

### DIFF
--- a/src/Telemetry/UserLevelCacheWriter.cs
+++ b/src/Telemetry/UserLevelCacheWriter.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Telemetry
+{
+    public sealed class UserLevelCacheWriter
+    {
+        private const string AzureFunctionsSqlBindingsProfileDirectoryName = ".azurefunctions-sqlbindings";
+        private readonly string _azureFunctionsSqlBindingsTryUserProfileFolderPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), AzureFunctionsSqlBindingsProfileDirectoryName);
+
+        public string RunWithCache(string cacheKey, Func<string> getValueToCache)
+        {
+            string cacheFilepath = this.GetCacheFilePath(cacheKey);
+            try
+            {
+                if (!File.Exists(cacheFilepath))
+                {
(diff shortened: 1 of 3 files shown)


src/Telemetry/UserLevelCacheWriter.cs
```diff
@@ -0,0 +1,54 @@
...
+                    return runResult;
+                }
+                else
+                {
+                    return File.ReadAllText(cacheFilepath);
+                }
+            }
+            catch (Exception ex)
+            {
+                if (ex is UnauthorizedAccessException
```
chlafreniere, line 38: Curious about your thoughts around adding a telemetry event here to track issues in this codepath

Charles-Gagnon: It's kind of a catch-22 situation there because we need these common properties to do logging currently. I could refactor it to send a special event in this case, but I don't see that being something worth the effort at this time unless we think this is something that is more likely to happen.

chlafreniere: Gotcha. Yeah, I'm most worried about deployed scenarios. I just don't know that much about how permissions are setup in Azure, and whether even the user dir is locked down.

Charles-Gagnon: It's something I need to look at anyways - it feels like we need to have a flag for when it's running in Azure. I'm going to investigate what that might look like (hopefully some env var I can use?)

Charles-Gagnon: Looked into it and it was pretty easy to pipe through so I went ahead and added some exception logging here.

src/Telemetry/UserLevelCacheWriter.cs
```diff
@@ -39,13 +46,13 @@ public string RunWithCache(string cacheKey, Func<string> getValueToCache)
                     || ex is PathTooLongException
                     || ex is IOException)
                 {
+                    this._telemetryClient.TrackException(ex);
```
chlafreniere, line 49: Does the actual exception message get sent?

chlafreniere (approved):

src/Telemetry/UserLevelCacheWriter.cs
Charles-Gagnon: (followed up separately)

PR #208, opened 2022-02-08
add [AutoResolve] for reading from app setting

### LINKED ISSUE #200: Add support for app settings binding expressions
Can you please add support for binding expression in SQL command text

MaddyDev:
Adding  [AutoResolve] tag enables reading the %<appKey>% values from app config during run time. This was already added to parameters but missed from CommandText. Adding it to enable reading stored procedure values from app config.
This fixes #200.


Charles-Gagnon (commented):
Please add a sample and test for this

lucyzhang929 (approved):

samples/Common/ProductUtilities.cs
```diff
@@ -19,7 +19,7 @@ public static List<Product> GetNewProducts(int num)
                 var product = new Product
                 {
                     ProductID = i,
-                    Cost = 100,
+                    Cost = i,
```
chlafreniere, line 22: do we have to change this as well?

MaddyDev: not really, but to query I thought it would be better to have different values.

Charles-Gagnon: Let's at least make the cost `100 + i` or something to clearly differentiate it from the ID

samples/InputBindingSamples/GetProductsStoreProcedureFromAppSetting.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.InputBindingSamples
+{
+
+    public static class GetProductsStoredProcedureFromAppSetting
```
Charles-Gagnon, line 13: Please add a test that verifies this route as well

samples/local.settings.json
```diff
@@ -3,6 +3,8 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "SqlConnectionString": ""
+    "SqlConnectionString": "",
+    "Sp_SelectCost": "SelectProductsCost",
+    "Cost100": 100
```
Charles-Gagnon, line 8: Let's try to keep the samples as generic as possible - so for this I'd suggest `ProductCost` instead

samples/InputBindingSamples/GetProductsStoreProcedureFromAppSetting.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        [FunctionName("GetProductsStoredProcedureFromAppSetting")]
+        public static IActionResult Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "getproductscost100-storedprocedure-appsetting")]
```
Charles-Gagnon, line 17: Along with updating the app setting name I'd suggest we name the route something more generic like `getproductsbycost`.

The route really we should just be thinking of from the users point of view - what makes sense for someone calling this API. We have other ways like the function name to indicate to the developer looking at this sample what this sample is demonstrating.

samples/InputBindingSamples/GetProductsStoreProcedureFromAppSetting.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.InputBindingSamples
+{
+
+    public static class GetProductsStoredProcedureFromAppSetting
+    {
+        [FunctionName("GetProductsStoredProcedureFromAppSetting")]
+        public static IActionResult Run(
```
Charles-Gagnon, line 16: Please add a doc comment explaining what this function is demonstrating

Charles-Gagnon (changes requested):

samples/InputBindingSamples/GetProductsStoreProcedureFromAppSetting.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.InputBindingSamples
+{
+    /// <summary>
+    /// This shows an example of a SQL Input binding that uses a store procedure from app setting
```
Charles-Gagnon, line 13:
```
This shows an example of a SQL Input binding that uses a stored procedure from an app setting value to query for Products with a specific cost that is also defined as an app setting value.
```

test/Integration/SqlInputBindingIntegrationTests.cs
```diff
@@ -98,6 +98,25 @@ public async void GetProductsNameEmptyTest(int n, int cost)
             Assert.Equal(expectedResponse, actualResponse, StringComparer.OrdinalIgnoreCase);
         }
 
+        [Fact]
+        public async void GetProductsByCostTest()
+        {
+            this.StartFunctionHost(nameof(GetProductsStoredProcedureFromAppSetting));
+
+            // Generate T-SQL to insert n rows of data with cost
+            Product[] products = GetProductsWithSameCost(1, 100);
```
Charles-Gagnon, line 107: You should also insert some products with cost != 100 to make sure we're only grabbing the expected products

Charles-Gagnon (approved):
Few more cleanup things, but overall looks good!

PR #209, opened 2022-02-14
Add readme to Nuget package and update SDK version

MaddyDev:


docs/README.md
```diff
@@ -0,0 +1,12 @@
+# SQL binding extension for Azure Functions
+
+Azure SQL bindings in Azure Functions supports input and output bindings for the Azure SQL and SQL Server products.
+- **Input Binding**: takes a SQL query to run and returns the output of the query in the function.
+- **Output Binding**: takes a list of rows and upserts them into the user table (i.e. If a row doesn't already exist, it is added. If it does, it is updated).
+
+
```
Charles-Gagnon, line 7: Remove extra space

docs/README.md
```diff
@@ -0,0 +1,12 @@
+# SQL binding extension for Azure Functions
+
+Azure SQL bindings in Azure Functions supports input and output bindings for the Azure SQL and SQL Server products.
```
Charles-Gagnon, line 3: `Azure SQL bindings for Azure Functions adds input and output bindings for Azure SQL and SQL Server products.`

That reads clearer to me - @dzsquared thoughts?

Charles-Gagnon (approved):

lucyzhang929 (approved):

docs/README.md
dzsquared: I'm ok with it either way - the text was likely taken from the docs, which I can update to match the wordsmithing

dzsquared (approved):

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -10,13 +10,15 @@
     <GenerateAssemblyInfo>true</GenerateAssemblyInfo>
     <DebugSymbols>true</DebugSymbols>
     <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+     <!-- Removing SymbolPackageFormat snupkg since it only supports *.pdb and some more
```
Charles-Gagnon, line 13: Don't remove this - we should just not be including the README in the snupkg. We still want to produce the snupkg though.

Charles-Gagnon (changes requested):

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
MaddyDev: from my understanding it still does produce the package but if you mention it, it filters out files with certain extensions only for packaging and md files are filtered so the pr validation is failing since it's not packaged.

Charles-Gagnon: That's a long thread and I'm not really sure the context here...what do you mean the PR validation is failing? Do you have a link to a build with the failure?

If we can't embed the README then it's reasonable to just move this stuff into a nuspec file I think and hardcode the description there.

Charles-Gagnon: The thread mentions there's a fix for it out - we should also be able to update the SDK version we're using as well. As long as we still targeting netstandard2.1 I believe that should be fine.

MaddyDev: This is the failing build: https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=141092&view=logs&j=ccb1d05f-c091-5851-4ae8-18f7fca67013&t=0f04aa1d-081c-58aa-9042-59cde02cc630

Let me check, I've only read till the part that said to remove the snupkg.

Charles-Gagnon (approved):

Issue #210, opened 2022-02-15
Review audit logging


Charles-Gagnon:
Ensure that we're logging appropriate amount of information to allow customers to audit the data requests being made and results returned.

PR #211, opened 2022-02-16
Add logs for # of rows queried and upserted

lucyzhang929:
fixes: https://github.com/Azure/azure-functions-sql-extension/issues/210

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -214,6 +214,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     { TelemetryMeasureName.CommandDurationMs.ToString(), commandSw.ElapsedMilliseconds }
                 };
                 TelemetryInstance.TrackEvent(TelemetryEventName.UpsertEnd, props, measures);
+                this._logger.LogInformation($"Upserted {rows.Count()} row(s) into database: {connection.Database} and table: {fullTableName}.");
             }
             catch (Exception ex)
             {
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -11,7 +11,6 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using static Microsoft.Azure.WebJobs.Extensions.Sql.Telemetry.Telemetry;
-using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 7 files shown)


src/SqlConverters.cs
```diff
@@ -68,16 +70,20 @@ internal class SqlGenericsConverter<T> : IAsyncConverter<SqlAttribute, IEnumerab
...
             /// <param name="configuration"></param>
+            /// <param name="loggerFactory">Logger Factory for creating an ILogger</param>
             /// <exception cref="ArgumentNullException">
             /// Thrown if the configuration is null
             /// </exception>
-            public SqlGenericsConverter(IConfiguration configuration)
+            public SqlGenericsConverter(IConfiguration configuration, ILoggerFactory loggerFactory)
             {
                 this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+                this._logger = loggerFactory?.CreateLogger(LogCategories.Bindings) ?? throw new ArgumentNullException(nameof(loggerFactory));
```
Charles-Gagnon, line 86: Instead of passing in the logger factory and creating a new logger for each converter I'd suggest we instead create the logger in the config provider and pass that in directly.

src/SqlConverters.cs
```diff
@@ -158,6 +164,7 @@ public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attri
                 await connection.OpenAsync();
                 var dataTable = new DataTable();
                 adapter.Fill(dataTable);
+                this._logger.LogInformation($"{dataTable.Rows.Count} row(s) queried from database: '{connection.Database}' using Command: '{command.CommandText}'");
```
Charles-Gagnon, line 167: Here you're adding single quotes but you didn't do that above. We should be consistent - and I'd probably say the quotes are unnecessary here.

Charles-Gagnon (changes requested):

src/SqlAsyncCollector.cs
```diff
@@ -214,6 +214,7 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                     { TelemetryMeasureName.CommandDurationMs.ToString(), commandSw.ElapsedMilliseconds }
                 };
                 TelemetryInstance.TrackEvent(TelemetryEventName.UpsertEnd, props, measures);
+                this._logger.LogInformation($"Upserted {rows.Count()} row(s) into database: {connection.Database} and table: {fullTableName}.");
```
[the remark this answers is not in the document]
lucyzhang929: The exceptions themselves have detailed messages and are shown in the logs. Do we want to add logging in addition to that?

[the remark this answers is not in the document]
Charles-Gagnon: That sounds fine to me!

MaddyDev (approved):

src/SqlConverters.cs
```diff
@@ -68,16 +69,20 @@ internal class SqlGenericsConverter<T> : IAsyncConverter<SqlAttribute, IEnumerab
...
             /// <param name="configuration"></param>
+            /// <param name="logger">ILogger used to log information and warnings</param>
             /// <exception cref="ArgumentNullException">
             /// Thrown if the configuration is null
             /// </exception>
-            public SqlGenericsConverter(IConfiguration configuration)
+            public SqlGenericsConverter(IConfiguration configuration, ILogger logger)
             {
                 this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+                this._logger = logger;
```
ilmax, line 85: null check?

lucyzhang929: I don't think it's needed since we create the logger in the provider and we have a null check for the logger factory there.

PR #214, opened 2022-03-02
Centrally manage packages

Charles-Gagnon:
Easier to maintain and ensure we use the same versions for all our projects. 

As part of this I had to specify a version for these three packages : 

Microsoft.Azure.WebJobs
Microsoft.Data.SqlClien
Microsoft.SourceLink.GitHub

which previously were using * versioning. In general specifying the exact version should be preferred anyways to ensure that we control exactly what we're building and don't end up with situations where different build locations end up with different dependent package versions.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,17 @@
+<Project>
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="4.0.3" />
+    <PackageVersion Include="Microsoft.NET.Sdk.Functions" Version="3.0.13" />
+    <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.17.0" />
+    <PackageVersion Include="Microsoft.Azure.WebJobs" Version="3.0.31" />
+    <PackageVersion Include="Microsoft.Data.SqlClient" Version="3.0.1" />
+    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
+    <PackageVersion Include="morelinq" Version="3.3.2" />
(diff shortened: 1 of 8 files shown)


Directory.Packages.props
```diff
@@ -0,0 +1,17 @@
+<Project>
+  <ItemGroup>
```
chlafreniere, line 2: Which of these packages aren't using the latest versions?

MaddyDev (approved):

MaddyDev:
@chlafreniere Do I need to update the EO Compliance sheet?

chlafreniere:
@MaddyDev taking offline

Directory.Packages.props
Charles-Gagnon: Latest versions, and if there was a newer major version included both latest of the current major version + latest major version.

| Package | Current Version | Latest Version |
| --------- | ----------------- | --------------- |
|Microsoft.Azure.WebJobs.Extensions.Storage | 4.0.3 | 4.0.5/5.0.0 |
|Microsoft.NET.Sdk.Functions | 3.0.13 | latest/4.1.0 |
|Microsoft.ApplicationInsights | 2.17.0 | 2.20.0 |
|Microsoft.Azure.WebJobs | 3.0.31 | latest |
|Microsoft.Data.SqlClient | 3.0.1 | latest/4.1.0 |
|Microsoft.SourceLink.GitHub | 1.1.1 | latest |
|morelinq | 3.3.2 | latest |
|System.Runtime.Caching | 4.7.0 | latest/6.0.0 |
|Microsoft.NET.Test.Sdk | 17.0.0 | 17.1.0 |
|Moq | 4.14.3 | 4.17.1 |
|xunit | 2.4.0 | 2.4.1 |
|xunit.runner.visualstudio | 2.4.0 | 2.4.3 |

PR #216, opened 2022-03-03
Make Assembly Version Match NuGet Version

### LINKED ISSUE #213: Assembly version does not match package version
The assembly version seems to always be 1.0.0.0. This should instead match the package version when doing a full product build.

chlafreniere:
Fixes #213.

Wondering aloud about breaking change possibilities, however. Curious on thoughts here.

Before:
image

After:
image

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -47,4 +47,5 @@ stages:
         solution: '$(solution)'
         configuration: '$(configuration)'
         nugetVersion: '1.0.0-test'
+        binariesVersion: '1.0.0'
         testFilter: '$(testFilter)'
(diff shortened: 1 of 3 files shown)


Charles-Gagnon (approved):

lucyzhang929 (approved):

MaddyDev (approved):

chlafreniere:
@Charles-Gagnon do you have any concerns around the major version being different here? What specific testing can we do to ensure that we're not regressing any mainline scenarios?

@lucyzhang929 are there any places where we refer to the sql bindings dll + file version? Thinking about things like templates + vs code experience.

Charles-Gagnon:
I can't think of any place that cares about the assembly version. Everything is by the nuget package version.

To test you can get a private build and then reference that version (and set up nuget to pull from local file location or just stick directly into nuget cache) and make a new extension proj.

chlafreniere:
Thanks @Charles-Gagnon. Tested it locally and it worked just fine. Will merge on Monday unless I hear any objections from our friend @lucyzhang929.

PR #221, opened 2022-04-05
Add SQL Trigger support

AmeyaRele:


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,1048 @@
...
+        /// Resets the in-memory state of the change monitor and sets it to start polling for changes again.
+        /// </summary>
+        /// <param name="error">
+        /// The error messages the logger will report describing the reason function execution failed (used only in the case of a failure)
+        /// </param>
+        /// <param name="acquireLock">
+        /// True if ClearRows should acquire the _rowsLock (only true in the case of a failure)
+        /// </param>
+        /// <returns></returns>
+        private async Task ClearRows(string error, bool acquireLock)
```
ilmax, line 454:
```suggestion
        private async Task ClearRowsAsync(string error, bool acquireLock)
```

src/SqlBindingUtilities.cs
```diff
@@ -161,6 +161,99 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
             return result;
         }
 
+        /// <summary>
+        /// Returns [tableName] if tableName is not prefixed by a schema, otherwise returns [schema].[table] in the case that
+        /// tableName = schema.table
+        /// </summary>
+        /// <param name="tableName">The name of the user's table</param>
+        /// <returns>The normalized table name</returns>
+        public static string NormalizeTableName(string tableName)
```
Charles-Gagnon, line 170: You should use https://github.com/Azure/azure-functions-sql-extension/blob/main/src/SqlObject.cs for parsing object names

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,1048 @@
...
+        public const int BatchSize = 10;
+        public const int MaxDequeueCount = 5;
+        public const int MaxLeaseRenewalCount = 5;
+        public const int LeaseIntervalInSeconds = 30;
+        public const int PollingIntervalInSeconds = 5;
+
+        private readonly string[] variableLengthTypes = new string[] { "varchar", "nvarchar", "nchar", "char", "binary", "varbinary" };
+        private readonly string[] variablePrecisionTypes = new string[] { "numeric", "decimal" };
+
+        private readonly string _workerId;
```
JatinSanghvi, line 40: Check if the number of mutable fields in the class can be reduced, for example by passing some of them between functions, and accepting output of function via its return value rather than through shared class member variables.

JatinSanghvi:
This PR was closed as it was replaced with https://github.com/Azure/azure-functions-sql-extension/pull/226 as we changed the source branch to `triggerbindings` in this repo. Having a common branch would let both me and Ameya be able to push new commits. The intention was also to give credit to other contributors in that branch, but it seems those other commits do not show up as there were intermediate merges that occurred between `dev` and `triggerbindings` branches.

Issue #145, opened 2021-11-03
Figure out a better way to know when Azure Function Host is started for integration tests


lucyzhang929:
Currently we're waiting 5 seconds, but sometimes the host is not started within 5 seconds, causing failures in our pipelines: https://github.com/Azure/azure-functions-sql-extension/blob/bb550e83d2280f05c16f8431929358f6fd5a1143/test/Integration/IntegrationTestBase.cs#L184

PR #222, opened 2022-04-05
Wait for Function host to fully start in Integration tests

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/145

There still seems to be some other issues with sending the HTTP requests, will investigate those separately.

https://github.com/Azure/azure-functions-sql-extension/issues/223


test/Integration/IntegrationTestBase.cs
```diff
@@ -189,7 +188,19 @@ protected void StartFunctionHost(string functionName, bool useTestFolder = false
             this.FunctionHost.BeginOutputReadLine();
             this.FunctionHost.BeginErrorReadLine();
 
-            Thread.Sleep(10000);     // This is just to give some time to func host to start, maybe there's a better way to do this (check if port's open?)
+            var taskCompletionSource = new TaskCompletionSource<bool>();
```
Charles-Gagnon, line 191: We don't actually need the bool, but the framework version we target doesn't have a non-generic TaskCompletionSource so we have to specify something.

lucyzhang929 (approved):

PR #225, opened 2022-04-12
Capture SqlException Number

chlafreniere:
Extract SqlException number to help disambiguate SqlExceptions so that we can better fix customer issues with SqlClient.

I'm certainly unfamiliar here, so please feel free to take a hard look here. Should I still be adding the telemetry property if it's an empty string?


Charles-Gagnon (approved):

MaddyDev (approved):

Charles-Gagnon:
Merging this in to include in the new release.

PR #226, opened 2022-04-19
Add trigger binding support

AmeyaRele:


JatinSanghvi:
@Charles-Gagnon - I do not have access to the build pipeline to check the failure logs. I can login to https://dev.azure.com/mssqltools and see the DevOps projects underneath but do not know which project contains the CI pipeline. Clicking on the [build-result link](https://mssqltools.visualstudio.com/ae14e11c-7eb2-46af-b588-471e6116d635/_build/results?buildId=150541) above shows 401 page.

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -56,7 +57,7 @@ steps:
   inputs:
     command: build
     projects: '${{ parameters.solution }}'
-    arguments: '--configuration ${{ parameters.configuration }} -p:GeneratePackageOnBuild=false'
+    arguments: '--configuration ${{ parameters.configuration }} -p:GeneratePackageOnBuild=false -p:Version=${{ parameters.binariesVersion }}'
```
Charles-Gagnon, line 60: You seem to have some commits that are already in main. Please make sure you've rebased and merged in commits from main correctly so they don't show up.

src/SqlBindingUtilities.cs
```diff
@@ -135,30 +136,11 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
-        /// Used to determine the columns of the table as well as the next SQL row to process
-        /// </param>
-        /// <param name="cols">
-        /// Filled with the columns of the table if empty, otherwise assumed to be populated
-        /// with their names already (used for cacheing so we don't retrieve the column names every time)
-        /// </param>
+        /// <param name="reader">Used to determine the columns of the table as well as the next SQL row to process</param>
         /// <returns>The built dictionary</returns>
-        public static Dictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader, List<string> cols)
+        public static IReadOnlyDictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader)
```
Charles-Gagnon, line 141: Please make generic "cleanup" changes like this and the public -> internal sealed in a separate PR. They don't seem to be related to the feature and should be reviewed separately.

src/SqlBindingUtilities.cs
```diff
@@ -136,11 +135,30 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
+        /// <param name="reader">
+        /// Used to determine the columns of the table as well as the next SQL row to process
+        /// </param>
+        /// <param name="cols">
+        /// Filled with the columns of the table if empty, otherwise assumed to be populated
+        /// with their names already (used for cacheing so we don't retrieve the column names every time)
+        /// </param>
         /// <returns>The built dictionary</returns>
-        public static IReadOnlyDictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader)
+        public static Dictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader, List<string> cols)
```
JatinSanghvi, line 146: I think we should leave this change as it is also affecting trigger-binding logic.

src/TriggerBinding/SqlChange.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Specifies the type of change that occurred to the row.
+        /// </summary>
+        public SqlChangeOperation Operation { get; }
+
+        /// <summary>
+        /// A copy of the row that was updated/inserted in the user's table.
+        /// In the case that the row no longer exists in the user's table, Data is only populated with the primary key values
+        /// of the deleted row.
+        /// </summary>
+        public T Item { get; }
```
Charles-Gagnon, line 40: Is there a way for users to see what the item was like before the change? (i.e. having both a `OldItem` and `NewItem` property)

README.md
```diff
@@ -522,6 +586,62 @@ This changes if one of the primary key columns is an identity column though. In
...
+- **ConnectionStringSetting**: Specifies the name of the app setting that contains the SQL connection string used to connect to a database. The connection string must follow the format specified [here](https://docs.microsoft.com/dotnet/api/microsoft.data.sqlclient.sqlconnection.connectionstring?view=sqlclient-dotnet-core-2.0).
+
+The following are valid binding types for trigger
+
+- **IReadOnlyList<SqlChange\<T\>>**: Each element is a `SqlChange`, which stores change metadata about a modified row in the user table as well as the row itself. In the case that the row was deleted, only the primary key values of the row are populated. The user table row is represented by `T`, where `T` is a user-defined POCO, or Plain Old C# Object. `T` should follow the structure of a row in the queried table. See the [Query String](#query-string) section for an example of what `T` should look like. The two fields of a `SqlChange` are the `Item` field of type `T` which stores the row, and the `Operation` field of type `SqlChangeOperation` which indicates the type of operaton done to the row (either an insert, update, or delete).
```
Charles-Gagnon, line 625: Why are only the primary key values populated when an item is deleted?

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        private async Task<string> GetUserFunctionIdAsync()
+        {
+            string hostId = await this._hostIdProvider.GetHostIdAsync(CancellationToken.None);
+
+            var methodInfo = (MethodInfo)this._parameter.Member;
+            string functionName = $"{methodInfo.DeclaringType.FullName}.{methodInfo.Name}";
+
+            using (var sha256 = SHA256.Create())
```
Charles-Gagnon, line 143: What is this logic doing - did you copy it from somewhere?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+
+        /// <summary>
+        /// Gets the column names of the user's table.
+        /// </summary>
+        private async Task<IReadOnlyList<string>> GetUserTableColumnsAsync(SqlConnection connection, CancellationToken cancellationToken)
+        {
+            string getUserTableColumnsQuery = $@"
+                SELECT name
+                FROM sys.columns
+                WHERE object_id = OBJECT_ID(N'{this._userTableName}', 'U');
```
Charles-Gagnon, line 234: Is this escaping single quotes in the table name?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+
+        /// <summary>
+        /// Inserts row for the user table and the function inside the global state table, if one does not already exist.
+        /// </summary>
+        private async Task InsertGlobalStateTableRowAsync(SqlConnection connection, SqlTransaction transaction, int userTableId, CancellationToken cancellationToken)
+        {
+            string insertRowGlobalStateTableQuery = $@"
+                IF NOT EXISTS (
+                    SELECT * FROM {SqlTriggerConstants.GlobalStateTableName}
+                    WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {userTableId}
```
Charles-Gagnon, line 299: Is this function ID guaranteed not to have single quotes in it? (you should probably escape it regardless just to be safe)

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+        {
+            this.StopAsync(CancellationToken.None).Wait();
+        }
+
+        /// <summary>
+        /// Disposes resources held by the listener to poll for changes.
+        /// </summary>
+        public void Dispose()
+        {
+            // Nothing to dispose.
```
Charles-Gagnon, line 69: Shouldn't you cancel the listener when this happens?

test/Integration/IntegrationTestBase.cs
```diff
@@ -324,17 +342,6 @@ public void Dispose()
             {
                 this.TestOutput.WriteLine($"Failed to close connection. Error: {e1.Message}");
             }
-            try
```
Charles-Gagnon, line 327: Why are you changing these?

test/Unit/SqlTriggerBindingTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            Assert.Throws<ArgumentNullException>(() => new SqlTriggerBinding<TestData>(tableName, connectionString, TriggerBindingFunctionTest.GetParam(), hostIdProvider.Object, null));
+        }
+
+        [Fact]
+        public async void TestTriggerBindingProviderTryCreateAsync()
+        {
+            var triggerBindingProviderContext = new TriggerBindingProviderContext(TriggerBindingFunctionTest.GetParam(), cancellationTokenSource.Token);
+            var triggerAttributeBindingProvider = new SqlTriggerAttributeBindingProvider(config.Object, hostIdProvider.Object, loggerFactory.Object);
+
+            await Assert.ThrowsAsync<InvalidOperationException>(() => triggerAttributeBindingProvider.TryCreateAsync(triggerBindingProviderContext));
```
Charles-Gagnon, line 75: Make sure that the test name indicates what is being tested - and if needed more details about what to expect as comments in the test code itself. For example for this I don't know what it means to test "TryCreateAsync" and why it's expected to throw here.

test/Unit/SqlTriggerBindingTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+        private static class TriggerBindingFunctionTest
```
Charles-Gagnon, line 77: Add comments explaining what this class is and for each method explain what it specifically is used for

src/TriggerBinding/SqlChange.cs
AmeyaRele: I don't believe so as change tracking has a clean up period, and the change tracking table only stores the primary key values.

README.md
AmeyaRele: This is because of the change tracking table, which stores only primary key values.

src/TriggerBinding/SqlTriggerBinding.cs
AmeyaRele: The logic is used to have a unique ID for identifying user function. This was inspired from TimerTrigger code
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/01b1a826351bdd1883100d2af2470f1bde88ae5c/src/WebJobs.Extensions/Extensions/Timers/Bindings/TimerTriggerBinding.cs#L39-L40

test/Integration/IntegrationTestBase.cs
AmeyaRele: Changed the order for function host to be killed before the database is dropped. As the trigger periodically checks for changes in the database, the database was not able to be dropped previously.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    public class SqlTriggerBindingIntegrationTests : IntegrationTestBase
+    {
+        public SqlTriggerBindingIntegrationTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        /// <summary>
+        /// Tests for insertion of products triggering the function.
+        /// </summary>
+        [Fact]
```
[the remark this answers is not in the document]
Charles-Gagnon: @AmeyaRele FYI - discussed this with my team and for now let's plan on just adding C# support in this PR and then can follow up with other languages later. So what you have is ok for now (although generally we'd prefer that all tests are made as Theories and then use the UnsupportedLanguages attribute to indicate what languages they don't support)

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,775 @@
...
+                            using (SqlCommand updateTablesPostInvocationCommand = this.BuildUpdateTablesPostInvocation(connection, transaction, newLastSyncVersion))
+                            {
+                                await updateTablesPostInvocationCommand.ExecuteNonQueryAsync(token);
+                            }
+
+                            transaction.Commit();
+                        }
+                        catch (Exception ex)
+                        {
+                            Console.WriteLine("Commit Exception Type: {0}", ex.GetType());
```
Charles-Gagnon, line 455: Don't use Console.WriteLine in product code - use the logger.

src/TriggerBinding/SqlTriggerListener.cs
AmeyaRele: I do think so, as it's a hash, but have escaped it regardless during initialization of function ID.

dzsquared:
reminder: this PR needs to be retargeted to a feature branch for trigger support and not merged to main

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: The WebJobs SDK calls either `StopAsync` or `Cancel` on the `SqlTriggerListener` which in turn stops the change-monitor. I can see extensions like RabbitMQ having empty `Dispose` method while [EventHub calling `StopAsync` from inside both `Cancel` and `Dispose`](https://github.com/Azure/azure-sdk-for-net/blob/a5f2ba22c34a6d931b3717aa36f1c286c4647d13/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubListener.cs#L65-L95). I will update the code following this same pattern.

src/TriggerBinding/SqlTriggerBinding.cs
JatinSanghvi: Added the explanation in the method summary.

src/SqlBindingUtilities.cs
```diff
@@ -134,30 +134,11 @@ public static SqlCommand BuildCommand(SqlAttribute attribute, SqlConnection conn
...
-        /// Used to determine the columns of the table as well as the next SQL row to process
-        /// </param>
-        /// <param name="cols">
-        /// Filled with the columns of the table if empty, otherwise assumed to be populated
-        /// with their names already (used for cacheing so we don't retrieve the column names every time)
-        /// </param>
+        /// <param name="reader">Used to determine the columns of the table as well as the next SQL row to process</param>
         /// <returns>The built dictionary</returns>
-        public static Dictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader, List<string> cols)
+        public static IReadOnlyDictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader)
```
Charles-Gagnon, line 139: Is this something directly related to your PR? In general this kind of cleanup was something I was hoping we'd get around to eventually so I'm not opposed to looking at doing this, but it's general enough that I'd rather it be split into its own PR.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// is triggered (whether or not the trigger is successful).
+        /// </summary>
+        public void Stop()
+        {
+            this._cancellationTokenSourceCheckForChanges.Cancel();
+        }
+
+        public void Dispose()
+        {
+            throw new NotImplementedException();
```
Charles-Gagnon, line 136: Is this needed? And shouldn't dispose be calling stop?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+    /// Periodically polls SQL's change table to determine if any new changes have occurred to a user's table.
+    /// </summary>
+    /// <remarks>
+    /// Note that there is no possiblity of SQL injection in the raw queries we generate. All parameters that involve
+    /// inserting data from a user table are sanitized. All other parameters are generated exclusively using information
+    /// about the user table's schema (such as primary key column names), data stored in SQL's internal change table, or
+    /// data stored in our own worker table.
+    /// </remarks>
+    /// <typeparam name="T">A user-defined POCO that represents a row of the user's table</typeparam>
+    internal sealed class SqlTableChangeMonitor<T> : IDisposable
```
Charles-Gagnon, line 29: Please add debug logging to all this stuff like we do in https://github.com/Azure/azure-functions-sql-extension/blob/main/src/SqlAsyncCollector.cs

src/TriggerBinding/SqlTriggerAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        /// <param name="tableName">The name of the table to monitor for changes</param>
+        public SqlTriggerAttribute(string tableName)
+        {
+            this.TableName = tableName ?? throw new ArgumentNullException(nameof(tableName));
+        }
+
+        /// <summary>
+        /// The name of the app setting where the SQL connection string is stored
+        /// (see https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection?view=sqlclient-dotnet-core-2.0).
```
Charles-Gagnon, line 28: Remove en-us from all user-visible links

src/SqlBindingUtilities.cs
JatinSanghvi: The change to method `BuildDictionaryFromSqlRow` simplifies the way it gets called from `SqlTableChangeMonitor.cs`. I am okay with moving this change to a separate PR if that works.

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// The listener
+        /// </returns>
+        public async Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context), "Missing listener context");
+            }
+
+            string userFunctionId = SqlBindingUtilities.AsSingleQuoteEscapedString(await this.GetUserFunctionIdAsync());
```
JatinSanghvi, line 121: `SqlBindingUtilities.AsSingleQuoteEscapedString` is called twice, once inside `GetUserFunctionIdAsync()` and once in the caller. Both are unnecessary, as method `GetUserFunctionIdAsync` will return a string with lower-case hex characters like `"36a6cc45b7bd4c83"`.

JatinSanghvi:
I will push new commits to PR branch by around mid of next week. I will try to cover unit and integration tests by then. Not pushing commits here regularly so as to avoid attracting attention from reviewers.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// </exception>
+        private async Task<IReadOnlyList<(string name, string type)>> GetPrimaryKeyColumnsAsync(SqlConnection connection, CancellationToken cancellationToken)
+        {
+            string getPrimaryKeyColumnsQuery = $@"
+                SELECT c.name, t.name, c.max_length, c.precision, c.scale
+                FROM sys.indexes AS i
+                INNER JOIN sys.index_columns AS ic ON i.object_id = ic.object_id AND i.index_id = ic.index_id
+                INNER JOIN sys.columns AS c ON ic.object_id = c.object_id AND ic.column_id = c.column_id
+                INNER JOIN sys.types AS t ON c.user_type_id = t.user_type_id
+                WHERE i.is_primary_key = 1 AND i.object_id = OBJECT_ID(N{this._userTable.QuotedName}, 'U');
```
Charles-Gagnon, line 180: This needs to be `QuotedFullName`, otherwise you could get the wrong one if there's multiple tables with that name with different schemas.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// <summary>
+        /// Returns the OBJECT_ID of userTable
+        /// </summary>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown if the query to retrieve the OBJECT_ID of the user table fails to correctly execute
+        /// This can happen if the OBJECT_ID call returns NULL, meaning that the user table might not exist in the database
+        /// </exception>
+        private async Task<int> GetUserTableIdAsync(SqlConnection connection, CancellationToken cancellationToken)
+        {
+            string getObjectIdQuery = $"SELECT OBJECT_ID(N{this._userTable.QuotedName}, 'U');";
```
Charles-Gagnon, line 139: `QuotedFullName`

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            using (var getObjectIdCommand = new SqlCommand(getObjectIdQuery, connection))
+            {
+                using (SqlDataReader reader = await getObjectIdCommand.ExecuteReaderAsync(cancellationToken))
+                {
+
+                    // TODO: Check if the below if-block ever gets hit.
```
Charles-Gagnon, line 146: Has this TODO been done?

It shouldn't ever be hit, but doesn't hurt to be explicit. Would prefer a better error message though so it's clear what the problem is if a user files an issue with this message.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <param name="newLastSyncVersion">The new LastSyncVersion to store in the _globalStateTable for this _userTableName</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildUpdateTablesPostInvocation(SqlConnection connection, SqlTransaction transaction, long newLastSyncVersion)
+        {
+            string workerTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col} = w.{col}"));
+
+            // TODO: Need to think through all cases to ensure the query below is correct, especially with use of < vs <=.
```
Charles-Gagnon, line 707: Has this been done?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                // the state is checking for changes and not renew the (just released) leases.
+                await this.ClearRowsAsync(string.Empty, false);
+            }
+        }
+
+        /// <summary>
+        /// Calculates the new version number to attempt to update LastSyncVersion in global state table to. If all
+        /// version numbers in _rows are the same, use that version number. If they aren't, use the second largest
+        /// version number. For an explanation as to why this method was chosen, see 9c in Steps of Operation in this
+        /// design doc: https://microsoft-my.sharepoint.com/:w:/p/t-sotevo/EQdANWq9ZWpKm8e48TdzUwcBGZW07vJmLf8TL_rtEG8ixQ?e=owN2EX.
```
Charles-Gagnon, line 489: Remove this - not public doc (and then add the relevant parts in comments below. The description shouldn't be in the summary since that's an implementation detail)

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                                this._logger.LogError("  Message: {0}", ex2.Message);
+                            }
+                        }
+                    }
+                }
+
+            }
+            catch (Exception e)
+            {
+                // What should we do if releasing the leases fails? We could try to release them again or just wait,
```
Charles-Gagnon, line 472: A retry here would be preferable - if it fails after 3ish retries then the default behavior of the lease expiring will have to do.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                {
+                    await connection.OpenAsync(token);
+
+                    while (!token.IsCancellationRequested)
+                    {
+                        await this._rowsLock.WaitAsync(token);
+
+                        await this.RenewLeasesAsync(connection, token);
+
+                        // Want to make sure to renew the leases before they expire, so we renew them twice per lease period.
```
Charles-Gagnon, line 326: I'm not sure just twice per period is enough - it's essentially just 1 since the other will be happening as the lease is expiring so could result in another worker coming in between when the lease expires and the 2nd renewal happens (in case the first failed for some reason).

At least 4 per period would be a lot safer.

src/SqlBindingUtilities.cs
Charles-Gagnon: @JatinSanghvi Please merge in latest from main so these changes don't show up.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+
+        /// <summary>
+        /// Builds the query to check for changes on the user's table (<see cref="RunChangeConsumptionLoopAsync()"/>).
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransaction transaction)
+        {
+            string selectList = string.Join(", ", this._userTableColumns.Select(col => this._primaryKeyColumns.Contains(col) ? $"c.{col}" : $"u.{col}"));
```
lucyzhang929, line 583: If a column was deleted from the user table, would this query fail? In that case would the user have to restart the function in order for userTableColumns to be updated?

README.md
```diff
@@ -85,6 +90,18 @@ ALTER TABLE ['{table_name}'] ALTER COLUMN ['{primary_key_column_name}'] int 
...
+3. If you plan to use the trigger support, you need to enable [change tracking](https://docs.microsoft.com/sql/relational-databases/track-changes/about-change-tracking-sql-server) on the SQL database and the SQL table. Please note that enabling change tracking will add to the cost of the SQL server.
+
+```sql
+ALTER DATABASE ['your database name']
+SET CHANGE_TRACKING = ON
+(CHANGE_RETENTION = 2 DAYS, AUTO_CLEANUP = ON)
+
+ALTER TABLE ['your table name']
+ENABLE CHANGE_TRACKING
+WITH (TRACK_COLUMNS_UPDATED = ON)
```
lucyzhang929, line 102: Since we are returning all columns for an update operation and not using SYS_CHANGE_COLUMNS, TRACK_COLUMNS_UPDATED does not need to be on right?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    /// <typeparam name="T">A user-defined POCO that represents a row of the user's table</typeparam>
```
Charles-Gagnon, line 17: Make this into an actual xml-doc comment and add a description for the class

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+                    using (SqlTransaction transaction = connection.BeginTransaction(System.Data.IsolationLevel.RepeatableRead))
+                    {
+                        await CreateSchemaAsync(connection, transaction, cancellationToken);
+                        await CreateGlobalStateTableAsync(connection, transaction, cancellationToken);
+                        await this.InsertGlobalStateTableRowAsync(connection, transaction, userTableId, cancellationToken);
+                        await CreateWorkerTablesAsync(connection, transaction, workerTableName, primaryKeyColumns, cancellationToken);
+                        transaction.Commit();
+                    }
+
+                    // TODO: Check if passing cancellation token would be beneficial.
```
Charles-Gagnon, line 99: I would assume we need to do this to properly handle cancellations from the host - is this something you're planning on fixing?

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: Although you could also probably simplify this by using the table ID retrieved in the call before this directly.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+                    string[] variableLengthTypes = new string[] { "varchar", "nvarchar", "nchar", "char", "binary", "varbinary" };
+                    string[] variablePrecisionTypes = new string[] { "numeric", "decimal" };
+
+                    var primaryKeyColumns = new List<(string name, string type)>();
+
+                    while (await reader.ReadAsync(cancellationToken))
+                    {
+                        string type = reader.GetString(1);
+
+                        if (variableLengthTypes.Contains(type))
```
Charles-Gagnon, line 197: Can we guarantee these will always be lower case? Would suggest lower casing the value regardless to be safe.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+
+        /// <summary>
+        /// Gets the column names of the user's table.
+        /// </summary>
+        private async Task<IReadOnlyList<string>> GetUserTableColumnsAsync(SqlConnection connection, CancellationToken cancellationToken)
+        {
+            string getUserTableColumnsQuery = $@"
+                SELECT name
+                FROM sys.columns
+                WHERE object_id = OBJECT_ID(N{this._userTable.QuotedName}, 'U');
```
Charles-Gagnon, line 234: Same here with using QuotedFullName (or just the object ID directly)

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+            _ = primaryKeyColumns ?? throw new ArgumentNullException(nameof(primaryKeyColumns));
+            _ = executor ?? throw new ArgumentNullException(nameof(executor));
+            _ = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            this._connectionString = connectionString;
+            this._userTableId = userTableId;
+            this._userTable = userTable;
+            this._userFunctionId = userFunctionId;
+            this._workerTableName = workerTableName;
+            this._userTableColumns = primaryKeyColumns.Concat(userTableColumns.Except(primaryKeyColumns)).ToList();
```
Charles-Gagnon, line 95: Is this extra logic actually necessary? If the user table columns already has the primary key columns (which I would expect they do) why do we need to remove them and then concat them back on?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+
+                using (var connection = new SqlConnection(this._connectionString))
+                {
+                    await connection.OpenAsync(token);
+
+                    while (!token.IsCancellationRequested)
+                    {
+                        if (this._state == State.CheckingForChanges)
+                        {
+                            // What should we do if this call gets stuck?
```
Charles-Gagnon, line 161: What does "stuck" here mean? If it's taking a long time to execute then it will eventually time out and throw an exception so it should be handled just like any other exception that occurs when running these queries.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                                using (SqlCommand acquireLeasesCommand = this.BuildAcquireLeasesCommand(connection, transaction))
+                                {
+                                    await acquireLeasesCommand.ExecuteNonQueryAsync(token);
+                                }
+                            }
+                            transaction.Commit();
+                        }
+                        catch (Exception ex)
+                        {
+                            this._logger.LogError("Commit Exception Type: {0}", ex.GetType());
```
Charles-Gagnon, line 238: Let's give this a better log message, something like `Exception getting list of changes for table...`

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                            this._logger.LogError("  Message: {0}", ex.Message);
+                            try
+                            {
+                                transaction.Rollback();
+                            }
+                            catch (Exception ex2)
+                            {
+                                this._logger.LogError("Rollback Exception Type: {0}", ex2.GetType());
```
Charles-Gagnon, line 246: Same here, better message would be something like `Exception rolling back transaction after error occurred getting list of changes for table...`

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        private async Task ProcessChangesAsync(CancellationToken token)
+        {
+            if (this._rows.Count > 0)
+            {
+                this._state = State.ProcessingChanges;
+                IReadOnlyList<SqlChange<T>> changes = null;
+
+                try
+                {
+                    // What should we do if this fails? It doesn't make sense to retry since it's not a connection based
```
Charles-Gagnon, line 271: Make sure to document this behavior in the design doc

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+            if (acquireLock)
+            {
+                this._logger.LogError(error);
+                await this._rowsLock.WaitAsync();
+            }
+
+            this._leaseRenewalCount = 0;
+            this._state = State.CheckingForChanges;
+            this._rows = new List<IReadOnlyDictionary<string, string>>();
+            this._rowsLock.Release();
```
Charles-Gagnon, line 415: Shouldn't this only be done if `acquireLock` is true?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// </summary>
+        /// <param name="error">
+        /// The error messages the logger will report describing the reason function execution failed (used only in the case of a failure).
+        /// </param>
+        /// <param name="acquireLock">True if ClearRowsAsync should acquire the "_rowsLock" (only true in the case of a failure)</param>
+        private async Task ClearRowsAsync(string error, bool acquireLock)
+        {
+            if (acquireLock)
+            {
+                this._logger.LogError(error);
```
Charles-Gagnon, line 408: Just move the logging to before the ClearRowsAsync call - the only time we pass in a string is when acquireLock is true anyways so there's no reason to have that done in here.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        private long RecomputeLastSyncVersion()
+        {
+            var changeVersionSet = new SortedSet<long>();
+            foreach (IReadOnlyDictionary<string, string> row in this._rows)
+            {
+                string changeVersion = row["SYS_CHANGE_VERSION"];
+                changeVersionSet.Add(long.Parse(changeVersion, CultureInfo.InvariantCulture));
+            }
+
+            // If there are at least two version numbers in this set, return the second highest one. Otherwise, return
```
Charles-Gagnon, line 500: Why the second highest?

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: This piece of code must be added since the class `SqlTableChangeMonitor` otherwise gets flagged by CodeAnalysis as it owns disposable resources of type `CancellationTokenSource` but does not implement `IDisposable` itself. I have replaced `Stop()` method with `Dispose()` for now as there is no requirement posed by WebJobs SDK for change-monitor class to have method specifically with same `Stop`. Please reopen the conversation if you have a better solution for this. Other way is to add suppression for the warning message or surround the code by `#pragma disable` directive, but I see at no place in the rest of project are we suppressing messages other than 'unused parameters'.

JatinSanghvi: A few notes: In general, most other extensions have the logic that listens to new events entirely encoded inside the `*Listener` class. In SQL extension, we have the main `SqlTriggerListener` class and its extension `SqlTableChangeMonitor` class. The reason is that we require a couple of variables (e.g., `userTableId') that need to be initialized after the listener is started. But these variables then need to be set to mutable which will get error-prone considering that the listener logic contains concurrency. So, we kept all one-time initialization steps inside `SqlTriggerListener`, got the local variables initialized, and passed them to `SqlTableChangeMonitor` which stores them in read-only variables.

Earlier we started having similar method definitions inside both classes i.e., `StartAsync` and `StopAsync` but since `StartAsync` calls `SqlTableChangeMonitor`'s constructor, having `StopAsync` call it's `Dispose` method kinds of feels consistent to me.

By the way, I will be updating the `Dispose` method in outer `SqlTriggerListener` class to do nothing like in [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/48fe14f0a373a3a279849eae0cf0b3756fa3d47f/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerListener.cs#L85-L88) as from the implementation, it seems that WebJobs SDK would be always call either of `StopAsync` or `Cancel` or both.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: I would prefer to not have this additional condition. I tried a couple of ways to hit the if-block, but it was either `connection.Open()` or `ExecuteReaderAsync()` that failed. There are a few other locations in the repo where `ReadAsync` method gets called but they do not get special treatment. If the method fails in any case, we would still be encountering exception, just not the one of our choice.

JatinSanghvi: On second thought, we should be taking into account all return values before moving ahead. I will update the exception message.

README.md
```diff
@@ -401,8 +458,6 @@ public static async Task<IActionResult> Run(
 
 The output binding takes a list of rows to be upserted into a user table. If the primary key value of the row already exists in the table, the row is interpreted as an update, meaning that the values of the other columns in the table for that primary key are updated. If the primary key value does not exist in the table, the row is interpreted as an insert. The upserting of the rows is batched by the output binding code.
 
-  > **NOTE:** By default the Output binding uses the T-SQL [MERGE](https://docs.microsoft.com/sql/t-sql/statements/merge-transact-sql) statement which requires [SELECT](https://docs.microsoft.com/sql/t-sql/statements/merge-transact-sql#permissions) permissions on the target database.
```
Charles-Gagnon, line 404: Was this removed on purpose?

README.md
```diff
@@ -524,6 +579,64 @@ This changes if one of the primary key columns is an identity column though. In
 1. If the identity column isn't included in the output object then a straight insert is always performed with the other column values. See [AddProductWithIdentityColumn](./samples/samples-csharp/OutputBindingSamples/AddProductWithIdentityColumn.cs) for an example.
 2. If the identity column is included (even if it's an optional nullable value) then a merge is performed similar to what happens when no identity column is present. This merge will either insert a new row or update an existing row based on the existence of a row that matches the primary keys (including the identity column). See [AddProductWithIdentityColumnIncluded](./samples/samples-csharp/OutputBindingSamples/AddProductWithIdentityColumnIncluded.cs) for an example.
 
+### Trigger Binding
```
Charles-Gagnon, line 582: Add a statement indicating that currently this is only supported for C# functions.

README.md
Charles-Gagnon: There's more information about that below, but it just says that this functionality isn't currently used by the trigger. Because of that I think we should probably leave this out of the samples/docs so users aren't needlessly enabling it.

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// We call the WebJobs SDK library method to generate the host ID. The host ID is essentially a hash of the
+        /// assembly name containing the user function(s). This ensures that if the user ever updates their application,
+        /// unless the assembly name is modified, the new application version will be able to resume from the point
+        /// where the previous version had left. Appending another hash of class+method in here ensures that if there
+        /// are multiple user functions within the same process and tracking the same SQL table, then each one of them
+        /// gets a separate view of the table changes.
+        /// </summary>
+        private async Task<string> GetUserFunctionIdAsync()
+        {
+            string hostId = await this._hostIdProvider.GetHostIdAsync(CancellationToken.None);
```
Charles-Gagnon, line 84: Are there any public docs on this function? I don't see any here : https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.host?view=azure-dotnet

But from your comment and looking at https://github.com/Azure/azure-webjobs-sdk/tree/dev/src/Microsoft.Azure.WebJobs.Host/Executors - does this mean that if a user deploys a the same function (with trigger binding) to two separate function apps that the host ID will be the same? Which means they'd share the same worker tables and state - is that the intended behavior?

src/TriggerBinding/SqlTriggerBindingProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            ParameterInfo parameter = context.Parameter;
+            SqlTriggerAttribute attribute = parameter.GetCustomAttribute<SqlTriggerAttribute>(inherit: false);
+
+            if (attribute == null)
+            {
+                return Task.FromResult(default(ITriggerBinding));
```
Charles-Gagnon, line 43: Shouldn't this throw? This doesn't seem like a valid scenario to return null here.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+                FROM {SqlTriggerConstants.GlobalStateTableName}
+                WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {this._userTableId};
+
+                SELECT TOP {BatchSize}
+                    {selectList},
+                    c.SYS_CHANGE_VERSION, c.SYS_CHANGE_OPERATION,
+                    w.ChangeVersion, w.AttemptCount, w.LeaseExpirationTime
+                FROM CHANGETABLE (CHANGES {this._userTable.FullName}, @last_sync_version) AS c
+                LEFT OUTER JOIN {this._workerTableName} AS w WITH (TABLOCKX) ON {workerTableJoinCondition}
+                LEFT OUTER JOIN {this._userTable.FullName} AS u ON {userTableJoinCondition}
```
Charles-Gagnon, line 594: `this._userTable.BracketQuotedFullName`

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+                DECLARE @last_sync_version bigint;
+                SELECT @last_sync_version = LastSyncVersion
+                FROM {SqlTriggerConstants.GlobalStateTableName}
+                WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {this._userTableId};
+
+                SELECT TOP {BatchSize}
+                    {selectList},
+                    c.SYS_CHANGE_VERSION, c.SYS_CHANGE_OPERATION,
+                    w.ChangeVersion, w.AttemptCount, w.LeaseExpirationTime
+                FROM CHANGETABLE (CHANGES {this._userTable.FullName}, @last_sync_version) AS c
```
Charles-Gagnon, line 592: `this._userTable.BracketQuotedFullName`

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+            string updateTablesPostInvocationQuery = $@"
+                DECLARE @current_last_sync_version bigint;
+                SELECT @current_last_sync_version = LastSyncVersion
+                FROM {SqlTriggerConstants.GlobalStateTableName}
+                WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {this._userTableId};
+
+                DECLARE @unprocessed_changes bigint;
+                SELECT @unprocessed_changes = COUNT(*) FROM (
+                    SELECT c.SYS_CHANGE_VERSION
+                    FROM CHANGETABLE(CHANGES {this._userTable.FullName}, @current_last_sync_version) AS c
```
Charles-Gagnon, line 712: `this._userTable.BracketQuotedFullName`

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+        /// <summary>
+        /// Builds the query to check for changes on the user's table (<see cref="RunChangeConsumptionLoopAsync()"/>).
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransaction transaction)
+        {
+            string selectList = string.Join(", ", this._userTableColumns.Select(col => this._primaryKeyColumns.Contains(col) ? $"c.{col}" : $"u.{col}"));
+            string userTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col} = u.{col}"));
```
Charles-Gagnon, line 579: Column names should be bracket quoted

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+
+        /// <summary>
+        /// Builds the query to check for changes on the user's table (<see cref="RunChangeConsumptionLoopAsync()"/>).
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransaction transaction)
+        {
+            string selectList = string.Join(", ", this._userTableColumns.Select(col => this._primaryKeyColumns.Contains(col) ? $"c.{col}" : $"u.{col}"));
```
Charles-Gagnon, line 578: Column names should be bracket quoted

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+        /// the user's function. If the global version number is updated, also cleans the worker table and removes all
+        /// rows for which ChangeVersion <= newLastSyncVersion.
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <param name="newLastSyncVersion">The new LastSyncVersion to store in the _globalStateTable for this _userTableName</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildUpdateTablesPostInvocation(SqlConnection connection, SqlTransaction transaction, long newLastSyncVersion)
+        {
+            string workerTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col} = w.{col}"));
```
Charles-Gagnon, line 700: Column names should be bracket quoted

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// </summary>
+        private static async Task CreateWorkerTablesAsync(
+            SqlConnection connection,
+            SqlTransaction transaction,
+            string workerTableName,
+            IReadOnlyList<(string name, string type)> primaryKeyColumns,
+            CancellationToken cancellationToken)
+        {
+            string primaryKeysWithTypes = string.Join(", ", primaryKeyColumns.Select(col => $"{col.name} {col.type}"));
+            string primaryKeys = string.Join(", ", primaryKeyColumns.Select(col => col.name));
```
Charles-Gagnon, line 337: Bracket quote names here too

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            // TODO: Check if some of the table columns below can have 'NOT NULL' property.
+            string createWorkerTableQuery = $@"
```
Charles-Gagnon, line 340: It would be good to check the primary key column names and make sure there's none that match the ones we defined as well - and if there are throw a error that indicates exactly what the problem is (otherwise that would be difficult for a user to find out)

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Fixed. Now using table ID queried from here in the other two queries inside class `SqlTriggerListener`.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Still pending. The comment must be added by me when I did not know about the design document. I will be able to check this tomorrow.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Thanks for point out. Dropping non-primary-key columns should result in query failure Restarting the user application should re-populate the column list and fix the error. I do not know of a way to resolve this problem as querying table schema every time would be too expensive.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: This is done to have all primary key columns before all non-primary-key columns. I am not sure, but it might be required in one of the SQL queries. I will check this out.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: We should probably call out in the docs that any modification to the table schema will result in undefined behavior for any triggers currently watching for changes - that'll cover any such scenarios.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: It's documented in the design doc. There must still be more pending items with the highest version, so the second highest version is safe one for cleanup. Though there's an interesting bug noted in the design document for not updating the global version to the highest one, when an insert (v1 to v2) and an update (v2 to v3) in rare case might get treated as two inserts (v1 to v2 and v1 to v3).

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Alrighty, if it ends up being necessary make sure to add a comment explaining that

README.md
JatinSanghvi: I will fix this. I must have overlooked it while merging.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Please add info about why this is done to the comment - we want to be able to know why unintuitive stuff like this was done without having to dig up the design doc for everything.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                // If there's an exception in any part of the process, we want to clear all of our data in memory and
+                // retry checking for changes again.
+                this._rows = new List<IReadOnlyDictionary<string, string>>();
```
MaddyDev, line 252: Is there a reason for re-initializing the list rather than just clearing them?

src/TriggerBinding/SqlTriggerBinding.cs
JatinSanghvi: This is the best effort approach. The class behind `IHostProvider` is `DefaultHostIdProvider`. As can be seen in [its implementation](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultHostIdProvider.cs#L56) hashes the name of the assembly that contains user functions. This ensures that if user modifies and rebuilds their application, until the assembly's name is not updated, the worker table and global state is reused. If this is not a plane user application but a function apps, the functions host overwrites the `DefaultHostIdProvider` with its [own implementation](https://github.com/Azure/azure-functions-host/blob/bfecdc29f78cd29cb6a8d01181d9f24ca79fe3a8/src/WebJobs.Script/Host/ScriptHostIdProvider.cs) which takes into account more parameters and updates the parameter set based on where the function app is hosted (on Azure or inside Kubernetes cluster).

src/TriggerBinding/SqlTriggerBindingProvider.cs
JatinSanghvi: This is how it is implemented in other extensions. Two examples: [Cosmos DB](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/48fe14f0a373a3a279849eae0cf0b3756fa3d47f/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTriggerAttributeBindingProvider.cs#L45-L48), [Service Bus](https://github.com/Azure/azure-functions-servicebus-extension/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerAttributeBindingProvider.cs#L54-L57).

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: I will need to recheck on this to confirm. Since `this._rows` is shared among both concurrent loops, if the other loop is processing the list referenced by `_rows`, it would be safer to not touch that list by instead start with a fresh one.

src/TriggerBinding/SqlTriggerBinding.cs
Charles-Gagnon: Good to know about the web host ID, thanks.

There's a couple concerns/questions I have : 

1. If a user wants to have the trigger start "fresh" (picking up events from the current state going forward) then they have to make sure to clear out existing worker tables/global state. But it's not clear how they would know what to do - the IDs used don't really help indicate which one belongs to which instance so the user might end up deleting something they didn't mean to. 
2. Trigger instances might end up sharing state that isn't intended. If I have two applications running that are essentially the same but let's say have a different config controlling what they do with the events then they'll end up sharing an ID, right? That doesn't seem ideal since there isn't any way to work around that. Maybe not a common scenario but what if we at least documented a config value that could be set to manually control the ID in case the user wants to either guarantee uniqueness (or specifically re-use the same state)

How do existing triggers work - are they similar in that they try to keep the state for the trigger across multiple instances? I would have initially thought that the SQL trigger would always just start listening for events as soon as it started - not caring about picking up past events or anything. Is all this state logic here mostly meant for handling scaling (where my understanding is that there might be multiple instances of the same worker running in parallel) or is there other functionality that this is intended for?

src/TriggerBinding/SqlTriggerBindingProvider.cs
Charles-Gagnon: Do you know why? I guess maybe it's because of the "Try" part of "TryCreateAsync" - is this method called for all trigger providers when a trigger binding attribute is detected or something and returning null is just saying "I don't know how to handle this type of trigger binding"? It would be great to document that here to make it clear why this is different from every other check we do where we throw.

src/TriggerBinding/SqlTriggerBinding.cs
Charles-Gagnon: Once we decide on what the correct behavior is make sure to document this clearly in the public docs as well - users need to know: 

1. In what specific cases state will be shared - be as explicit as possible with how this is determined
2. How to clean up state for an instance manually

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,769 @@
...
+                        {
+                            this._logger.LogError("Commit Exception Type: {0}", ex.GetType());
+                            this._logger.LogError("  Message: {0}", ex.Message);
+                            try
+                            {
+                                transaction.Rollback();
+                            }
+                            catch (Exception ex2)
+                            {
+                                this._logger.LogError("Rollback Exception Type: {0}", ex2.GetType());
```
MaddyDev, line 457: Follow Charles's above suggestion for all logging statements (something like Exception releasing leases on rows for table...)

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// Creates the worker table associated with the user's table, if one does not already exist.
+        /// </summary>
+        private static async Task CreateWorkerTablesAsync(
+            SqlConnection connection,
+            SqlTransaction transaction,
+            string workerTableName,
+            IReadOnlyList<(string name, string type)> primaryKeyColumns,
+            CancellationToken cancellationToken)
+        {
+            string primaryKeysWithTypes = string.Join(", ", primaryKeyColumns.Select(col => $"{col.name} {col.type}"));
```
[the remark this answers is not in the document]
JatinSanghvi: This is fixed. I have verified the functionality with table and column names containing Unicode characters and spaces and of maximum allowed name-lengths.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,356 @@
...
+                    using (SqlTransaction transaction = connection.BeginTransaction(System.Data.IsolationLevel.RepeatableRead))
+                    {
+                        await CreateSchemaAsync(connection, transaction, cancellationToken);
+                        await CreateGlobalStateTableAsync(connection, transaction, cancellationToken);
+                        await this.InsertGlobalStateTableRowAsync(connection, transaction, userTableId, cancellationToken);
+                        await CreateWorkerTablesAsync(connection, transaction, workerTableName, primaryKeyColumns, cancellationToken);
+                        transaction.Commit();
+                    }
+
+                    // TODO: Check if passing the cancellation token would be beneficial.
```
Charles-Gagnon, line 104: This seems like something we should be doing

README.md
JatinSanghvi: Done in line 585.

README.md
JatinSanghvi: Removed `TRACK_COLUMNS_UPDATED` option from Readme and code samples.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Did not use constants for column names since the query texts too contain literal values for column names. Having constant variables for column names inside the query strings will possibly degrade code readability.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: The rows in `sys.tables` should always have values under `Name` column in lower case at least for the system types. I will be updating the earlier check for conflicting column names to be case insensitive in next commit.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: I will log the column names and worker table name in `SqlTriggerListener.cs` so those get logged even if there is a failure before `SqlTableChangeMonitor` instance gets constructed. User table name and function ID are already getting logged in `SqlTriggerListener.cs`.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: The usage of `<=` is correct. Changes whose versions equal `newLastSyncVersion` should also be counted as unprocessed. I will remove the comment. Though I still need to take some time to wrap my head completely around why the last-sync version computation is the way it is.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Created #282 to track add retries if releasing the lease fails.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Thought the comment says otherwise, the code intends to have enough time to get the lease renewed once only before it expired. It assumes that it will always succeed if the worker is up and running. I will update the `LeaseIntervalInSeconds` to 60 seconds and renewal interval to 15 seconds.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <returns>The SqlCommand populated with the query and appropriate parameters</returns>
+        private SqlCommand BuildRenewLeasesCommand(SqlConnection connection)
+        {
+            var renewLeasesQuery = new StringBuilder();
+
+            for (int index = 0; index < this._rows.Count; index++)
+            {
+                renewLeasesQuery.Append($@"
+                    UPDATE {this._workerTableName} WITH (TABLOCKX)
```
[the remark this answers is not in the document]
JatinSanghvi: I have added this task in #282 for tracking.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+                                var rows = new List<IReadOnlyDictionary<string, string>>();
+                                using (SqlDataReader reader = await getChangesCommand.ExecuteReaderAsync(token))
+                                {
+                                    while (await reader.ReadAsync(token))
+                                    {
+                                        rows.Add(SqlBindingUtilities.BuildDictionaryFromSqlRow(reader));
+                                    }
+                                }
+
+                                this._rows = rows;
```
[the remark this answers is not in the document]
JatinSanghvi: If I get you comment right, I think rethrow is missing here. The outermost `catch` in this method is cleaning up `this._rows`. I will need some time to check if I can simplify the try-catch blocks. I will park this one for later. Tracking it with #282

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -0,0 +1,774 @@
...
+            catch (Exception e)
+            {
+                // If there's an exception in any part of the process, we want to clear all of our data in memory and
+                // retry checking for changes again.
+                this._rows = new List<IReadOnlyDictionary<string, string>>();
+                this._logger.LogWarning($"Failed to check {this._userTable.FullName} for new changes due to error: {e.Message}");
+            }
+        }
+
+        private async Task ProcessChangesAsync(CancellationToken token)
```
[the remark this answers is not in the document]
JatinSanghvi: I may be missing something here, but won't the item get re-picked after the current lease expires? The worker will need to definitely acquire lease before processing and there's no easy way to revert the changes made to the worker-table if the processing fails so if the item gets available again after lease period expires, perhaps that is an acceptable behavior to live with?

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: The ordering of columns is no longer needed, probably because of the changes we made to how query text is constructed. I will simplify the statement.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: I will remove the comment as there is handling added for this. Currently the `MaxRenewalCount` was 5 and taking into account 15 seconds or lease renewal interval, the lease will be release after 75 seconds. I am updating the `MaxRenewalCount` to 10 as 75 seconds seem to short. @Charles-Gagnon, in your comment, do you mean we should have a different strategy for dealing with functions that are stuck?

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Tracking with #283.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: None of these are my changes so I can only lookup and guess. Here, the code tries to maintain that whenever `acquireLock` is passed as `false`, the lock is already acquired so that it is not tried to be re-acquired which would result in a deadlock (?). In this case, the first line inside method `ReleaseLeasesAsync` has acquired the lock, so it's safe to update the shared variables.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Will fix this. This is probably done to save the number of locations where error is logged and `acquireLock` variable has nothing to do with logging.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Moving this inside #282. I will add comment once I completely understand the intricacies of last sync version selection and associated behaviors.

src/TriggerBinding/SqlTriggerBinding.cs
JatinSanghvi: Tracking this in #283

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: This is correct. I have deliberately defined `this._rows` to be `IReadOnlyList<IReadOnlyDictionary<string, string>>` so we can only call read-only method on the list pointed by it. We cannot modify the list by reassign new list to `this._rows`. This ensures the updating the list is an atomic operation.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Leaving the TODO comment for now and adding the task to #282.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Done. @Charles-Gagnon, please review the updated renewal intervals.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Done in Readme file under 'Known Issues' section.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: I haven't added the 'after error occurred ...' part as that would already be logged in preceding log message.

src/TriggerBinding/SqlTriggerBindingProvider.cs
JatinSanghvi: Here's the explanation: During startup, the WebJobs host calls [`CompositeTriggerBindingProvider.TryAsync`](https://github.com/Azure/azure-webjobs-sdk/blob/46514f23a5b5f3cc885d97a5c634f7f80e92a42e/src/Microsoft.Azure.WebJobs.Host/Triggers/CompositeTriggerBindingProvider.cs#L18-L30) which calls `TryAsync` on all the trigger binding providers registered for all the user function parameters. The first provider to return non-null binding is the one that is chosen whose `BindAsync` method will be called to construct the parameter value before every function invocation. Hence, the binding provider is supposed to return null if it does not find the trigger-attribute of its choice on the function parameter, so that next registered binding provider can get a chance.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Update on previous comment: The check for conflicting column name is case insensitive now.

src/TriggerBinding/SqlTableChangeMonitor.cs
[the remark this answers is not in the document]
Charles-Gagnon: I think the main problem is that we update the change version when we take out the lease instead of only after it's been processed.

This is something we should follow up on - please make an issue to track this and we can take a deeper look at it later.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Looks better, thanks. Could consider making it user-configurable later on to give users more direct control over the performance but this should be fine for now.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Yeah, it'll deadlock if called a second time. 

Make an issue to look into this please - at the very least I'd like to try and refactor this slightly to make this easier to understand. All this locking stuff I want to go over anyways since right now it seems kind of arbitrary where the locks are acquired/released which could easily lead to issues later on.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Sounds good, that's definitely something that we need to spend more time going over in detail too.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: I don't know about a different strategy - but we should make sure we handle that scenario deliberately. Right now will it actually get "stuck" or does `ITriggeredFunctionExecutor.TryExecuteAsync` have some sort of timeout that will throw an exception?

In general that would usually imply user error so I'm less concerned about handling that immediately - but this would be another good item to add as a follow-up to investigate.

src/TriggerBinding/SqlTriggerBindingProvider.cs
Charles-Gagnon: Yeah, sorta makes sense (although I'm curious why they don't have the binding attribute itself be used to determine which provider to use....).

Could you add a comment about that? Given that I've been unable to find any public docs about that behavior that seems like a useful thing to have to ensure that no one accidently removes it.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,366 @@
...
+
+                var primaryKeyColumns = new List<(string name, string type)>();
+
+                while (await reader.ReadAsync(cancellationToken))
+                {
+                    string name = reader.GetString(0);
+
+                    if (reservedColumnNames.Contains(name))
+                    {
+                        throw new InvalidOperationException($"Found reserved column name: '{name}' in table: '{this._userTable.FullName}'.");
```
Charles-Gagnon, line 205: 1. Let's make sure this is documented in the README
2. I'd suggest adding a bit more text to tell the user how to resolve it, I think it's a bit unclear right now what the user can do about it. Something like `This column must be removed or renamed for the table to be used by this trigger.`
3. I'd also make it find all of the columns that are reserved and list them, not just the first one

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -0,0 +1,366 @@
...
+                INNER JOIN sys.index_columns AS ic ON i.object_id = ic.object_id AND i.index_id = ic.index_id
+                INNER JOIN sys.columns AS c ON ic.object_id = c.object_id AND ic.column_id = c.column_id
+                INNER JOIN sys.types AS t ON c.user_type_id = t.user_type_id
+                WHERE i.is_primary_key = 1 AND i.object_id = {userTableId};
+            ";
+
+            using (var getPrimaryKeyColumnsCommand = new SqlCommand(getPrimaryKeyColumnsQuery, connection))
+            using (SqlDataReader reader = await getPrimaryKeyColumnsCommand.ExecuteReaderAsync(cancellationToken))
+            {
+                string[] reservedColumnNames = new string[] { "ChangeVersion", "AttemptCount", "LeaseExpirationTime" };
```
Charles-Gagnon, line 193: I know you said you didn't use constants because of code readability, but I don't like the potential risk of regression here by having these values be separate. So I'd strongly prefer that we make constants for each of these - even if it does decrease the readability of the query slightly.

Charles-Gagnon (approved):
Still a few comments to take a look at but overall looks good enough to me. 

Although I want there to be a lot more debug logging done in all the parts. I'm not seeing any in SqlTableChangeMonitor for example. In general here are the things I'll be looking for : 

1. Message at the start of every task and sub task (`Getting changes from table`, `Updating state tables`, `Getting list of changes`, `Got changed rows {....}`) and so on. The goal should be to be able to tell exactly where something went wrong and what it was doing when an issue occurs, regardless of where that was. 
2. Log queries executed and how long they took
3. Extra logging around any of the locking (include thread IDs if that isn't included by default)
4. Make sure all errors are logged with any extra information that may be helpful in determining cause of the error

@lucyzhang929 @MaddyDev Please help make sure that we get as complete of a logging story as we can - we want to have as much as we can now before we give this to customers.

Also we should make sure to get basic usage telemetry set up - we can get some amount of stuff from the AF folks but that won't include people running locally so it'd be good to make sure we have a complete story ourselves as well. I'll follow up separately on my thoughts for that to get started.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Created #286 for this. It is not clear to me right now how to not share `_this.rows` between get-changes and process-changes loops. This might require some redesigning (maybe only have the second loop manage the leases and versions and the first loop get and process rows?).

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Just to make sure there is no disconnect, I will be updating all SQL queries to use constants for column names in worker table. That was what I mentioned in the other conversation.

JatinSanghvi:
> Although I want there to be a lot more debug logging done in all the parts. I'm not seeing any in SqlTableChangeMonitor for example. In general here are the things I'll be looking for :
> 
> 1. Message at the start of every task and sub task (`Getting changes from table`, `Updating state tables`, `Getting list of changes`, `Got changed rows {....}`) and so on. The goal should be to be able to tell exactly where something went wrong and what it was doing when an issue occurs, regardless of where that was.
> 2. Log queries executed and how long they took
> 3. Extra logging around any of the locking (include thread IDs if that isn't included by default)
> 4. Make sure all errors are logged with any extra information that may be helpful in determining cause of the error

Created #287 for this one.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: This is resolved in new PR #288

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: This one requires discussion before I push any change. I have added it to #282 to ensure this gets discussed and work gets done.

src/TriggerBinding/SqlTriggerBindingProvider.cs
JatinSanghvi: Done in #288.

JatinSanghvi:
Closing this pull request as it need not be merged soon and need not be reviewed further. When the branch needs to be merged into main branch, it can be either merged directly or through a new PR.

PR #229, opened 2022-04-21
javascript language support changes and basic samples

MaddyDev:
Changes for javascript language support for SQL Bindings needs to have converters between JObject and whatever types we support. 
- Added IAsyncConverter<SqlAttribute, JArray> for input bindings to return the data as JArray.
- Updated the  code to support JObject parsing for output bindings.

As part of adding node samples in the future, moved the csharp samples to samples-csharp folder under samples and added samples-js to include node samples. We can further add samples-"language" as and when we add new languages support.


Directory.Packages.props
```diff
@@ -12,5 +12,7 @@
     <PackageVersion Include="Moq" Version="4.14.3" />
     <PackageVersion Include="xunit" Version="2.4.0" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageVersion Include="Microsoft.Azure.WebJobs.Extensions.Sql" Version="0.1.223-preview" />
```
chlafreniere, line 15: do we need this here? seems weird to have this in the root

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +310,38 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
                     }
-                    combinedPrimaryKey.Append(value.ToString());
                 }
-
-                // If we have already seen this unique primary key, skip this update
-                if (uniqueUpdatedPrimaryKeys.Add(combinedPrimaryKey.ToString()))
-                {
-                    // This is the first time we've seen this particular PK. Add this row to the upsert query.
+                else {
+                    // add check for duplicate primary keys.
```
Charles-Gagnon, line 338: I'm not sure how this comment relates to the code below

src/SqlBindingConfigProvider.cs
```diff
@@ -54,8 +55,27 @@ public void Initialize(ExtensionConfigContext context)
...
+            inputOutputRule.BindToCollector<SQLObjectOpenType>(typeof(SqlAsyncCollectorBuilder<>), this._configuration, logger);
             inputOutputRule.BindToInput<OpenType>(typeof(SqlGenericsConverter<>), this._configuration, logger);
         }
     }
+
+    internal class SQLObjectOpenType : OpenType.Poco
+    {
+        public override bool IsMatch(Type type, OpenTypeMatchContext context)
+        {
+            if (type.IsGenericType
```
Charles-Gagnon, line 67: Please comments explaining what this code is doing

samples/samples-js/AddProduct/function.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "name": "$return",
+      "type": "http",
+      "direction": "out"
+    },
+    {
+      "name": "product",
+      "type": "sql",
+      "direction": "out",
+      "commandText": "[TestDB].[dbo].[Products]",
```
chlafreniere, line 22: I needed to change this locally. Do we hardcode TestDB anywhere else?

src/SqlAsyncCollector.cs
```diff
@@ -95,7 +96,19 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
                 TelemetryInstance.TrackEvent(TelemetryEventName.AddAsync);
                 try
                 {
-                    this._rows.Add(item);
+                    if (item is string)
+                    {
+                        this._rows.Add(JsonConvert.DeserializeObject<T>(item.ToString()));
```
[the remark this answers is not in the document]
chlafreniere: This may be a weird compiler thing. @MaddyDev if you remove the ToString() does the compiler complain at you?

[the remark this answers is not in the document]
MaddyDev: yes, it's a compiler thing, it's not recognizing item as string.

src/SqlBindingConfigProvider.cs
MaddyDev: To be honest, I don't really know what this code exactly does, there are no comments on the CosmosDB repro either (from where I added this). I added a comment on why I added this, I'll check with them and add further comments.

Charles-Gagnon: It's best to never add code that we have no idea what it does, so yes please follow up and make sure we have a clear understanding of what's happening here.

src/SqlConverters.cs
```diff
@@ -184,6 +185,34 @@ public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attri
...
+
+            /// <summary>
+            /// Opens a SqlConnection, reads in the data from the user's database, and returns it as a list of POCOs.
+            /// </summary>
+            /// <param name="attribute">
+            /// Contains the information necessary to establish a SqlConnection, and the query to be executed on the database
+            /// </param>
+            /// <param name="cancellationToken">The cancellationToken is not used in this method</param>
+            /// <returns>JArray containing the rows read from the user's database in the form of the user-defined POCO</returns>
+            async Task<JArray> IAsyncConverter<SqlAttribute, JArray>.ConvertAsync(SqlAttribute attribute, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
MaddyDev: I tried adding Public but it showed me the below error: The modifier 'public' is not valid for this item.

samples/samples-js/AddProduct/function.json
MaddyDev: good catch, yeah it was my dbname, removed it!

samples/samples-js/AddProduct/sample.dat
```diff
@@ -0,0 +1,5 @@
+{
+    "id": "2000",
```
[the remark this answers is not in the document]
MaddyDev: removed it.

samples/samples-csharp/OutputBindingSamples/TimerTriggerProducts.cs
```diff
@@ -21,7 +21,7 @@ public static class TimerTriggerProducts
             [TimerTrigger("*/5 * * * * *")] TimerInfo req, ILogger log,
             [Sql("Products", ConnectionStringSetting = "SqlConnectionString")] ICollector<Product> products)
         {
-            int totalUpserts = 1000;
+            int totalUpserts = 100;
```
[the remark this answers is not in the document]
MaddyDev: updated it for local testing and forgot to undo it, will do! Thanks

Directory.Packages.props
MaddyDev: nope not needed, removed it, thanks

src/SqlAsyncCollector.cs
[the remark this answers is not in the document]
Charles-Gagnon: Well I'm pretty confused then - the type of item is `T`, which means that the DeserializeObject here is converting a string to...a string? Or is T actually something else at this point and the parameter typing isn't correct? Because if that's the case shouldn't the type of item be `object` or something more generic?

Anyways it would be better just to cast it directly (`item as string`)- calling ToString on a string is pretty odd.

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,57 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
         /// </summary>
         /// <param name="columns"> The columns of the table to upsert to </param>
+        /// <param name="rows"> The rows to be upserted </param>
         /// <returns>List of property names that don't exist in the table</returns>
-        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns)
+        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns, IEnumerable<T> rows)
         {
+            if (typeof(T) == typeof(JObject))
+            {
+                var jsonObj = JObject.Parse(rows.First().ToString());
```
Charles-Gagnon, line 261: What if there aren't any rows returned?

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,57 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
         /// to upsert to and returns the extra property names in a List.
         /// </summary>
         /// <param name="columns"> The columns of the table to upsert to </param>
+        /// <param name="rows"> The rows to be upserted </param>
```
Charles-Gagnon, line 255: It would be good to have this actually describe how the rows are used in this function - it doesn't really have anything to do with upserting.

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,57 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
         /// <returns>List of property names that don't exist in the table</returns>
-        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns)
+        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns, IEnumerable<T> rows)
```
Charles-Gagnon, line 257: I'd suggest just making this just take in a single item T - we only care about a single example to use for parsing so having the entire set isn't really necessary.

Charles-Gagnon (changes requested):

host.json
```diff
@@ -0,0 +1,7 @@
+{
+  "version": "2.0",
+  "extensionBundle": {
```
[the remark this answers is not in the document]
MaddyDev: It is in the node js function folder right?

.vscode/tasks.json
```diff
@@ -64,18 +64,35 @@
...
+			"type": "shell",
+			"label": "npm install (functions)",
+			"command": "npm install",
+			"options": {
+				"cwd": "${workspaceFolder}/samples/samples-js"
+			}
+		},
+		{
+			"type": "shell",
+			"label": "npm prune (functions)",
```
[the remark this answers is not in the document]
MaddyDev: this got automatically added when doing func extensions sync, will remove it since it's not being used.

samples/local.settings.json
```diff
@@ -1,10 +0,0 @@
-{
```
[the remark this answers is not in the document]
MaddyDev: I see it in the samples-csharp folder and I also tested them.

samples/samples-js/AddProduct/index.js
```diff
@@ -0,0 +1,20 @@
...
+        "productid": 1001,
+        "name": "Shoes",
+        "cost": "100"
+    }, {
+        "productid": 1002,
+        "name": "Flowers",
+        "cost": "500"
+    }];
+
+    context.bindings.product = JSON.stringify(products);
```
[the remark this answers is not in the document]
MaddyDev: whatever is defined in the function.json in the bindings part becomes context.bindings. Since I have the name as product in there, I have to use that name to define data. I initially tested with one product, will update it.

samples/samples-js/GetProductsByCost/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = async function (context, req, product) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+    if (product.length === 0) {
```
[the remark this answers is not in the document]
MaddyDev: again same reason, I used the name product in the bindings and was testing with single product - will update it.

src/SqlAsyncCollector.cs
MaddyDev: It would come to this if rows doesn't have any data. There is a check in the parent method ->  if (this._rows.Count != 0)

samples/local.settings.json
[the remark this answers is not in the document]
MaddyDev: my bad, it was not being tracked so it's not added when I was checking in, removed it from .gitignore and checked it now.

host.json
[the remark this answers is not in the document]
Charles-Gagnon: What do you mean? This is at the root of the project currently https://github.com/Azure/azure-functions-sql-extension/blob/maddy/jsLangSupport/host.json

There should be a host.json in the root of the each of the sample projects

samples/samples-js/GetProductsByCost/function.json
```diff
@@ -0,0 +1,29 @@
...
+      "direction": "out"
+    },
+    {
+      "name": "products",
+      "type": "sql",
+      "direction": "in",
+      "commandText": "select * from Products where Cost = @Cost",
+      "commandType": "Text",
+      "parameters": "@Cost=100",
+      "connectionStringSetting": "SqlConnectionString"
```
Charles-Gagnon, line 25: You should be checking in the sample `local.settings.json`

src/SqlAsyncCollector.cs
```diff
@@ -95,7 +96,19 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
                 TelemetryInstance.TrackEvent(TelemetryEventName.AddAsync);
                 try
                 {
-                    this._rows.Add(item);
+                    if (item is string)
```
Charles-Gagnon, line 99: So for these it seems like T might be a string or a JObject? It would be clearer to compare T directly instead of the item IMO (since the item is always type `T` per the signature typing)

src/SqlAsyncCollector.cs
```diff
@@ -95,7 +96,19 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
...
+                    {
+                        this._rows.Add(JsonConvert.DeserializeObject<T>(item as string));
+                    }
+                    else if (item.GetType() == typeof(JObject))
+                    {
+                        dynamic jsonString = JObject.Parse(item.ToString());
```
Charles-Gagnon, line 105: Why is this dynamic? And the name `jsonString` implies it's converting it to a string but `JObject.Parse` returns a JObject. 

Are you sure this shouldn't just be

`this._rows.Add(item.ToString());`

?

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,41 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
         /// to upsert to and returns the extra property names in a List.
         /// </summary>
         /// <param name="columns"> The columns of the table to upsert to </param>
+        /// <param name="row"> First row to get the column names when T is JObject </param>
         /// <returns>List of property names that don't exist in the table</returns>
-        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns)
+        private static IEnumerable<string> GetExtraProperties(IDictionary<string, string> columns, T row)
```
Charles-Gagnon, line 257: Shouldn't this be `T item` or something? A row implies a row in a data table or something, but T is the type of object we're currently handling if I'm understanding this correctly.

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,41 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
         {
+            if (typeof(T) == typeof(JObject))
+            {
+                var jsonObj = JObject.Parse(row.ToString());
```
Charles-Gagnon, line 261: Again you're doing this "turn JObject into a string and then parse it back into a JObject" - which seems unnecessary in this case since we already know it's a JObject

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,41 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
         /// to upsert to and returns the extra property names in a List.
         /// </summary>
         /// <param name="columns"> The columns of the table to upsert to </param>
+        /// <param name="row"> First row to get the column names when T is JObject </param>
```
Charles-Gagnon, line 255: `Sample item used to get the column names when item is a JObject` seems clearer to me

src/SqlAsyncCollector.cs
```diff
@@ -238,21 +253,55 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
+                Dictionary<string, string> dictObj = jsonObj.ToObject<Dictionary<string, string>>();
+                return dictObj.Keys.Where(prop => !columns.ContainsKey(prop))
+                .Select(prop => prop);
+            }
             return typeof(T).GetProperties().ToList()
                 .Where(prop => !columns.ContainsKey(prop.Name))
                 .Select(prop => prop.Name);
         }
 
+        private IEnumerable<string> GetColumnNamesFromPOCO(bool bracketed = false)
```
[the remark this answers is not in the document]
Charles-Gagnon: Ping, do you ever call this with bracketed set to true?

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,41 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
```
Charles-Gagnon, line 271: XML doc please

src/SqlAsyncCollector.cs
```diff
@@ -237,22 +252,41 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
+        {
+            if (typeof(T) == typeof(JObject))
+            {
+                var jsonObj = JObject.Parse(this._rows.First().ToString());
+                Dictionary<string, string> dictObj = jsonObj.ToObject<Dictionary<string, string>>();
+                return dictObj.Keys;
```
Charles-Gagnon, line 277: You aren't handling bracketed here if that ends up actually being necessary

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +310,38 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
             // If there are duplicate primary keys, we'll need to pick the LAST (most recent) row per primary key.
             foreach (T row in rows.Reverse())
             {
-                // SQL Server allows 900 bytes per primary key, so use that as a baseline
-                var combinedPrimaryKey = new StringBuilder(900 * table.PrimaryKeys.Count());
-
-                // Look up primary key of T. Because we're going in the same order of fields every time,
-                // we can assume that if two rows with the same primary key are in the list, they will collide
-                foreach (PropertyInfo primaryKey in table.PrimaryKeys)
+                if (table.PrimaryKeys.Any())
```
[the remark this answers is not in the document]
Charles-Gagnon: ping

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +295,40 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
                 {
-                    // This is the first time we've seen this particular PK. Add this row to the upsert query.
+                    // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
                     rowsToUpsert.Add(row);
                 }
             }
 
             rowData = JsonConvert.SerializeObject(rowsToUpsert, table.JsonSerializerSettings);
-            IEnumerable<string> columnNamesFromPOCO = typeof(T).GetProperties().Select(prop => prop.Name);
+            IEnumerable<string> columnNamesFromPOCO = this.GetColumnNamesFromPOCO();
```
Charles-Gagnon, line 331: Instead of making this a member function it would be better to make `GetColumnNamesFromPOCO` into a static function and pass in the item to use if it's a JObject

src/SqlAsyncCollector.cs
```diff
@@ -318,6 +360,10 @@ public class TableInformation
             /// </summary>
             public string Query { get; }
 
+            /// <summary>
+            /// All of the column names, for SQL to use when data is passed in as JObject
+            /// </summary>
+            public static IEnumerable<string> ColumnNames { get; set; }
```
Charles-Gagnon, line 366: This shouldn't be static - we can have multiple executions of this which may have different columns running at the same time.

And really this should probably be a member on TableInfo or something - the async collector doesn't seem like the correct place to be storing information about the table being upserted to.

src/SqlAsyncCollector.cs
```diff
@@ -336,6 +382,19 @@ public TableInformation(IEnumerable<MemberInfo> primaryKeys, IDictionary<string,
                     ContractResolver = new DynamicPOCOContractResolver(columns, comparer)
                 };
             }
+            /// <summary>
+            /// returns the properties of the object T as Column Names and
+            /// if T is JObject we access the ColumnNames porperties which has to be set
+            /// by reading the keys from the JObject.
+            /// </summary>
+            public static IEnumerable<string> GetColumnNames()
```
Charles-Gagnon, line 390: This also shouldn't be static

src/SqlBindingConfigProvider.cs
```diff
@@ -54,8 +55,34 @@ public void Initialize(ExtensionConfigContext context)
...
+
+    /// <summary>
+    /// Wrapper around OpenType to receive data correctly from output bindings (not as byte[])
+    /// This can be used for general "T --> JObject" bindings. 
+    /// The exact definition here comes from the WebJobs v1.0 Queue binding.
+    /// refer https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs#L390
+    /// </summary>
+    internal class SQLObjectOpenType : OpenType.Poco
+    {
+        // override the OpenType.Poco method to return true when type is an "System.Object"
```
Charles-Gagnon, line 71: Why do we return true in that case?

src/SqlConverters.cs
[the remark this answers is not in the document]
Charles-Gagnon: Oh ok - missed that it was an explicit definition https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

Charles-Gagnon (changes requested):

Charles-Gagnon:
You also have test failures so please fix those

src/SqlAsyncCollector.cs
MaddyDev: since we're using item for the input, I thought it would make sense to have it as row, since T is actually the object representation of Table.

src/SqlAsyncCollector.cs
MaddyDev: I did that and compiler complained -> Argument 1: cannot convert from 'string' to 'T' So I used dynamic.

src/SqlAsyncCollector.cs
MaddyDev: again here the compiler doesn't seem to recognize the row as JObject and am not able to call the ToObject<> method. So I had to the the extra step, or else I probably should do row as JObject.

src/SqlAsyncCollector.cs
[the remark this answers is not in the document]
MaddyDev: I added the comment -> When T is JObject, currently there isn't way to find out the Primary key columns, so it'll be null, adding check. That's why I added a TODO below -> to check for duplicate primary keys in the rows.

src/SqlAsyncCollector.cs
Charles-Gagnon: This isn't even a string though. You're parsing an object - so `jsonString` is actually a JObject after this...

src/SqlAsyncCollector.cs
MaddyDev: I tries using T and compiler won't let me -> 'T' is a type, which is not valid in the given context

src/SqlAsyncCollector.cs
Charles-Gagnon: At the very least something like `rowItem` would be clearer I think. A row has well-known meanings and types, which this isn't.

src/SqlAsyncCollector.cs
MaddyDev: renamed it as itemAsJObject, hope that helps.

src/SqlAsyncCollector.cs
Charles-Gagnon: You have to do `typeof(T) == typeof(JObject)`

src/SqlAsyncCollector.cs
```diff
@@ -95,7 +96,19 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
...
                 try
                 {
-                    this._rows.Add(item);
+                    if (typeof(T) == typeof(string))
+                    {
+                        this._rows.Add(JsonConvert.DeserializeObject<T>(item as string));
+                    }
+                    else if (typeof(T) == typeof(JObject))
+                    {
+                        dynamic itemAsJObject = JObject.Parse(item.ToString());
```
Charles-Gagnon, line 105: So I realized that there's a more root problem - T is going to be whatever the type is originally (string, JObject, whatever), but then you're trying to change the type that you insert into the rows list. But the rows list is defined as `List<T>` which implies that it'll have the same type....so what you're trying to do here is breaking that contract. 

You need to be looking at either changing the type of _rows or not doing any conversion here and instead doing it later on somewhere.

src/SqlAsyncCollector.cs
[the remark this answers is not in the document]
Charles-Gagnon: It would be better to check against the type so that we're sure that we're not skipping over an error happening for C# functions I think then

src/SqlAsyncCollector.cs
```diff
@@ -446,7 +484,7 @@ AS NewData
             /// <param name="fullName">Full name of table, including schema (if exists).</param>
             /// <param name="logger">ILogger used to log any errors or warnings.</param>
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
-            public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger)
+            public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> columnNames)
```
Charles-Gagnon, line 482: Update xml doc

src/SqlAsyncCollector.cs
```diff
@@ -543,7 +581,11 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                 // Match SQL Primary Key column names to POCO field/property objects. Ensure none are missing.
                 StringComparison comparison = caseSensitive ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
                 IEnumerable<MemberInfo> primaryKeyFields = typeof(T).GetMembers().Where(f => primaryKeys.Any(k => string.Equals(k.Name, f.Name, comparison)));
-                IEnumerable<string> primaryKeysFromPOCO = primaryKeyFields.Select(f => f.Name);
+                if (!columnNames.Any())
```
Charles-Gagnon, line 579: Add a comment about what this is doing, but also should we even be doing the logic on line 583 for JObjects? It seems like we should just be having completely separate code paths for JObject and other types.

Directory.Packages.props
```diff
@@ -12,5 +12,6 @@
     <PackageVersion Include="Moq" Version="4.14.3" />
     <PackageVersion Include="xunit" Version="2.4.0" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageVersion Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator" Version="1.2.2" />
```
chlafreniere, line 15: Any reason why we're not using 1.2.3? https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator/

samples/samples-js/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "samples-js",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "start": "func start",
+    "test": "echo \"No tests yet...\""
```
chlafreniere, line 7: What should this really be instead of the echo command?

src/SqlAsyncCollector.cs
```diff
@@ -95,7 +96,14 @@ public async Task AddAsync(T item, CancellationToken cancellationToken = default
                 TelemetryInstance.TrackEvent(TelemetryEventName.AddAsync);
                 try
                 {
-                    this._rows.Add(item);
+                    if (typeof(T) == typeof(string))
+                    {
+                        this._rows.Add(JsonConvert.DeserializeObject<T>(item as string));
```
Charles-Gagnon, line 101: This seems odd to me too. item is a string, and then we're calling JsonConvert.DeserializeObject on it...but T is also string. So we're deserializing a string into...a string? I'm pretty sure this is also doing nothing unless there's some behavior I'm missing.

Charles-Gagnon (changes requested):

Directory.Packages.props
MaddyDev: we ran into issues with the older version when setting up the functions earlier and updating to latest solved it.

src/SqlAsyncCollector.cs
MaddyDev: not really sure, but that's the same thing cosmosdb folks have on their repo, I just replicated what they suggested and this code is currently not being hit since what we actually get is a JObject.

Directory.Packages.props
chlafreniere: Yeah, but latest 1.x version is 1.2.3 but we're using 1.2.2

samples/samples-js/package.json
MaddyDev: Will have to look up ways of adding tests to the node js app and add that here. I'll be working on it.

src/SqlAsyncCollector.cs
Charles-Gagnon: We shouldn't be doing something just because their code is doing it - we need to know what our code is doing and how it works regardless of what their implementation looks like.

I'd suggest removing this if the code isn't even being hit. If that causes problems then we can dive into what exactly needs to happen here.

Directory.Packages.props
MaddyDev: will update it then

src/SqlAsyncCollector.cs
MaddyDev: Well, those are changes they suggested to add javascript support, I wasn't really sure what works and what don't until the entire thing came along. And it made sense that when we get the input from javascript as a string, to deserialize it and add it to rows.

src/SqlBindingConfigProvider.cs
MaddyDev: I sent a message to Mathew Charles (Alexey's contact), will update comments on why this is needed once I hear back.

samples/samples-js/package.json
Charles-Gagnon: Just remove it then. No reason to keep stuff around if it's not doing anything.

src/SqlAsyncCollector.cs
```diff
@@ -237,14 +239,32 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
+                Dictionary<string, string> dictObj = (rowItem as JObject).ToObject<Dictionary<string, string>>();
+                return dictObj.Keys.Where(prop => !columns.ContainsKey(prop))
+                .Select(prop => prop);
+            }
             return typeof(T).GetProperties().ToList()
                 .Where(prop => !columns.ContainsKey(prop.Name))
                 .Select(prop => prop.Name);
         }
 
+        private static IEnumerable<string> GetColumnNamesFromPOCO(T row, bool bracketed = false)
```
Charles-Gagnon, line 257: I still don't see bracketed ever being set to true. Can you point me to where that happens? If not then just remove this parameter.

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +281,40 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
             // If there are duplicate primary keys, we'll need to pick the LAST (most recent) row per primary key.
             foreach (T row in rows.Reverse())
             {
-                // SQL Server allows 900 bytes per primary key, so use that as a baseline
-                var combinedPrimaryKey = new StringBuilder(900 * table.PrimaryKeys.Count());
-
-                // Look up primary key of T. Because we're going in the same order of fields every time,
-                // we can assume that if two rows with the same primary key are in the list, they will collide
-                foreach (PropertyInfo primaryKey in table.PrimaryKeys)
+                // When T is JObject, currently there isn't way to find out the Primary keys, so it'll be null, adding check.
```
Charles-Gagnon, line 284: I'd just remove this comment, it's clear enough now what the if statement is doing and the else already mentions needing to find out how to get primary keys.

src/SqlAsyncCollector.cs
```diff
@@ -543,7 +570,12 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                 // Match SQL Primary Key column names to POCO field/property objects. Ensure none are missing.
                 StringComparison comparison = caseSensitive ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
                 IEnumerable<MemberInfo> primaryKeyFields = typeof(T).GetMembers().Where(f => primaryKeys.Any(k => string.Equals(k.Name, f.Name, comparison)));
-                IEnumerable<string> primaryKeysFromPOCO = primaryKeyFields.Select(f => f.Name);
+                // When T is JObject, primaryKeyFields cannot be used to get the column names, so use the passed in columnNames variable
```
Charles-Gagnon, line 573: You should refactor this to have the column names always be passed in - and then for POCO objects it would just be the list of properties from the object. Having to sometimes use the parameter and sometimes not is confusing and could easily result in logic mistakes.

src/SqlAsyncCollector.cs
MaddyDev: I removed the one place it was being used since it's not actually being referred anywhere -> https://github.com/Azure/azure-functions-sql-extension/blob/a61c8478bf6f256842c0b9792135a3ed5eeff717/src/SqlAsyncCollector.cs#L399

Charles-Gagnon: Rename to `GetColumnNamesFromItem`

And please add comments explaining the reasoning for getting columns in this way for each type

Charles-Gagnon: I'm not really sure what you're saying. I'm pointing out that you have a parameter `bracketed` with a default of false, but never actually call this method in a way that sets it to true. So you can just remove the parameter since it's not doing anything.

(assuming of course this isn't a bug and you actually don't care about this being bracketed ever...)

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +281,39 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
                 {
-                    // This is the first time we've seen this particular PK. Add this row to the upsert query.
+                    // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
                     rowsToUpsert.Add(row);
                 }
             }
 
             rowData = JsonConvert.SerializeObject(rowsToUpsert, table.JsonSerializerSettings);
-            IEnumerable<string> columnNamesFromPOCO = typeof(T).GetProperties().Select(prop => prop.Name);
+            IEnumerable<string> columnNamesFromPOCO = GetColumnNamesFromPOCO(rows.First());
```
Charles-Gagnon, line 316: `columnNamesFromItem`

src/SqlAsyncCollector.cs
```diff
@@ -414,9 +439,9 @@ public static string GetMergeQuery(IList<PrimaryKey> primaryKeys, SqlObject tabl
                 }
 
                 // Generate the UPDATE part of the merge query (all columns that should be updated)
-                IEnumerable<string> bracketedColumnNamesFromPOCO = typeof(T).GetProperties()
-                    .Where(prop => !primaryKeys.Any(k => k.IsIdentity && string.Equals(k.Name, prop.Name, comparison))) // Skip any identity columns, those should never be updated
-                    .Select(prop => prop.Name.AsBracketQuotedString());
+                IEnumerable<string> bracketedColumnNamesFromPOCO = columnNames
```
Charles-Gagnon, line 442: `bracketedColumnNamesFromItem`

src/SqlAsyncCollector.cs
```diff
@@ -543,9 +569,9 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                 // Match SQL Primary Key column names to POCO field/property objects. Ensure none are missing.
                 StringComparison comparison = caseSensitive ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase;
                 IEnumerable<MemberInfo> primaryKeyFields = typeof(T).GetMembers().Where(f => primaryKeys.Any(k => string.Equals(k.Name, f.Name, comparison)));
-                IEnumerable<string> primaryKeysFromPOCO = primaryKeyFields.Select(f => f.Name);
+                IEnumerable<string> primaryKeysFromItem = columnNames.Where(f => primaryKeys.Any(k => string.Equals(k.Name, f, comparison)));
                 IEnumerable<PrimaryKey> missingPrimaryKeysFromPOCO = primaryKeys
```
Charles-Gagnon: `missingPrimaryKeysFromItem`

samples/samples-js/package.json
Charles-Gagnon: @MaddyDev This seems to have snuck back in again...

Charles-Gagnon (changes requested):

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +281,39 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
                     }
-                    combinedPrimaryKey.Append(value.ToString());
                 }
-
-                // If we have already seen this unique primary key, skip this update
-                if (uniqueUpdatedPrimaryKeys.Add(combinedPrimaryKey.ToString()))
+                else
                 {
-                    // This is the first time we've seen this particular PK. Add this row to the upsert query.
+                    // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
```
Charles-Gagnon, line 310: As a note - I'm not really sure why this is something you can't do now. Primary keys themselves are a part of the  SQL table so have nothing to do with the item type. 

The only tricky thing is to figure out how to get the values from a JObject - currently this looks like it's storing the MemberInfo from the POCOs and then using those to get the values at runtime for each item/row. But that won't work for JObject - so there will need to be some slight reworking of that class - probably instead having two different implementations of TableInformation - one for JObject and one for the other generic types.

MaddyDev: yes, I'll work on that after I get this in.

samples/samples-js/package.json
MaddyDev: I added it back since one test started failing once I removed it. Going to look into that but don't want to block the PR on it.

src/SqlAsyncCollector.cs
```diff
@@ -445,8 +475,9 @@ AS NewData
             /// <param name="sqlConnection">An open connection with which to query SQL against</param>
             /// <param name="fullName">Full name of table, including schema (if exists).</param>
             /// <param name="logger">ILogger used to log any errors or warnings.</param>
+            /// <param name="columnNames">Column names of the table</param>
```
Charles-Gagnon, line 478: This isn't accurate though, right? It's actually column names from the object (that should map to ones from the table, but that's not really important here)

src/SqlAsyncCollector.cs
```diff
@@ -261,32 +286,39 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
-                    // This is the first time we've seen this particular PK. Add this row to the upsert query.
+                    // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
                     rowsToUpsert.Add(row);
                 }
             }
 
             rowData = JsonConvert.SerializeObject(rowsToUpsert, table.JsonSerializerSettings);
-            IEnumerable<string> columnNamesFromPOCO = typeof(T).GetProperties().Select(prop => prop.Name);
+            IEnumerable<string> columnNamesFromPOCO = GetColumnNamesFromItem(rows.First());
             IEnumerable<string> bracketColumnDefinitionsFromPOCO = table.Columns.Where(c => columnNamesFromPOCO.Contains(c.Key, table.Comparer))
```
Charles-Gagnon, line 322: `bracketedColumnDefinitionsFromItem`

Charles-Gagnon (approved):

chlafreniere (approved):

Issue #232, opened 2022-04-30
`Cannot insert the value NULL into column <COLUMN_NAME>` on SQL output binding in Javascript function


horihiro:
I understand that this extension supports Javascript/Typescript from `v.0.1.304-preview`.
https://github.com/Azure/azure-functions-sql-extension/releases/tag/v.0.1.304-preview

Do you have any samples in Javascript/Typescript?

horihiro:
I found examples in [tihs](https://github.com/Azure/azure-functions-sql-extension/tree/maddy/jsSamples/samples/samples-js) and created the following repo.
https://github.com/horihiro/azure-functions-sqlbinding-js-example

I set up SQL Database by following this [instruction](https://github.com/Azure/azure-functions-sql-extension#sql-setup).

In a local environment, I can make the function that has the input binding work by the following additional steps. 

  1. `func extensions install`
  2. `cp ./bin/extensions.deps.json ./bin/function.deps.json`

But the one that has output binding cannot work and shows the following error even though the above steps are added.
![image](https://user-images.githubusercontent.com/4566555/166128851-32f5eaef-63a8-46c5-9753-7b10556cfc9e.png)

<details>
<summary>logs</summary>

```
[2022-05-01T**:**:**.***Z] Executing 'Functions.HttpTriggerOutputBinding' (Reason='This function was programmatically called via the host APIs.', Id=1d3ae9fc-5f19-46f5-8cd0-948d53f76373)
[2022-05-01T**:**:**.***Z] JavaScript HTTP trigger function processed a request.
[2022-05-01T**:**:**.***Z] Sending event Create
[2022-05-01T**:**:**.***Z] Sending event AddAsync
[2022-05-01T**:**:**.***Z] Sending event FlushAsync
[2022-05-01T**:**:**.***Z] Sending event TableInfoCacheMiss
[2022-05-01T**:**:**.***Z] Sending event GetTableInfoStart
[2022-05-01T**:**:**.***Z] Sending event GetCaseSensitivity
[2022-05-01T**:**:**.***Z] Sending event GetColumnDefinitions
[2022-05-01T**:**:**.***Z] Sending event GetPrimaryKeys
[2022-05-01T**:**:**.***Z] Sending event GetTableInfoEnd
[...]

PR #234, opened 2022-05-04
JavaScript samples and tests

MaddyDev:
Added input and output binding samples and tests for each sample. 
For tests, reused the existing code and updated the StartFunctionHost() method in IntegrationTestBase class to start the func.exe from JavaScript sample folder.

Also addressed the fix for #232 where the column names are case insensitive, we are building the sql query converting the columns to lowercase but the javascript object keys are being passed as defined and the sql query is expecting lowercased values. 
Fix: Added an extra step to convert the JObjects keys to lower case and passed that to the sql query.
Tried using the JsonSerializerSettings and passing the NamingStrategy to convert property names to lowercase but that was not working and still returned keys as is. Alternatively had to convert the keys manually by converting the object to dictionary, then converting keys to lowercase and then converting it back to JObject for serialization.
P.S: Still figuring out how to access the js samples folder for integration tests on build machines.


Charles-Gagnon:
Please do a separate PR for the casing fix first so we can get that in and verified separately

test/Integration/SqlOutputBindingJSIntegrationTests.cs
```diff
@@ -0,0 +1,239 @@
...
+{
+    [Collection("IntegrationTests")]
+    public class SqlOutputBindingJSIntegrationTests : IntegrationTestBase
+    {
+        private readonly string workingDirectoryFolder = "samples-js";
+        public SqlOutputBindingJSIntegrationTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        private async Task<HttpResponseMessage> SendOutputRequest(string functionName, IDictionary<string, string> query = null)
```
Charles-Gagnon, line 22: We shouldn't be needing to duplicate all this code - the tests to verify the APIs should be the same across all the languages. The only thing that we should expect to change is how the function host is started up.

Charles-Gagnon (changes requested):

test/Integration/SqlOutputBindingJSIntegrationTests.cs
MaddyDev: All I did is create a separate class and duplicated this one SendRequest method for easier management, if we add them all in once class, I have to name the tests tagging with language to make that distinction. I thought this was cleaner.
In this case it's justified because the c# output binding tests use Get requests for updating data as well (since they use query strings for reading the product data) But javascript samples I added, am reading the product data from body. So I have to call post request.

samples/samples-csharp/OutputBindingSamples/AddProductParams.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.OutputBindingSamples
+{
+    public static class AddProductParams
+    {
+        [FunctionName("AddProductParams")]
+        public static IActionResult Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproduct")]
+            HttpRequest req,
+            [Sql("dbo.Products", CommandType = System.Data.CommandType.Text, ConnectionStringSetting = "SqlConnectionString")] out Product product)
```
Charles-Gagnon, line 17: Shouldn't need CommandType

samples/samples-js/AddProduct/index.js
```diff
@@ -0,0 +1,16 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const testProduct = {
```
Charles-Gagnon, line 4: These samples should be behaving identically - so this one should match the C# one which is pulling from the body

samples/samples-js/AddProduct/function.json
```diff
@@ -0,0 +1,28 @@
+{
+  "bindings": [
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get",
```
Charles-Gagnon, line 9: Just POST to match the C# binding sample with the same name

samples/samples-js/AddProductExtraColumns/index.js
```diff
@@ -0,0 +1,18 @@
+module.exports = async function (context) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const products = [{
+        "productid": 1001,
+        "name": "Shoes",
+        "cost": "100",
+		"extraint": 1,
```
Charles-Gagnon, line 8: format

samples/samples-js/AddProductExtraColumns/index.js
```diff
@@ -0,0 +1,18 @@
+module.exports = async function (context) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const products = [{
+        "productid": 1001,
```
Charles-Gagnon, line 5: Let's fix these to have the correct casing now that we have a fix for that

samples/samples-csharp/OutputBindingSamples/AddProductParams.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.OutputBindingSamples
+{
+    public static class AddProductParams
+    {
+        [FunctionName("AddProductParams")]
+        public static IActionResult Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproduct")]
```
Charles-Gagnon, line 15: Route is incorrect here

samples/samples-js/AddProductParams/function.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get",
+        "post"
+      ],
+      "route": "addproduct/{productid}/{name}/{cost}"
```
Charles-Gagnon, line 12: The route should generally match the function name for our samples. And this isn't how you define the route - the query parameters for GET requests aren't included as part of this (what you have here is part of the path itself, e.g. `addproduct/1/MyItem/100`, parameters would look like `addproduct?id=1&Name=MyItem&Cost=100`)

samples/samples-js/AddProductParams/index.js
```diff
@@ -0,0 +1,16 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const newProduct = {
+        "productid": req.params?.productid ?? null,
```
Charles-Gagnon, line 5: `?? null` isn't really doing anything - it's already null (that's what the ?? operator is checking)

samples/samples-js/AddProductWithIdentityColumn/index.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const itemToInsert = {
+        "name": "productWithIdentity",
```
Charles-Gagnon, line 5: You should be parsing the name and cost from the parameters like the C# function

samples/samples-js/AddProductWithIdentityColumnIncluded/index.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const itemToInsert = {
```
Charles-Gagnon, line 4: You should be parsing the name and cost from the parameters like the C# function

samples/samples-js/AddProductWithMultiplePrimaryColumnsAndIdentity/index.js
```diff
@@ -0,0 +1,16 @@
+module.exports = async function (context, request) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const testProduct = {
```
Charles-Gagnon, line 4: You should be parsing the name and cost from the parameters like the C# function

samples/samples-js/AddProductWithMultiplePrimaryColumnsAndIdentity/function.json
```diff
@@ -0,0 +1,28 @@
+{
+  "bindings": [
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get",
+        "post"
```
Charles-Gagnon, line 10: Just GET, not POST

(please go through all the functions you've made and verify that they match the method types of the corresponding C# functions. We should follow up later on whether we should be using GET or POST for each one, but for now let's at least keep them consistent)

samples/samples-js/AddProductsArray/index.js
```diff
@@ -0,0 +1,20 @@
...
+        "productid": 1,
+        "name": "Cup",
+        "cost": 2
+    }, {
+        "productid": 2,
+        "name": "Glasses",
+        "cost": 12
+    }];
+
+    context.bindings.products = req["body"] ?? JSON.stringify(products);
```
Charles-Gagnon, line 14: Let's keep this the same as the C# function and only insert the local items directly, not parse from body. 

(Also you should generally access properties with . notation, e.g. `req.body`)

samples/samples-js/AddProductsNoPartialUpsert/index.js
```diff
@@ -0,0 +1,24 @@
+module.exports = async function (context) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    var products = [];
+    for(let i = 0; i < 1000; i++) {
+        products.push(...{"productid": i,
```
Charles-Gagnon, line 6: Is this working correctly? You're expanding the object here so I'm pretty sure this is going to end up with the values being inserted into the array instead of the item objects.

samples/samples-js/GetProductNamesView/function.json
```diff
@@ -0,0 +1,29 @@
...
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "methods": [
+        "get",
+        "post"
+      ],
+      "route":"getproduct-namesview/"
```
Charles-Gagnon, line 12: Remove /

samples/samples-js/GetProductNamesView/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = async function (context, req, products) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+    if (products.length === 0) {
```
Charles-Gagnon, line 3: Clean up, we don't need this logging stuff here (and if we decide to add later we should add it to all languages)

(do this for all the other functions as well)

samples/samples-js/GetProductsByName/function.json
```diff
@@ -0,0 +1,30 @@
...
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "methods": [
+        "get",
+        "post"
+      ],
+      "route":"getproductsbyname/{name}"
```
Charles-Gagnon, line 12: I don't see a C# function for this - is there a reason you're making a new one of this?

samples/samples-js/GetProductNamesView/index.js
```diff
@@ -0,0 +1,12 @@
...
+        context.log("Product(s) not found");
+    } else {
+        context.log("Found Product(s), Count=" + products.length + " -> " + JSON.stringify(products));
+    }
+    return {
+        status: 201,
```
Charles-Gagnon, line 9: 201 is created which isn't correct, this should be 200 (go through and verify all functions are returning the correct code)

samples/samples-js/README.md
```diff
@@ -0,0 +1,9 @@
+## Run Host
+
+To run the function host locally follow these steps :
+
+1. Run `func extensions install --package Microsoft.Azure.WebJobs.Extensions.Sql --version 0.1.286-preview` from the function folder to install the required extension package. This step will add bin folder with all the dependencies and extension.csproj.
```
Charles-Gagnon, line 5: Remove this now that we don't need it anymore

samples/samples-js/UpsertProducts/function.json
```diff
@@ -0,0 +1,28 @@
+{
```
Charles-Gagnon, line 1: Remove, this looks like a duplicate of the AddProductsArray

samples/samples-js/host.json
```diff
@@ -0,0 +1,15 @@
...
+    "applicationInsights": {
+      "samplingSettings": {
+        "isEnabled": true,
+        "excludedTypes": "Request"
+      }
+    }
+  },
+  "extensionBundle": {
+    "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
+    "version": "[3.*, 4.0.0)"
```
Charles-Gagnon, line 13: Might need to bump this to 4.* so we get the casing fix

Although we need to figure out what to do here since right now this means that this is going to be running against the bundle version of the extension and NOT the current code so that kind of defeats the purpose of having tests.

samples/samples-js/packages.lock.json
```diff
@@ -0,0 +1,1735 @@
+{
```
Charles-Gagnon, line 1: delete

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -20,44 +21,90 @@ public SqlOutputBindingIntegrationTests(ITestOutputHelper output) : base(output)
         {
         }
 
-        private async Task<HttpResponseMessage> SendOutputRequest(string functionName, IDictionary<string, string> query = null)
+        private async Task<HttpResponseMessage> SendOutputRequest(string functionName, IDictionary<string, string> query = null, bool asPost = false)
```
Charles-Gagnon, line 24: I'd rather we just split them up into separate functions, they're small enough and do different enough things that that'll be a bit cleaner.

Charles-Gagnon (changes requested):

test/Integration/SqlInputBindingIntegrationTests.cs
```diff
@@ -134,7 +147,28 @@ public async void GetProductNamesViewTest()
             string expectedResponse = "[{\"name\":\"test\"}]";
             string actualResponse = await response.Content.ReadAsStringAsync();
 
-            Assert.Equal(expectedResponse, actualResponse, StringComparer.OrdinalIgnoreCase);
+            Assert.Equal(expectedResponse, actualResponse.Trim().Replace(" ", "").Replace(Environment.NewLine, ""), StringComparer.OrdinalIgnoreCase);
+        }
+
+        [Theory]
+        [InlineData("same", "samples-js")]
```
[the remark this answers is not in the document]
MaddyDev: I added this test in JavaScript before refactoring and realized there is no equivalent c# test for it. They have GetProductsNameEmptyTest which is slightly different but queries by name. So didn't create a new one for it in c#.

test/Integration/SqlInputBindingIntegrationTests.cs
```diff
@@ -46,16 +49,19 @@ public async void GetProductsTest(int n, int cost)
             string expectedResponse = JsonConvert.SerializeObject(products);
             string actualResponse = await response.Content.ReadAsStringAsync();
 
-            Assert.Equal(expectedResponse, actualResponse, StringComparer.OrdinalIgnoreCase);
+            Assert.Equal(expectedResponse, actualResponse.Trim().Replace(" ", "").Replace(Environment.NewLine, ""), StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
MaddyDev: I tried parsing but still the JavaScript response has new lines and tabs so comparing the response without removing them is failing the test.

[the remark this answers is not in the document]
Charles-Gagnon: The parsed JSON object still had newlines and tabs? I'm not sure how that'd be possible - unless the values themselves had these. Can you give an example of what code you used to try this?

test/Integration/SqlInputBindingIntegrationTests.cs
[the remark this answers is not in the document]
Charles-Gagnon: Remove this (and any other js-only ones) for now. We can add them later, but if we do let's do it as a separate PR where we add for all languages and not just JS

samples/samples-js/host.json
MaddyDev: yeah while running tests locally, am removing this from host.json. Not sure how to suppress this on build machine, any ideas?

samples/samples-csharp/OutputBindingSamples/AddProduct.cs
```diff
@@ -14,12 +14,12 @@ public static class AddProduct
         public static IActionResult Run(
             [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproduct")]
             HttpRequest req,
-            [Sql("dbo.Products", ConnectionStringSetting = "SqlConnectionString")] out Product product)
+            [Sql("dbo.Products", CommandType = System.Data.CommandType.Text, ConnectionStringSetting = "SqlConnectionString")] out Product product)
```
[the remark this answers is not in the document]
MaddyDev: was playimg around to fix a test error and forgot to remove it. Will be taken care of in the next PR

samples/samples-js/AddProduct/index.js
MaddyDev: yeah but when am testing locally not from tests, I wanted it to return the data when I invoke it from browser so added it. Would it be okay to leave it as is?

samples/samples-js/GetProductsByName/function.json
MaddyDev: before replicating all the functions from c# samples, I added this.

samples/samples-csharp/OutputBindingSamples/AddProductParams.cs
MaddyDev: ?

samples/samples-js/AddProductExtraColumns/index.js
```diff
@@ -0,0 +1,18 @@
+module.exports = async function (context) {
```
Charles-Gagnon, line 1: This doesn't really belong in the samples folder since it's showing incorrect behavior. For now let's just only add JS versions of the functions in the samples folder - any in the test folder just skip and we'll add those separately when we add the tests later.

samples/samples-js/AddProductParams/function.json
```diff
@@ -0,0 +1,27 @@
...
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get"
+      ],
+      "route": "addproduct/{productId}/{name}/{cost}"
```
Charles-Gagnon, line 11: This route isn't correct - query params aren't a part of the route itself

samples/samples-js/GetProductsByName/function.json
Charles-Gagnon: Remove it for now then, if we think it's useful to add later we can (but should do it for all languages at that point)

Charles-Gagnon (changes requested):

samples/samples-js/AddProductWithIdentityColumn/index.js
MaddyDev: AM thinking of leaving the reading from body part so we support both get and post in the sample. What do you think?

Charles-Gagnon: I would definitely agree that it would be great to do that at some point, but for now let's focus on getting the samples at parity and then we can come in and make targeted changes like that.

MaddyDev: all the samples are at parity with c# now, leaving that wouldn't do any harm and would avoid the rework.

samples/samples-js/AddProductWithIdentityColumn/index.js
```diff
@@ -0,0 +1,14 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const itemToInsert = {
+        "name": req.params?.name ?? "productWithIdentity",
```
Charles-Gagnon, line 5: Don't have defaults, we can add that functionality in later if we think it's necessary

samples/samples-js/AddProductWithIdentityColumn/index.js
```diff
@@ -0,0 +1,14 @@
...
+        "cost": req.params?.cost ?? 100
+    };
+    context.bindings.products = req.body ?? JSON.stringify(itemToInsert);
```
Charles-Gagnon, line 8: Starting a new thread so it shows up correctly in the files view - for now let's keep it a priority to make these samples behave the same. Later on we can come back and rework all the checked in samples to add additoinal functionality such as parsing from the body. 

That won't be a huge amount of work - just change this line to 

`context.bindings.products = JSON.stringify(itemToInsert);`

(but also you should update `products` to `product` since there's only one object being added)

samples/samples-csharp/OutputBindingSamples/AddProductsArray.cs
```diff
@@ -1,39 +1,25 @@
...
         public static IActionResult Run(
-            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproducts-array")]
-            HttpRequest req,
+            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "addproducts-array")]
+            [FromBody] List<Product> products,
             [Sql("dbo.Products", ConnectionStringSetting = "SqlConnectionString")] out Product[] output)
         {
             // Suppose that the ProductID column is the primary key in the Products table, and the
             // table already contains a row with ProductID = 1. In that case, the row will be updated
             // instead of inserted to have values Name = "Cup" and Cost = 2.
```
Charles-Gagnon, line 21: Let's update this comment as well so it's more accurate. (and add this comment to the JS function version as well)
```
Upsert the products, which will insert them into the Products table if the primary key (ProductId) for that item doesn't exist. If it does then update it to have the new name and cost
```

Charles-Gagnon:
There's also some test failures, please investigate to see if they're related to your changes

samples/samples-js/AddProductParams/function.json
```diff
@@ -0,0 +1,27 @@
...
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get"
+      ],
+      "route": "addproduct-params/{productId}/{name}/{cost}"
```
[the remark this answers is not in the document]
MaddyDev: I tested it and the parameters are null if they are not added to the route.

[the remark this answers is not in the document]
Charles-Gagnon: Are you making sure to actually pass the query parameters?

e.g. 

`https://localhost:7070/api/addproduct-params?productId=1&name=Bike&Cost=30`

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -32,6 +34,13 @@ private async Task<HttpResponseMessage> SendOutputRequest(string functionName, I
             return await this.SendGetRequest(requestUri);
         }
 
+        private async Task<HttpResponseMessage> SendOutputPostRequest(string functionName, string query)
+        {
+            string requestUri = $"http://localhost:{this.Port}/api/{functionName}";
+
+            return await this.SendPostRequest(requestUri, query);
```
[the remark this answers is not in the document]
MaddyDev: compiler not letting me to do that, I should also update the return type, complier error -> "Since this is an async method, the return expression must be of type 'HttpResponseMessage' rather than 'Task<HttpResponseMessage>'"

samples/samples-js/AddProductParams/function.json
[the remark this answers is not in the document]
MaddyDev: image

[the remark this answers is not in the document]
Charles-Gagnon: 500 means that the route isn't even correct. Did you change the route value to just be `addproduct-params`?

And if so did you copy the address from the startup text that func.exe displays? What I put above was just from memory so I might have gotten something wrong - it should look something like this 

`HttpTrigger1:  http://localhost:7071/api/getproducts/{cost}`

And then just append `?productId=1&name=Bike&Cost=30` to the end of that.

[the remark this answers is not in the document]
MaddyDev: 404 means the route is incorrect and not able to find it, 500 is  internal server error-> the server encountered an unexpected condition that prevented it from fulfilling the request. 
And yes I did copy it from the function endpoints display and then appended the route parameters.

test/Integration/SqlOutputBindingIntegrationTests.cs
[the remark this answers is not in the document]
MaddyDev: Shouldn't we wait until the response for the tests? Having async and awaiting makes sense to me, why do you think that's not needed?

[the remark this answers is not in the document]
Charles-Gagnon: You're still returning the task. The caller should be awaiting that task. Functionally this should not change anything.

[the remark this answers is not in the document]
MaddyDev: ohh from the compiler error, I thought I also need to update it return HttpResponseMessage instead of Task along with removing the async! Gotcha, thanks.

samples/samples-js/AddProductWithIdentityColumnIncluded/index.js
```diff
@@ -0,0 +1,16 @@
...
+        "productId" : req.query?.productId,
+        "name": req.query?.name,
+        "cost": req.query?.cost
+    };
+
+    context.bindings.product = JSON.stringify(itemToInsert);
+
+    return {
+        status: 201,
+        body: context.bindings.products
```
Charles-Gagnon, line 14: `product`

Charles-Gagnon (approved):

PR #235, opened 2022-05-04
fix casing issue when sql table is case insensitive

### LINKED ISSUE #232: `Cannot insert the value NULL into column <COLUMN_NAME>` on SQL output binding in Javascript function

MaddyDev:
Fixes #232  where the column names are case insensitive, we are building the sql query converting the columns to lowercase but the javascript object keys are being passed as defined and the sql query is expecting lowercased values. 

Fix: Added an extra step to convert the JObjects keys to lower case and passed that to the sql query.
Tried using the JsonSerializerSettings and passing the NamingStrategy to convert property names to lowercase but that was not working and still returned keys as is. Alternatively manually updated the properties to lower case by iterating through them.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -313,9 +313,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    // lower case the json keys if sql table is case insensitive since the query we built expects everything lowercased.
                     if (table.Comparer == StringComparer.OrdinalIgnoreCase)
                     {
-                        ChangePropertiesToLowerCase(JObject.Parse(row.ToString()));
+                        ChangePropertiesToLowerCase(row as JObject);
                     }
                     rowsToUpsert.Add(row);
                 }
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -313,6 +313,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    if (table.Comparer == StringComparer.OrdinalIgnoreCase)
+                    {
+                        ChangePropertiesToLowerCase(JObject.Parse(row.ToString()));
+                    }
                     rowsToUpsert.Add(row);
                 }
             }
@@ -322,6 +326,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             IEnumerable<string> bracketColumnDefinitionsFromItem = table.Columns.Where(c => columnNamesFromItem.Contains(c.Key, table.Comparer))
                 .Select(c => $"{c.Key.AsBracketQuotedString()} {c.Value}");
             newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON({RowDataParameter}) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
+        }
+
+        private static void ChangePropertiesToLowerCase(JObject jsonObject)
+        {
+            foreach (JProperty property in jsonObject.Properties().ToList())
+            {
+                if (property.Value.Type == JTokenType.Object)
+                {
+                    ChangePropertiesToLowerCase((JObject)property.Value);
+                }
+                property.Replace(new JProperty(property.Name.ToLowerInvariant(), property.Value));// properties are read-only, so we have to replace them
+            }
         }
 
         public class TableInformation
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -313,10 +313,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
-                    // lower case the json keys if sql table is case insensitive since the query we built expects everything lowercased.
+                    // JObjects ignore serializer settings (https://web.archive.org/web/20171005181503/http://json.codeplex.com/workitem/23853)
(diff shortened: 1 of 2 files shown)


src/SqlAsyncCollector.cs
```diff
@@ -313,6 +313,10 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    if (table.Comparer == StringComparer.OrdinalIgnoreCase)
+                    {
+                        ChangePropertiesToLowerCase(JObject.Parse(row.ToString()));
```
Charles-Gagnon, line 318: How is this actually changing anything? You convert the JObject to a string and then turn it back into a JObject which you then modify. But the original row isn't modified at all...

src/SqlAsyncCollector.cs
```diff
@@ -313,6 +313,10 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    if (table.Comparer == StringComparer.OrdinalIgnoreCase)
```
Charles-Gagnon, line 316: Comments explaining why this is necessary

src/SqlAsyncCollector.cs
```diff
@@ -324,6 +328,18 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
             newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON({RowDataParameter}) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
         }
 
+        private static void ChangePropertiesToLowerCase(JObject jsonObject)
+        {
+            foreach (JProperty property in jsonObject.Properties().ToList())
+            {
+                if (property.Value.Type == JTokenType.Object)
```
Charles-Gagnon, line 335: Seems better to actually check the C# type here (make sure it's actually a JObject)

Charles-Gagnon (changes requested):

src/SqlAsyncCollector.cs
MaddyDev: I didn't understand what you mean?

Charles-Gagnon: `property.Value.GetType() == typeof(JObject)`

It seems odd to rely on Value.Type when you have the actual object - especially since we then cast it to JObject right below anyways.

src/SqlAsyncCollector.cs
```diff
@@ -313,6 +313,11 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    // lower case the json keys if sql table is case insensitive since the query we built expects everything lowercased.
+                    if (table.Comparer == StringComparer.OrdinalIgnoreCase)
+                    {
+                        ChangePropertiesToLowerCase(row as JObject);
```
Charles-Gagnon, line 319: So we already have JsonSerializerSettings in table information - which we then use to generate the rowData. Why isn't that handling converting it to lowercase? It would be good to look into using that same logic instead of doing this custom logic here.

Charles-Gagnon (changes requested):

src/SqlAsyncCollector.cs
MaddyDev: I already tried that before coming to this solution, it was not changing the case and I was not able to find out why.

src/SqlAsyncCollector.cs
MaddyDev: I was trying typeof(property.Value) == typeof(JObject) and it was showing me a compiler error, I need to learn when to use GetType vs typeof :|

src/SqlAsyncCollector.cs
Charles-Gagnon: Alright, so I looked into this and the root issue here is because contract resolvers are ignored by default when serializing JObjects (https://web.archive.org/web/20171005181503/http://json.codeplex.com/workitem/23853).

I looked around and unfortunately there doesn't seem to be a clean way to change this behavior - so manually going through and updating the property names like you have is probably our best solution.

Charles-Gagnon (approved):

src/SqlAsyncCollector.cs
```diff
@@ -315,6 +315,12 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
                 else
                 {
                     // ToDo: add check for duplicate primary keys once we find a way to get primary keys.
+                    // JObjects ignore serializer settings (https://web.archive.org/web/20171005181503/http://json.codeplex.com/workitem/23853)
+                    // so we have to manually convert property names to lower case before inserting into the query in that case
```
chlafreniere, line 319: What about case sensitive collations?

chlafreniere: (even if this works, clarifying that it works in a comment would be helpful to readers)

Charles-Gagnon: Case sensitive has always been up to the caller to get right - we have no way of knowing what the expected case is. Case insensitive is just always converted to lower case for comparisons for convenience (since the casing still has to be the same, but we can control that for case insensitive)

https://github.com/Azure/azure-functions-sql-extension/blob/main/src/SqlAsyncCollector.cs#L526

chlafreniere (approved):

src/SqlAsyncCollector.cs
chlafreniere: Thanks @Charles-Gagnon

PR #251, opened 2022-05-17
Bump .NET SDK to 6.0.300 and add rollForward

Charles-Gagnon:
Roll forward to latest feature to pick up security fixes and other improvements.


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "6.0.102"
+    "version": "6.0.300"
```
chlafreniere, line 3: https://docs.microsoft.com/en-us/dotnet/core/tools/global-json#examples

We may have talked about this previously, but what would you think about adding "rollForward": "latestFeature"? I just get annoyed when having to have the exact version of the sdk every time that we update 😄

Charles-Gagnon: I'm not strictly opposed, but in general I've been approaching this stuff as "lock everything to a known version to prevent unexpected changes/breaks based on the environment". So stuff like this, package locks, etc. 

I do feel like it's probably a bit too strict for the .NET SDK though so would be fine with that if you would like to go that route.

Charles-Gagnon: Eh, thinking about it feature version updates are typically security fixes so it'd be better to always grab the latest one of those I think. We can hopefully trust the .NET team to do the right thing here.

chlafreniere (approved):

PR #255, opened 2022-05-20
Add copyright to JS samples

Charles-Gagnon:
And add linting to the build for pipeline verification


Charles-Gagnon:
@lucyzhang929 If you have some spare time to look into adding a similar pylint thing for the Python samples that would be nice (although not super important if you're busy)

lucyzhang929 (approved):

PR #260, opened 2022-06-03
Update AddProduct.cs

azuregomez:
With VS2022, Functions Runtime 4.3.2.18186 and .net 6, it is required to include CommandType; otherwise there is an error thrown.


dzsquared:
@azuregomez  -  this is a bug in Microsoft.NET.Sdk.Functions - now being tracked here in #263

PR #262, opened 2022-06-06
Add SupportedLanguages Attribute to run tests against all supported langs.

MaddyDev:
Added the following : SqlInlineDataAttributebase class overriding the xunit DataAttribute and has the GetData method implemented that adds the supported languages data to the tests.
UnsupportedLanguagesAttribute class that has the list of language(s) to be excluded from test runs.

For adding new languages following steps are required:
1. Add a Copy step to the CopySamples Target in the csprpj file to copy the language specific samples to the output bin under <b>SqlExtensionSamples\\\<languageName\></b> folder.
2. Add the language to the SupportedLanguages  enum.

### DIFF
--- a/test/.editorconfig
+++ b/test/.editorconfig
@@ -6,4 +6,3 @@
 dotnet_diagnostic.CA1309.severity = silent # Use ordinal StringComparison - this isn't important for tests and just adds clutter
 dotnet_diagnostic.CA1305.severity = silent # Specify IFormatProvider - this isn't important for tests and just adds clutter
 dotnet_diagnostic.CA1707.severity = silent # Identifiers should not contain underscores - this helps make test names more readable
-dotnet_diagnostic.CA1051.severity = silent # Do not declare visible instance fields - this helps avoid creating properties when not needed and avoid clutter
--- a/test/.editorconfig
+++ b/test/.editorconfig
@@ -6,3 +6,4 @@
 dotnet_diagnostic.CA1309.severity = silent # Use ordinal StringComparison - this isn't important for tests and just adds clutter
 dotnet_diagnostic.CA1305.severity = silent # Specify IFormatProvider - this isn't important for tests and just adds clutter
 dotnet_diagnostic.CA1707.severity = silent # Identifiers should not contain underscores - this helps make test names more readable
+dotnet_diagnostic.CA1051.severity = silent # Do not declare visible instance fields - this helps avoid creating properties when not needed and avoid clutter
--- a/test/Common/SupportedLanguagesTestAttribute.cs
+++ b/test/Common/SupportedLanguagesTestAttribute.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
     /// </summary>
     public class SqlInlineDataAttribute : DataAttribute
     {
-        protected readonly Dictionary<string, List<List<object>>> testData = new Dictionary<string, List<List<object>>>();
-        protected readonly string _propertyName = "data";
+        private readonly Dictionary<string, List<List<object>>> testData = new Dictionary<string, List<List<object>>>();
+        private readonly string _propertyName = "data";
 
         /// <summary>
         /// Load supported languages as test parameters data for the test.
--- a/test/Common/SupportedLanguagesTestAttribute.cs
+++ b/test/Common/SupportedLanguagesTestAttribute.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
         /// <param name="args">The test parameters to insert inline for the test</param>
         public SqlInlineDataAttribute(params object[] args)
         {
+            // Add each supported language to the args and add the newly created list of args
+            // with the language parameter to the testData
             foreach (SupportedLanguages lang in Enum.GetValues(typeof(SupportedLanguages)))
             {
                 var listOfValues = new List<object>();
--- a/test/Common/SupportedLanguagesTestAttribute.cs
+++ b/test/Common/SupportedLanguagesTestAttribute.cs
@@ -1,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+using System;
+using System.Collections.Generic;
+using Xunit.Sdk;
+using System.Reflection;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
+{
+    public abstract class LanguageTestAttribute : DataAttribute
+    {
+        protected enum SupportedLanguages
+        {
+            CSharp
+            /*JavaScript,
+            Python,
+             TypeScript */
+        }
+        protected readonly Dictionary<string, List<List<object>>> testData = new Dictionary<string, List<List<object>>>();
+        protected readonly string _propertyName = "data";
+
+        /// <inheritDoc />
+        public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+        {
+            if (testMethod == null) { throw new ArgumentNullException(nameof(testMethod)); }
+
+            if (this.testData.Count == 0) { throw new EmptyException(nameof(testMethod)); }
+            string serializedData = JsonConvert.SerializeObject(this.testData);
+            var allData = JObject.Parse(serializedData);
+            JToken data = allData[this._propertyName];
+            return data.ToObject<List<object[]>>();
+        }
+
+    }
+    public class SupportedLanguagesAttribute : LanguageTestAttribute
+    {
+        /// <summary>
+        /// Load supported languages for inline data for a theory
+        /// </summary>
+        public SupportedLanguagesAttribute()
+        {
+            var langData = new List<List<object>>();
+            foreach (string lang in Enum.GetNames(typeof(SupportedLanguages)))
+            {
+                var listOfValues = new List<object>(1) { lang };
+                langData.Add(listOfValues);
+            }
+            this.testData.Add(this._propertyName, langData);
+        }
+        /// <summary>
+        /// Load data from a inline data for a theory
+        /// </summary>
+        /// <param name="n">The number of products to insert for the test</param>
+        /// <param name="cost">The cost of the products for the test</param>
+        public SupportedLanguagesAttribute(int n, int cost)
+        {
+            var langData = new List<List<object>>();
+            foreach (string lang in Enum.GetNames(typeof(SupportedLanguages)))
+            {
+                var listOfValues = new List<object>(3) { n, cost, lang };
+                langData.Add(listOfValues);
+            }
+            this.testData.Add(this._propertyName, langData);
+        }
+
+        /// <summary>
+        /// Load data from a inline data for a theory
+        /// </summary>
+        /// <param name="id">The id of products for the test</param>
+        /// <param name="name">The name of products for the test</param>
+        /// <param name="cost">The cost of the product for the test</param>
+        public SupportedLanguagesAttribute(int id, string name, int cost)
+        {
+            var langData = new List<List<object>>();
+            foreach (string lang in Enum.GetNames(typeof(SupportedLanguages)))
+            {
+                var listOfValues = new List<object>(4) { id, name, cost, lang };
+                langData.Add(listOfValues);
+            }
+            this.testData.Add(this._propertyName, langData);
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


test/.editorconfig
```diff
@@ -5,4 +5,5 @@
 # Disabled
 dotnet_diagnostic.CA1309.severity = silent # Use ordinal StringComparison - this isn't important for tests and just adds clutter
 dotnet_diagnostic.CA1305.severity = silent # Specify IFormatProvider - this isn't important for tests and just adds clutter
-dotnet_diagnostic.CA1707.severity = silent # Identifiers should not contain underscores - this helps make test names more readable
\ No newline at end of file
+dotnet_diagnostic.CA1707.severity = silent # Identifiers should not contain underscores - this helps make test names more readable
+dotnet_diagnostic.CA1051.severity = silent # Do not declare visible instance fields - this helps avoid creating properties when not needed and avoid clutter
```
Charles-Gagnon, line 9: Just create auto-properties instead 

`public string MyProp { get; set; }`

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            if (this.testData.Count == 0) { throw new EmptyException(nameof(testMethod)); }
+            string serializedData = JsonConvert.SerializeObject(this.testData);
+            var allData = JObject.Parse(serializedData);
+            JToken data = allData[this._propertyName];
+            return data.ToObject<List<object[]>>();
+        }
+
+    }
+    public class SupportedLanguagesAttribute : LanguageTestAttribute
```
Charles-Gagnon, line 41: As we discussed, we only want one attribute. By default all tests should be running against all languages, and then we should have an `UnsupportedLanguages` attribute that tells it when there are languages not supported for a test currently. 

Looking at this implementation I think we should refactor this slightly - have a single attribute 

`public class SqlInlineData extends DataAttribute`

The languages enum would be a [Flag](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-6.0) enumeration (and I'd suggest renaming it to just something like `TargetLanguage`)

Then there'd be two constructors for SqlInlineData : 

`public SqlInlineData(TargetLanguage unsupportedLanguages, params object[] args)`
`public SqlInlineData(params object[] args)`

Tests that supported all languages would use the second, ones that wanted to exclude languages would use the first (using the flag syntax, e.g. `TargetLanguage.Python | TargetLanguage.Javascript`)

Examples : 

`[SqlInlineData()]` - supports all languages, no params
`[SqlInlineData(3, "foo")]` - supports all languages, has params
`[SqlInlineData(TargetLanguage.Python)]` - doesn't support Python, no params
`[SqlInlineData(TargetLanguage.Python, 3, "foo")]` - doesn't support Python, has params

Charles-Gagnon (changes requested):

test/Common/SupportedLanguagesTestAttribute.cs
Charles-Gagnon: Actually, looking at DataAttribute it might be better to make two attributes, `SqlInlineData` and `UnsupportedLanguages`. SqlInlineData would extend DataAttribute and then use the MethodInfo passed in to `GetData` to look for `UnsupportedLanguages` attribute on that method - and if so read it to get the list of unsupported languages. 

So it'd look like this on the test method

```
[SqlInlineData(3, "foo")]
[UnsupportedLanguages(TargetLanguage.Python, TargetLanguage.Javascript)]
```

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            CSharp
+        }
+        protected readonly Dictionary<string, List<List<object>>> testData = new Dictionary<string, List<List<object>>>();
+        protected readonly string _propertyName = "data";
+
+        /// <inheritDoc />
+        public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+        {
+            if (testMethod == null) { throw new ArgumentNullException(nameof(testMethod)); }
```
Charles-Gagnon, line 31: Are you copying this code from somewhere? It seems overly complex for what we need to do - especially with all the JSON parsing. But let's address the other issues and then look at this closer if this still ends up being necessary.

test/Common/SupportedLanguagesTestAttribute.cs
MaddyDev: So simply have one attribute class and remove the rest. Sounds good to me. But can I suggest a change, instead of having the list of unsupported ones, what if I make the logic for support-specific-language(s). 

Examples 

[SqlInlineData(TargetLanguage.Python)] - supports only Python, no params
[SqlInlineData(TargetLanguage.Python, 3, "foo")] - supports only Python, has params

Usually it's more like a test is specific for a language and is not applicable to others so having it the above way reflects that.

Charles-Gagnon: I'd rather not, we should expect that most of the tests support all languages (that should be the goal when writing tests). So it should be rarer to need to specify unsupported languages. 

As long as it's understood that tests by default using the SqlInlineData run against all languages that should be fine.

Charles-Gagnon: And to clarify - I think it'd be best to have two attributes. `SqlInlineData` (which is only used to specify the parameters if they exist) and `UnsupportedLanguages` (which only exists if the test doesn't support some languages)

test/Common/SupportedLanguagesTestAttribute.cs
MaddyDev: I referenced a blog that extends the DataAtrribute but this step is specific to us, the blog reads data from file so just parsing and returning. But in our case it's in a list so added the serialization before I parse.

test/Common/SupportedLanguagesTestAttribute.cs
MaddyDev: For some reason my browser didn't refresh your second suggestion of having two attributes (reading the UnsupportedLanguages attribute from method info part). Ok, I still think the support-specific-languages will work better and is a simpler implementation but having it as you suggested also works with a little bit more work. I'll get working on it. Thanks!

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        private readonly Dictionary<string, List<List<object>>> testData = new Dictionary<string, List<List<object>>>();
+        private readonly string _propertyName = "data";
+
+        /// <summary>
+        /// Load supported languages as test parameters data for the test.
+        /// </summary>
+        public SqlInlineDataAttribute()
+        {
+            var langData = new List<List<object>>();
```
Charles-Gagnon, line 27: You shouldn't need this - the other constructor will just be given an empty array if not parameters are specified.

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,103 @@
...
+            foreach (string lang in Enum.GetNames(typeof(SupportedLanguages)))
+            {
+                var listOfValues = new List<object>(1) { lang };
+                langData.Add(listOfValues);
+            }
+            this.testData.Add(this._propertyName, langData);
+
+        }
+        /// <summary>
+        /// Load the test parameters data for all supported languages for inline use of the test.
```
Charles-Gagnon, line 37: `Adds a language parameter to the test data which will contain the language that the test is currently running against from the list of supported languages`

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,103 @@
...
+        public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+        {
+            if (testMethod == null) { throw new ArgumentNullException(nameof(testMethod)); }
+            if (this.testData.Count == 0) { throw new EmptyException(nameof(testMethod)); }
+
+            UnsupportedLanguagesAttribute unsupportedLangAttr = testMethod.GetCustomAttribute<UnsupportedLanguagesAttribute>();
+            if (unsupportedLangAttr != null)
+            {
+                var unsupportedLanguages = new List<string>();
+                unsupportedLanguages.AddRange(unsupportedLangAttr.UnsuppportedLanguages);
```
Charles-Gagnon, line 66: The UnsupportedLanguagesAttribute already has UnsupportedLanguages defined as a List - why do we need to create a new list here?

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,103 @@
...
+                    List<List<object>> dataList = this.testData.GetValueOrDefault(this._propertyName);
+                    dataList.RemoveAll(l => l.Contains(lang));
+                    this.testData[this._propertyName] = dataList;
+                }
+            }
+
+            string serializedData = JsonConvert.SerializeObject(this.testData);
+            var allData = JObject.Parse(serializedData);
+            JToken data = allData[this._propertyName];
+            return data.ToObject<List<object[]>>();
```
Charles-Gagnon, line 78: I'm pretty sure you don't need all this JSON parsing stuff. This method returns an IEnumerable of object arrays - so basically it returns an enumerable where every item is the parameters to pass to the test method. 

See https://hovermind.com/xunit/custom-data-attribute.html for examples. 

In our case all we're doing is adding the language to the list of args passed in. Each attribute will have a single set of args passed in, and then we iterate over the list of languages supported (filtering out the unsupported ones) and for each language we return an array that consists of the original args + the language.

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,103 @@
...
+    [AttributeUsage(AttributeTargets.Method)]
+    public class UnsupportedLanguagesAttribute : Attribute
+    {
+        public List<string> UnsuppportedLanguages { get; set; } = new List<string>();
+        /// <summary>
+        /// Load only supported languages excluding the ones from the provided parameters.
+        /// </summary>
+        public UnsupportedLanguagesAttribute(params SupportedLanguages[] argsWithUnsupportedLangs)
+        {
+            foreach (SupportedLanguages s in argsWithUnsupportedLangs)
```
Charles-Gagnon, line 91: Let's just simplify this and have this attribute (and the SqlInlineData) keep the list of enum values directly. That way we're guaranteeing what the values are going to be for anything that might be using it. The StartFunctionHost method can take in the enum value and then convert it to a string for using in the path.

Charles-Gagnon (changes requested):

Charles-Gagnon (approved):

Charles-Gagnon:
It would be good to update the description of the PR as well though so if someone comes back and looks at it they aren't confused by it being out of date. The description should always be kept up to date with the current state of the PR, changing it as necessary if you make changes in the PR itself

test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            foreach (SupportedLanguages lang in Enum.GetValues(typeof(SupportedLanguages)))
+            {
+                var listOfValues = new List<object>();
+                foreach (object val in args)
+                {
+                    listOfValues.Add(val);
+                }
+                listOfValues.Add(lang);
+                this.testData.Add(listOfValues.ToArray());
```
chlafreniere, line 34: I'm having trouble understanding why we're adding both `lang` and `val` to listOfValues just from reading this. Can you clarify?

MaddyDev: We have inline data coming in the form of [ [a, b, 3], [d, e, 6]...] and we need to translate that to each language -> 
[ [a, b, 3, CSharp], [a, b, 3, JavaScript], [a, b, 3, Python], ...,
  [d, e, 6, CSharp], [d, e, 6, JavaScript], [d, e, 6, Python], ....
  ...
]
So creating a list of values and adding each supported language. Hope that explains it better. Let me know if you need more clarity.

chlafreniere: Yeah, I hear ya. Can you add more comments in the code?

test/Integration/IntegrationTestBase.cs
```diff
@@ -78,7 +78,7 @@ private void SetupDatabase()
             string testServer = Environment.GetEnvironmentVariable("TEST_SERVER");
             if (string.IsNullOrEmpty(testServer))
             {
-                testServer = "localhost";
+                testServer = "localhost\\SQL2019";
```
chlafreniere, line 81: was this intentional?

MaddyDev: oops, pushed the wrong file

chlafreniere (approved):

PR #269, opened 2022-06-14
Enable tests for JavaScript lang

MaddyDev:
Added the JavaScript language to the SupportedLanguages.

For the negative test cases, added them to folder samples-js-extra in the test project and copied them over to the output bin JavaScript folder in the post build task.
Excluded the JavaScript language from tests that are not relevant and created the missing ones to samples-js.


test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -123,6 +123,7 @@ public void AddProductArrayTest(SupportedLanguages lang)
 
         [Theory]
         [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.JavaScript)]
```
Charles-Gagnon, line 126: For all these add comments explaining why they're unsupported

test/samples-js-extra/AddProductExtraColumns/function.json
```diff
@@ -0,0 +1,27 @@
+{
```
Charles-Gagnon, line 1: Let's put these under `test/Integration/test-js` so we're consistent with where we store the extra test functions. 

And then while we're doing that if you could make a `test/Integration/test-csharp` and put the C# ones there that'd be great

test/Common/TestUtils.cs
```diff
@@ -149,5 +149,14 @@ public static void Retry(Action action, int retryCount = 3, int waitDurationMs =
...
             }
         }
+
+        /// <summary>
+        /// Removes spaces, tabs and new lines from the JSON string.
+        /// </summary>
+        /// <param name="jsonStr">The json string for trimming</param>
+        public static string TrimJSONString(string jsonStr)
+        {
+            return jsonStr.Trim().Replace(" ", "").Replace(Environment.NewLine, "");
```
Charles-Gagnon, line 159: I would suggest trying out [Assert.Equivalent](https://github.com/xunit/xunit/issues/1604) (and then parse these into actual JObjects) - you'll need to update to use the 2.4.2-pre.12 version of xunit for that.

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
```diff
@@ -23,6 +23,21 @@
     </ItemGroup>
     <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\CSharp\%(RecursiveDir)" />
     <Message Text="Copied Samples output to $(OutDir)\SqlExtensionSamples" Importance="high" />
+    <ItemGroup>
+      <_CopyItems Include="..\samples\samples-js\**\*.*" />
+    </ItemGroup>
+    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\JavaScript\%(RecursiveDir)" />
```
Charles-Gagnon, line 29: Did you verify that this isn't going to copy the csharp stuff in here too? I'm not sure offhand whether _CopyItems will be overwritten each time, but just to be safe (and clearer) I'd suggest making each of these have a different name (so `_CSharpCopyItems`, `_JsCopyItems`, etc.) 

And you should only need to have 1 Copy task for each (so 1 for CSharp, 1 for JS, etc.). This should work 

```
    <ItemGroup>
      <_JsCopyItems Include="host.json" />
      <_JsCopyItems Include="..\samples\samples-js\**\*.*" />
      <_JsCopyItems Include="Integration/test-js\**\*.*" />
    </ItemGroup>
    <Copy SourceFiles="@(_JsCopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\JavaScript\" />
```

test/samples-js-extra/AddProductMissingColumns/index.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async function (context) {
+    context.log('JavaScript HTTP trigger function processed a request.');
```
Charles-Gagnon, line 2: Remove these logging statements, we don't need these in our tests/samples

test/samples-js-extra/AddProductExtraColumns/index.js
```diff
@@ -0,0 +1,18 @@
+module.exports = async function (context) {
```
Charles-Gagnon, line 1: For these failure scenario functions please copy over the comments explaining what the function is verifying

test/samples-js-extra/AddProductMissingColumnsExceptionFunction/function.json
```diff
@@ -0,0 +1,27 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "post"
```
Charles-Gagnon, line 9: I noticed that these are "post" here but only support "get" in the C# ones, a couple things :

1. Check why the tests would be passing (they are both passing right?) - the test should only do either a POST or a GET, which means I'd expect one or the other to be failing
2. Update the C# ones to be POST since that should be our preference for output bindings

test/samples-js-extra/AddProductsNoPartialUpsert/function.json
```diff
@@ -0,0 +1,28 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "direction": "in",
+      "type": "httpTrigger",
+      "methods": [
+        "get",
```
Charles-Gagnon, line 9: Just POST

test/samples-js-extra/AddProductsNoPartialUpsert/index.js
```diff
@@ -0,0 +1,24 @@
+module.exports = async function (context) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    var products = [];
+    for(let i = 0; i < 1000; i++) {
+        products.add({"productId": i,
+        "name": "test"+i,
```
Charles-Gagnon, line 7: In general if you're going to make functionality changes from the "original" ones they should be done explicitly. For example here the name is `test#` but in the C# ones it's only going to be `test` for every item. 

The implementations should be as exact as possible - so either update the "original" tests/samples/helpers with the new functionality or keep the implementation exactly the same across all supported languages.

test/samples-js-extra/AddProductsNoPartialUpsert/index.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    const invalidProduct = {
+        "productid": 1000,
+        "name": null,
+        "cost": 1000
+    };
+
+    products.add(invalidProduct);
+
+    context.bindings.products = JSON.stringify(products);
```
Charles-Gagnon, line 18: It might be worth it to note that this isn't really testing the same thing as the C# version since JS (and most of the other languages) don't support the ICollector style of streaming uploads (at least as far as I know)

This test is still fine to have for additional verification but it does have different behavior which should be called out.

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
```diff
@@ -19,10 +19,25 @@
...
+    </ItemGroup>
+    <Copy SourceFiles="@(_JsCopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\JavaScript\%(RecursiveDir)" />
+    <Message Text="Copied JS Samples output to $(OutDir)\samples-js" Importance="high" />
+    <ItemGroup>
+      <_HostCopyItems Include="host.json" />
+    </ItemGroup>
+    <Copy SourceFiles="@(_HostCopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\JavaScript\" />
+    <Message Text="Copied host.json to $(OutDir)\SqlExtensionSamples\JavaScript" Importance="high" />
+    <ItemGroup>
+      <_JsCopyItems Include="Integration\test-js\**\*.*" />
```
Charles-Gagnon, line 37: Did you try combining these into a single ItemGroup with a single copy? Right now I believe you'll be copying the samples-js content multiple times since you're just adding items to the _JsCopyItems list.

test/Integration/SqlOutputBindingIntegrationTests.cs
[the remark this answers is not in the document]
MaddyDev: For the Collectors test, since it's chsarp language specific, I can add that. Can you help me on what to add for the QueryTrigger and TimerTrigger tests, am not sure what to add except they aren't valid tests for JavaScript.

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
MaddyDev: Wasn't sure if I can add more than one copy items in an itemGroup, will try, thanks

test/host.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
MaddyDev: yes, but probably will be used by other languages as well, so kept it under the root.

samples/samples-csharp/local.settings.json
```diff
@@ -2,6 +2,7 @@
   "IsEncrypted": false,
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "AzureWebJobsSecretStorageType": "files",
```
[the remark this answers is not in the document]
MaddyDev: Ohh, pushed this by mistake, on my local dev machine, C# tests are failing with "Azure.Storage.Blobs: Service request failed" error and a quick online search suggested I add this and that did the magic.

test/Integration/SqlOutputBindingIntegrationTests.cs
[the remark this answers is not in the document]
Charles-Gagnon: Why aren't they valid? They both support JS as well (I'm assuming you meant QueueTrigger, not QueryTrigger) :

 https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer?tabs=in-process&pivots=programming-language-javascript
 
 https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue-trigger?tabs=in-process%2Cextensionv5&pivots=programming-language-javascript

test/Common/TestUtils.cs
MaddyDev: I tried that but even then it fails since the javascript properties begin with lowercase and c# properties with caps.

Charles-Gagnon: You can lowercase the JSON string before converting it in to an object

MaddyDev: if casing is needed either way then can I leave what I've instead of updating it?

lucyzhang929 (approved):

test/Common/TestUtils.cs
Charles-Gagnon: What do the failures look like? Is it clear where exactly the difference is? The idea behind using the actual objects is it allows for a much more detailed error since it knows exactly what properties didn't match. But if the error message message for strings is clear enough I guess it's fine...

But if you do change the name to `CleanJsonString` - "trimming" is a well-known term that refers specifically to remove values on the ends of a string and so the fact that this is also removing all spaces wouldn't fit the name of the method as it is currently.

MaddyDev: Assert.Equal() Failure
Expected: [{"ProductID":1,"Name":"test","Cost":100}]
Actual:   [
  {
    "ProductId": 1,
    "Name": "test",
    "Cost": 100
  }
]
So unless I lowercase and check it won't work since ProductID and ProductId are different.

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -123,6 +123,8 @@ public void AddProductArrayTest(SupportedLanguages lang)
...
+        // Collectors are CSharp specific and are not relevant for JavaScript
```
Charles-Gagnon, line 127: Put these kinds of comments on the same line as the `UnsupportedLanguages` attribute so it's clear that's what they're tied to.

I'd also suggest making this comment more generic so we don't have to update it for future languages - `Collectors are only available in C#`

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
```diff
@@ -19,10 +19,16 @@
 
   <Target Name="CopySamples" AfterTargets="Build">
     <ItemGroup>
-      <_CopyItems Include="..\samples\samples-csharp\bin\$(Configuration)\$(TargetFramework)\**\*.*" />
+      <_CSharpCopyItems Include="..\samples\samples-csharp\bin\$(Configuration)\$(TargetFramework)\**\*.*" />
     </ItemGroup>
-    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\CSharp\%(RecursiveDir)" />
+    <Copy SourceFiles="@(_CSharpCopyItems)" DestinationFolder="$(OutDir)\SqlExtensionSamples\CSharp\%(RecursiveDir)" />
     <Message Text="Copied Samples output to $(OutDir)\SqlExtensionSamples" Importance="high" />
```
Charles-Gagnon, line 25: `Copied C# Samples output to $(OutDir)\SqlExtensionSamples\CSharp`

Charles-Gagnon (approved):
Some minor things to look at, but otherwise looks good. 

Make sure everything is passing correctly though - it seemed to have test failures last time I checked.

Issue #271, opened 2022-06-20
SELECT permission is required for output SQL binding


springcomp:
I’m using the _output_ [SQL binding](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql?tabs=in-process%2Cextensionv3&pivots=programming-language-csharp) for Azure Functions bound to an `IAsyncCollector<T>` type parameter.

```c#
[FunctionName("InsertReportingToDataBase")]
public async Task<IActionResult> Run(
  [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
  [Sql("dbo.Tickets", ConnectionStringSetting = "SQLAzure:DefaultConnection")] IAsyncCollector<TicketDto> records,
  ILogger log)
  {
    await records.AddAsync(new TicketDto { … });
    await records.FlushAsync();
  }
```

The Function App’s _system-assigned managed identity_ is configured with proper permissions against the database and corresponding database objects. However, it seems the `SELECT` permission is required as the function raises the following error:

![image](https://user-images.githubusercontent.com/8488398/174618342-59dce599-03ed-4d82-a1f9-3350fddcb0d7.png)

```
The SELECT permission was denied on the object 'tickets', database 'db-name', schema 'dbo'.
```

I would expect the SELECT permission to **not** be required when performing insert operations against a database object.

PR #272, opened 2022-06-20
Add note about output requiring select perms

Charles-Gagnon:
Customer asked about this in https://github.com/Azure/azure-functions-sql-extension/issues/271 - we should make it clearly documented. It would be good to go through and get the full list of permissions required for each type of binding to run, but this is a good starting point.


MaddyDev (approved):

Charles-Gagnon:
@dzsquared @lucyzhang929 This seem reasonable to add? Should we add it to the MS docs as well?

dzsquared:
> @dzsquared @lucyzhang929 This seem reasonable to add? Should we add it to the MS docs as well?

ack

lucyzhang929 (approved):

PR #284, opened 2022-07-29
Replace non-breaking spaces with regular spaces

JatinSanghvi:
The pull request replaces [non-breaking spaces](https://en.wikipedia.org/wiki/Non-breaking_space) with regular [spaces](https://en.wikipedia.org/wiki/Space_(punctuation)). The presence of non-breaking spaces is causing confusion when someone searches for text like `CREATE TABLE Employees` (with regular spaces) in Visual Studio Code and gets no results back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,21 +68,21 @@ This will require connecting to and running queries - you can use [Azure Data St
     Otherwise connect to your database and run the following query to create a simple table to start with.
 
 ```sql
-CREATE TABLE Employees (
-        EmployeeId int,
-        FirstName varchar(255),
-        LastName varchar(255),
-        Company varchar(255),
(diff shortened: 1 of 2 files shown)


Charles-Gagnon (approved):

Charles-Gagnon:
There's that weird error again, I pinged the guardian team about it but they weren't very helpful unfortunately...anyways we'll just have to live with the failures for now.

Charles-Gagnon:
Although you should have contributor permission to the repo so you could just work from it directly instead of using a fork - that seems to be the problem here.

JatinSanghvi:
Thanks for the approval. I will create branches in the main repo for future pull requests.

PR #288, opened 2022-08-01
Add trigger binding support - Part 2

JatinSanghvi:
As it was discussed a while back, starting with fresh pull request targeting `triggerbindings` branch. This will also keep the discussion focused on new set of changes. The intention is to only update `triggerbindings` branch through commits from pull requests or through merges from `main` branch.

**PR changelog**
1. Improve error handling when user table has primary column with conflicting names _(addressed review comment from #226)_.
2. Add comment for why trigger-binding provider should return null binding in particular case _(addressed review comment from #226)_.
3. Simplify SQL variable names (no need to use hash of column names).
4. Make renew-leases operation a single SQL statement _(addressed review comment from #226)_.
5. Have higher locking granularity instead of from `TABLOCKX`.

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -41,7 +40,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private readonly string _workerTableName;
         private readonly IReadOnlyList<string> _userTableColumns;
         private readonly IReadOnlyList<string> _primaryKeyColumns;
-        private readonly IReadOnlyList<(string col, string hash)> _primaryKeyColumnHashes;
         private readonly IReadOnlyList<string> _rowMatchConditions;
         private readonly ITriggeredFunctionExecutor _executor;
         private readonly ILogger _logger;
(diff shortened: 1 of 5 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -595,7 +578,7 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
                     c.SYS_CHANGE_VERSION, c.SYS_CHANGE_OPERATION,
                     w.ChangeVersion, w.AttemptCount, w.LeaseExpirationTime
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
-                LEFT OUTER JOIN {this._workerTableName} AS w WITH (TABLOCKX) ON {workerTableJoinCondition}
+                LEFT OUTER JOIN {this._workerTableName} AS w ON {workerTableJoinCondition}
```
MaddyDev, line 581: why is this updated?

MaddyDev (approved):
@JatinSanghvi It would be great if you could add a section in the design doc about the locks you used and why.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Hi, after this change, I had chance to develop more understanding on the use of appropriate isolation level and locking modes, so I may need to update the lock modes further in this file. To get to some history, there were known deadlock issues in the SQL trigger support when our team took the ownership. Following my suggestion, Ameya had added table-level exclusive locks just to make sure that there are no more deadlocks. The next task was to selectively reduce the degree or locking and/or increase locking granularity and make sure the triggers are still deadlock-free. But this was not done then so I am trying to get it done this time.

Particularly in the above case, I did not find the use of taking table-level exclusive lock as this is a SELECT statement with no follow-up INSERT/UPDATE statement. After the change to lock modes, I tested with 3 worker applications running together and could not reproduce deadlocks.

JatinSanghvi:
Changes will be done through #296.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: I will be pushing the change to update the locking scheme in a future PR. This PR is closed, and the changes are separated into multiple PRs.

PR #289, opened 2022-08-05
Add telemetry events for trigger binding

JatinSanghvi:


src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -57,6 +62,12 @@ public SqlTriggerListener(string connectionString, string tableName, string user
             this._executor = executor;
             this._logger = logger;
             this._listenerState = ListenerNotStarted;
+
+            this._telemetryProps = new Dictionary<string, string>
```
lucyzhang929, line 66: Can you include the connection string properties? Example here:
https://github.com/Azure/azure-functions-sql-extension/blob/767a1e12c26142ef8af09bd1c2b642adcb3ca4b8/src/SqlAsyncCollector.cs#L505

Not sure if it is safe to or if we want to include the UserFunctionId and UserTableName. @MaddyDev Thoughts?

JatinSanghvi: User Function ID would be a hash (e.g., `6a85afa0dc8937d3`) so IMO, it should not be a problem.

JatinSanghvi: Worker table name would be equal to `Worker_` + <user-function-id> + <user-table-object-id>. In that sense, just logging worker table name would be enough to find the function ID, but we can keep it redundant for easy search and in case the logic to generate worker table name changes in future. I can log user table's object ID instead of table name if required.

JatinSanghvi: The SQL connection object is not available at the point where `this._props` is initialized. I tried to not initialize `props` every time a new connection is opened as it would basically be the same dictionary initialized every time. Also, in many cases where we are logging telemetry event, there is not connection involved. Since we are only logging SQL server version at the moment which does not change with connection. I can initialize a connection object (not for reuse) in the `SqlTriggerListener` constructor and call `.AsConnectionProps()` on it.

JatinSanghvi: Removed table name from telemetry and initializing properties table with `connection.AsConnectionProps()`.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -57,6 +62,14 @@ public SqlTriggerListener(string connectionString, string tableName, string user
             this._executor = executor;
             this._logger = logger;
             this._listenerState = ListenerNotStarted;
+
+            using (var connection = new SqlConnection(connectionString))
+            {
+                connection.Open();
```
lucyzhang929, line 68: Instead of opening a connection just to get the connection properties, can we add the connection properties to this._telemetryProps [here](https://github.com/Azure/azure-functions-sql-extension/blob/4a43881144f6f1b15451bb6821c2cbf2f747b4b4/src/TriggerBinding/SqlTriggerListener.cs#L110) after the connection is opened? It is okay if the first event (StartListenerStart) does not have have the connection properties.

JatinSanghvi: Makes sense. I will update the code.

lucyzhang929 (approved):

chlafreniere (approved):

PR #291, opened 2022-08-10
Throw exception for unsupported column types

JatinSanghvi:
Example exception that will get thrown if table has one of the types defined in assemblies such as `microsoft.sqlserver.types.dll`:

`System.InvalidOperationException: Found column(s) with unsupported type(s): 'BirthPlace' (type: geography), 'PersonGeometry' (type: geometry) in table: 'dbo.ColumnTypeTest'`.


lucyzhang929 (approved):
Tested and verified that there are no issues with telemetry using the new older AppInsights version.

PR #294, opened 2022-08-17
Add benchmark project for performance tests

lucyzhang929:
Running `dotnet run` from the benchmark folder calls the main method in SqlBindingBenchmarks.cs which uses BenchmarkDotnet to run the benchmarks in SqlInputBindingBenchmarks.cs and SqlOutputBindingBenchmarks.cs. Both are based on IntegrationTestBase.cs 
I made a couple of changes to the IntegrationTestBase class:
- added a check for AZURE_SQL_DB_CONNECTION_STRING environment variable
- moved over some methods from SqlInputBindingIntegrationTests.cs and SqlOutputBindingIntegrationTests.cs so that they can be used by the benchmark classes also.
- log to Console if ITestOutputHelper is not passed in


benchmark/Microsoft.Azure.WebJobs.Extensions.Sql.Benchmark.csproj
```diff
@@ -0,0 +1,36 @@
...
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" />
+  </ItemGroup>
+
+  <ItemGroup>
+	<ProjectReference Include="..\test\Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj" />
+	<ProjectReference Include="..\samples\samples-csharp\Microsoft.Azure.WebJobs.Extensions.Sql.Samples.csproj" />
+  </ItemGroup>
+
+  <Target Name="CopySamples" AfterTargets="Build">
```
Charles-Gagnon, line 19: Let's make a targets file for these shared targets since we're using them in more than one project now

benchmark/README.md
```diff
@@ -0,0 +1,15 @@
+# Running Performance Tests
```
Charles-Gagnon, line 1: Since these are perf tests how about we name the folder `performance`. 

And actually - I feel like maybe we should just put this under the test folder - so `test/Performance`

benchmark/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Pre-requisites
+The performance tests are based on the IntegrationTestBase class. Follow the instructions to set up the pre-requisites for integration tests [here](https://github.com/Azure/azure-functions-sql-extension/tree/main/test#running-integration-tests).
+
+## Run
+The performance tests use BenchmarkDotNet to benchmark performance for input and output bindings.
+
+Run the tests from the terminal.
```
Charles-Gagnon, line 9: Include step to cd to the proper directory

benchmark/SqlBindingBenchmarks.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Benchmark
+{
+    public class SqlBindingBenchmarks
+    {
+        public static void Main()
+        {
+            BenchmarkRunner.Run<SqlInputBindingBenchmarks>(
+                ManualConfig
+                    .Create(DefaultConfig.Instance)
+                    .WithOptions(ConfigOptions.DisableOptimizationsValidator));
```
Charles-Gagnon, line 16: Add comment about why you're doing this

test/Integration/IntegrationTestBase.cs
```diff
@@ -57,7 +59,7 @@ public class IntegrationTestBase : IDisposable
         /// </summary>
         protected int Port { get; private set; } = 7071;
 
-        public IntegrationTestBase(ITestOutputHelper output)
+        public IntegrationTestBase(ITestOutputHelper output = null)
```
Charles-Gagnon, line 62: Make this and TestOutput nullable

test/README.md
```diff
@@ -14,7 +14,7 @@ Our integration tests are based on functions from the samples project. To run in
    ```
    npm install -g azurite
    ```
-3. A local SQL Server instance - This is used by tests to verify that data is correctly added/fetched from the database when a test Function is run. You just need the server to be up and running, the tests will create the database and tables which will be cleaned up afterwards.
+3. A local SQL Server instance or an Azure SQL Database - This is used by tests to verify that data is correctly added/fetched from the database when a test Function is run. You just need the server to be up and running, the tests will create the database and tables which will be cleaned up afterwards.
 
    - You can either have a SQL Server installation with `localhost` available for connection via integrated security, or
```
Charles-Gagnon, line 19: Instead of having these be bullet points let's make these sub-sections themselves with titles `Local Install`, `Docker Container`, and `Azure SQL Database`

test/Integration/IntegrationTestBase.cs
lucyzhang929: I think ITestOutputHelper is already nullable. When I add Nullable<ITestOutputHelper> I get the following error:
`The type 'ITestOutputHelper' must be a non-nullable value type in order to use it as parameter 'T' in the generic type or method 'Nullable<T>'`

benchmark/SqlBindingBenchmarks.cs
lucyzhang929: This config is actually not needed - will remove.

benchmark/README.md
lucyzhang929: Changed folder name to `performance`. BenchmarkDotnet needs to be run from it's own console project so I can't add this to the Tests project. We could nest the current Tests project into another folder and move performance under test so we would have `test/Performance` and `test/<current-tests-project>`

test/README.md
```diff
@@ -16,16 +16,21 @@ Our integration tests are based on functions from the samples project. To run in
...
-   - Start a SQL Server instance with Docker
-     ```
-     docker pull mcr.microsoft.com/mssql/server:2019-latest
-     docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD={your_password}" -e "MSSQL_PID=Express" -p 1433:1433 --name sql1 -h sql1 -d mcr.microsoft.com/mssql/server:2019-latest
-     ```
-     After the Docker image is running, you just need to set `SA_PASSWORD` environment variable to `{your_password}` and can run tests normally.
-     
-     Note: If `SA_PASSWORD` is not set, the tests will assume you're using a local MSSQL installation and default to using integrated auth. MSSQL on Docker does not support integrated auth by default.
-   - To use an Azure SQL Database, set the `AZURE_SQL_DB_CONNECTION_STRING` environment variable to your Azure SQL Database connection string.
+### Local Install
```
MaddyDev, line 19: I think you need to indent this to be part of step 3

test/Integration/IntegrationTestBase.cs
Charles-Gagnon: What I meant is `ITestOutputHelper? output` - but I forgot that since this is targeting netstandard2.0 we can't enable that. So nevermind!

performance/README.md
```diff
@@ -0,0 +1,16 @@
+# Running Performance Tests
+
+## Pre-requisites
+The performance tests are based on the IntegrationTestBase class. Follow the instructions to set up the pre-requisites for integration tests [here](https://github.com/Azure/azure-functions-sql-extension/tree/main/test#running-integration-tests).
```
Charles-Gagnon, line 4: Use a relative path link for this

performance/README.md
```diff
@@ -0,0 +1,16 @@
...
+The performance tests use BenchmarkDotNet to benchmark performance for input and output bindings.
+
+Run the tests from the terminal.
+```
+cd performance
+dotnet run -c Release
+```
+
+## Results
+The test results will be generated in the BenchmarkDotNet.Artifacts folder.
```
Charles-Gagnon, line 16: Add this (and any other generated artifacts) to .gitignore as well

test/Integration/IntegrationTestBase.cs
```diff
@@ -74,35 +76,44 @@ public IntegrationTestBase(ITestOutputHelper output)
         /// </remarks>
         private void SetupDatabase()
         {
-            // Get the test server name from environment variable "TEST_SERVER", default to localhost if not set
-            string testServer = Environment.GetEnvironmentVariable("TEST_SERVER");
-            if (string.IsNullOrEmpty(testServer))
+            SqlConnectionStringBuilder connectionStringBuilder;
+            string connectionString = Environment.GetEnvironmentVariable("AZURE_SQL_DB_CONNECTION_STRING");
```
Charles-Gagnon, line 80: This doesn't really seem to need to be specific to azure, but rather just allow for the user to specify the full connection string. So I'd just call this `SQL_CONNECTION_STRING` or `TEST_CONNECTION_STRING`

test/README.md
```diff
@@ -14,18 +14,24 @@ Our integration tests are based on functions from the samples project. To run in
    ```
    npm install -g azurite
    ```
-3. A local SQL Server instance - This is used by tests to verify that data is correctly added/fetched from the database when a test Function is run. You just need the server to be up and running, the tests will create the database and tables which will be cleaned up afterwards.
+3. A local SQL Server instance or an Azure SQL Database - This is used by tests to verify that data is correctly added/fetched from the database when a test Function is run. You just need the server to be up and running, the tests will create the database and tables which will be cleaned up afterwards.
```
Charles-Gagnon, line 17: I'd just update this to be `A SQL Server instance used by tests to insert/read data by the test functions.` It doesn't really matter what type of servers they are and that makes it easier to add new supported servers later on.

Charles-Gagnon (approved):

PR #296, opened 2022-08-18
Add new integration tests for trigger binding

### LINKED ISSUE #309: Improve error handling when user table has primary column with conflicting names

1. Let's make sure this is documented in the README
2. I'd suggest adding a bit more text to tell the user how to resolve it, I think it's a bit unclear right now what the user can do about it. Something like `This column must be removed or renamed for the table to be used by this trigger.`
3. I'd also make it find all of the columns that are reserved and list them, not just the first one

_Originally posted by @Charles-Gagnon in https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r933616315_

JatinSanghvi:
1. Closes #309
2. Add integration test to verify the behavior for various error conditions.


src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -409,7 +412,7 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
             IReadOnlyList<(string name, string type)> primaryKeyColumns,
             CancellationToken cancellationToken)
         {
-            string primaryKeysWithTypes = string.Join(",\n", primaryKeyColumns.Select(col => $"{col.name.AsBracketQuotedString()} [{col.type}]"));
+            string primaryKeysWithTypes = string.Join(", ", primaryKeyColumns.Select(col => $"{col.name.AsBracketQuotedString()} {col.type}"));
```
JatinSanghvi, line 415: Query looks better this way than one with unequally indented lines.

test/Integration/IntegrationTestBase.cs
```diff
@@ -188,26 +188,33 @@ protected void StartFunctionHost(string functionName, SupportedLanguages languag
...
+            taskCompletionSource.Task.Wait(60000);
+            this.TestOutput.WriteLine($"Azure Function host started!");
+            this.FunctionHost.OutputDataReceived -= SignalStartupHandler;
+
+            void SignalStartupHandler(object sender, DataReceivedEventArgs e)
             {
                 // This string is printed after the function host is started up - use this to ensure that we wait long enough
                 // since sometimes the host can take a little while to fully start up
-                if (e != null && !string.IsNullOrEmpty(e.Data) && e.Data.Contains($"http://localhost:{this.Port}/api"))
+                if (e.Data?.Contains(" Host initialized ") == true)
```
JatinSanghvi, line 213: Updated the log line that gets waited for to signal that the Functions Host has initialized. The old line works only for functions with HTTP triggers. As more tests are getting added, waiting for 60 seconds for each test was making the overall running time for tests slower. I have verified that none of the tests fail or wait until timeout due to this change.

samples/samples-csharp/Database/Tables/ProductsWithReservedPrimaryKeyColumnNames.sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE TABLE [ProductsWithReservedPrimaryKeyColumnNames] (
```
Charles-Gagnon, line 1: These should go in https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Integration/test-csharp, the samples folder should be only for valid samples that we want customers to be looking at for their own purposes. Objects used for negative/failure tests should go in the test folder. 

(you'll probably have to update the logic to look for the SQL files in both places though, I don't think we do that currently since all our current tests use the sample tables)

samples/samples-csharp/TriggerBindingSamples/ReservedPrimaryKeyColumnNamesTrigger.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.TriggerBindingSamples
+{
+    public static class ReservedPrimaryKeyColumnNamesTrigger
+    {
+        [FunctionName(nameof(ReservedPrimaryKeyColumnNamesTrigger))]
+        public static void Run(
+            [SqlTrigger("[dbo].[ProductsWithReservedPrimaryKeyColumnNames]", ConnectionStringSetting = "SqlConnectionString")]
+            IReadOnlyList<SqlChange<Product>> _)
+        {
+            throw new NotImplementedException();
```
Charles-Gagnon, line 17: For these ones that aren't expected to ever trigger the function let's include a message about this not being valid so it's easy to tell why the test is failing if that happens. 

Something like `throw new NotImplementedException("Test should fail before function code is ran");`

(do this for all similar tests below)

samples/samples-csharp/TriggerBindingSamples/ReservedPrimaryKeyColumnNamesTrigger.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.TriggerBindingSamples
+{
+    public static class ReservedPrimaryKeyColumnNamesTrigger
```
Charles-Gagnon, line 10: For each of these functions please add a doc comment on the class explaining what scenario it's testing.

test/Integration/IntegrationTestBase.cs
```diff
@@ -188,26 +188,33 @@ protected void StartFunctionHost(string functionName, SupportedLanguages languag
...
             this.FunctionHost.ErrorDataReceived += this.TestOutputHandler;
 
             this.FunctionHost.Start();
             this.FunctionHost.BeginOutputReadLine();
             this.FunctionHost.BeginErrorReadLine();
 
-            var taskCompletionSource = new TaskCompletionSource<bool>();
-            this.FunctionHost.OutputDataReceived += (object sender, DataReceivedEventArgs e) =>
+            this.TestOutput.WriteLine($"Waiting for Azure Function host to start...");
+            taskCompletionSource.Task.Wait(60000);
```
Charles-Gagnon, line 205: No magic numbers, make this a constant value with a descriptive name

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
...
+            };
+
+            this.StartFunctionHost(functionName, Common.SupportedLanguages.CSharp, false, OutputHandler);
+            Task completedTask = await Task.WhenAny(tcs.Task, Task.Delay(TimeSpan.FromSeconds(5)));
+            Assert.Equal(tcs.Task, completedTask);
+            this.FunctionHost.OutputDataReceived -= OutputHandler;
+
+            // WebJobs SDK retries forever to start the trigger-listener until it succeeds, which makes the Functions
+            // Host process never exit by itself in case of an error.
+            Assert.False(this.FunctionHost.HasExited);
```
Charles-Gagnon, line 235: `Assert.True` and `Assert.False` allow a message parameter, please use that to give a descriptive failure message.

But also - why do we need this assert? If the process exits (maybe the host changes its behavior) then do we really care? We can just kill the host regardless so this assert doesn't seem useful.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
                 id += 1;
             }
         }
+
+        private async Task<string> AssertTriggerListenerError(string functionName)
```
Charles-Gagnon, line 211: Doc comment explaining what this method is doing

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
...
+                    string exceptionPrefix = "Exception: System.InvalidOperationException: ";
+                    int index = e.Data.IndexOf(exceptionPrefix, StringComparison.Ordinal);
+                    Assert.NotEqual(-1, index);
+                    errorMessage = e.Data[(index + exceptionPrefix.Length)..];
+                    tcs.SetResult(true);
+                }
+            };
+
+            this.StartFunctionHost(functionName, Common.SupportedLanguages.CSharp, false, OutputHandler);
+            Task completedTask = await Task.WhenAny(tcs.Task, Task.Delay(TimeSpan.FromSeconds(5)));
```
Charles-Gagnon, line 229: No magic numbers, make this a const with a descriptive name

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
...
+            string errorMessage = null;
+            var tcs = new TaskCompletionSource<bool>();
+
+            void OutputHandler(object sender, DataReceivedEventArgs e)
+            {
+                if (errorMessage == null && e.Data?.Contains("Failed to start SQL trigger listener") == true)
+                {
+                    string exceptionPrefix = "Exception: System.InvalidOperationException: ";
+                    int index = e.Data.IndexOf(exceptionPrefix, StringComparison.Ordinal);
+                    Assert.NotEqual(-1, index);
```
Charles-Gagnon, line 222: Why the assert here? That isn't going to really do anything since nothing is handling the exception, right? I'd suggest just doing an `if (index >= 0) {` for the following code instead.

samples/samples-csharp/TriggerBindingSamples/PrimaryKeyNotPresentTrigger.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.Samples.TriggerBindingSamples
+{
+    public static class PrimaryKeyNotPresentTrigger
```
lucyzhang929, line 10: Functions used to test error cases should go in the test/Integration/test-csharp folder.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon, line 211: `StartHostAndWaitForError` seems like a better name to clearly indicate what this is doing.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
...
+                    errorMessage = e.Data[(index + exceptionPrefix.Length)..];
+                    tcs.SetResult(true);
+                }
+            };
+
+            this.StartFunctionHost(functionName, Common.SupportedLanguages.CSharp, false, OutputHandler);
```
Charles-Gagnon, line 228: Make sure to call out in the doc for this that it's only running for C# at the moment.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -163,5 +207,36 @@ private void DeleteProducts(int first_id, int last_id)
...
         }
+
+        private async Task<string> AssertTriggerListenerError(string functionName)
+        {
+            string errorMessage = null;
+            var tcs = new TaskCompletionSource<bool>();
+
+            void OutputHandler(object sender, DataReceivedEventArgs e)
+            {
+                if (errorMessage == null && e.Data?.Contains("Failed to start SQL trigger listener") == true)
```
Charles-Gagnon, line 218: Why not take in the error message we're looking for an only complete when it sees that specific one? That seems simpler since you can just have this method be doing the check instead of each calling test.

Charles-Gagnon (changes requested):

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon, line 229: Is there a reason you're not just using `Task.Wait(STARTUP_ERROR_TIMEOUT)` and using the value of that to determine if it succeeded or not instead of doing this + the not null check at the end?

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: It is not much useful. I can remove it. If I remember correctly, the functions host stops and does not retry and exit if there is an error in early stage, for example, when requesting a trigger-binding from binding provider, etc.

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: We are asserting on the exact log message in the respective test cases, and this on assists to those assertions by applying additional check on the log message. If I replace it with `if (index >= 0)`, a future assertion inside the test case methods would fail, which will make it slightly difficult to track the origin of the error.

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: @Charles-Gagnon, do you mean in the summary comment of this class?

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: Can be done. I don't see any issue. It was not clear in the beginning that all the new test methods will end up having the same set of statements.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: Correct (or `remark`. Or honestly even just as a comment in the method, whatever you like best. I just want it called out somewhere so people looking at this down the line know it was done on purpose)

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: @Charles-Gagnon - I am not able to follow this comment. Can you elaborate a little? `Task.Wait` blocks the thread, so instead using `Task.WhenAny()`. Just good practice.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: I'm not sure what you're saying. There's going to be a bunch of log messages - most of which aren't containing the error message. In those cases we should just no-op. The point here is not to fail if we get a message that doesn't contain the expected error message (since that isn't going to be the first message anyways) but rather we only care about completing the task once it sees that message. 

If it sees any other message then it should just skip it and wait for the next message.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: Then just use Task.WaitAsync. The point is that this code currently seems a bit convoluted when essentially all you care about is "did our task complete in the expected time period". WaitAsync with a timeout is just a simpler and more straightforward way of doing that in my opinion.

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: The other lines will be filtered out by condition `e.Data?.Contains("Failed to start SQL trigger listener") == true`. This particular assert ensures that `InvalidOperationException` is thrown.

test/Integration/IntegrationTestBase.cs
JatinSanghvi: This line already existed earlier. I do not understand what is about this number. It's just like any other literal value like `true` or `Common.SupportedLanguages.CSharp`. It simply says wait until 60,000 milliseconds.

samples/samples-csharp/TriggerBindingSamples/ReservedPrimaryKeyColumnNamesTrigger.cs
JatinSanghvi: @Charles-Gagnon, can you explain why we should have different requirements for trigger tests compared to input and output bindings tests?

test/Integration/IntegrationTestBase.cs
Charles-Gagnon: https://en.wikipedia.org/wiki/Magic_number_(programming)

`Unique values with unexplained meaning or multiple occurrences which could (preferably) be replaced with named constants`

In this case is the meaning potentially clear enough? Probably. But in our code we want to be as clear as possible and adding a const with a useful name instead of hardcoded numbers makes it easier to understand and helps encourage this pattern across the code base.

Charles-Gagnon: (the fact that it existed like this earlier was likely just an oversight or code that we inherited. I'd appreciate it if you could just change that now since you're making changes here anyways)

JatinSanghvi: I can make the change as it is nonconsequential. I think the line is blurred between what numbers can and cannot be treated as magic. In my opinion, a magic number would be something like `56764` instead of `60000` that will tend to perplex the reader and make it difficult for him/her decide why that exact number. Here, these are ballpark numbers to give ample time for the host does not come up. I am not sure whether moving these values to local variable for one-time use will make the improve the readability. I find the change to be similar to replacing:
```cs
this.StartFunctionHost(nameof(ProductTrigger), Common.SupportedLanguages.CSharp, true);
```
with
```cs
string functionName = nameof(ProductTrigger);
SupportedLanguages language = Common.SupportedLanguages.CSharp;
bool useTestFolder = true;

this.StartFunctionHost(functionName, language, useTestFolder);
```

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: `Task.WaitAsync` was my first thought but it's not available in .NET Standard 2.0 ([doc link](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-6.0&viewFallbackFrom=netstandard-2.0)).

JatinSanghvi: Using non-async `Task.Wait` as suggested in first comment.

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: Hi @Charles-Gagnon, please check my last comment and see whether or not a code change will be required to resolve the conversation.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: That's unfortunate. Task.Wait should be fine for tests - if we see any issues then going back to what you had originally is fine too in this case.

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: Generally I don't find it useful to check for specific exception types unless it's a custom type (since InvalidOperationException could be thrown for a variety of reasons). But if you really think it's useful to have then it's fine to keep as is, although I would suggest appending to the comment about why we want to check for this specific exception type.

Charles-Gagnon (approved):

Charles-Gagnon:
There are some tests failing - please fix those before merging in.

lucyzhang929 (approved):

PR #297, opened 2022-08-18
Renew all leases through single statement

JatinSanghvi:
1. Make renew-leases operation a single SQL statement (_addresses https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r930429371_).
2. Simplify SQL variable names (no need to use hashes of column names).
3. Add comment for why trigger-binding provider should return null binding in particular case. (_addresses https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r933613139_)

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,7 +607,7 @@ The trigger binding utilizes SQL [change tracking](https://docs.microsoft.com/sq
 
     For more information, please refer to the documentation [here](https://docs.microsoft.com/sql/relational-databases/track-changes/enable-and-disable-change-tracking-sql-server#enable-change-tracking-for-a-table). The trigger needs to have read access on the table being monitored for changes as well as to the change tracking system tables. It also needs write access to an `az_func` schema within the database, where it will create additional worker tables to store the trigger states and leases. Each function trigger will thus have an associated change tracking table and worker table.
 
-    > **NOTE:** The worker table contains all columns corresponding to the primary key from the user table and three additional columns, particularly with names: `ChangeVersion`, `AttemptCount` and `LeaseExpirationTime`. If any of the primary key columns happen to have the same name, that will result in an error message describing the name-conflict. In such case, renaming the primary key column in the user table should resolve the error.
+    > **NOTE:** The worker table contains all columns corresponding to the primary key from the user table and three additional columns named `ChangeVersion`, `AttemptCount` and `LeastExpirationTime`. If any of the primary key columns happen to have the same name, that will result in an error message listing any conflicts. In this case, the listed primary key columns must be renamed for the trigger to work.
 
 #### Trigger Samples
 The trigger binding takes two [arguments](https://github.com/Azure/azure-functions-sql-extension/blob/main/src/TriggerBinding/SqlTriggerAttribute.cs)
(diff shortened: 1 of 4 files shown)


Charles-Gagnon:
It would be great if you could make separate issues for each of these things you're fixing and then closing them as you fix them, that'll make it much easier to see how many tasks are left and who's working on them.

README.md
```diff
@@ -607,6 +607,8 @@ The trigger binding utilizes SQL [change tracking](https://docs.microsoft.com/sq
 
     For more information, please refer to the documentation [here](https://docs.microsoft.com/sql/relational-databases/track-changes/enable-and-disable-change-tracking-sql-server#enable-change-tracking-for-a-table). The trigger needs to have read access on the table being monitored for changes as well as to the change tracking system tables. It also needs write access to an `az_func` schema within the database, where it will create additional worker tables to store the trigger states and leases. Each function trigger will thus have an associated change tracking table and worker table.
 
+    > **NOTE:** The worker table contains all columns corresponding to the primary key from the user table and three additional columns, particularly with names: `ChangeVersion`, `AttemptCount` and `LeaseExpirationTime`. If any of the primary key columns happen to have the same name, that will result in an error message describing the name-conflict. In such case, renaming the primary key column in the user table should resolve the error.
```
Charles-Gagnon, line 610:
```
The worker table contains all columns corresponding to the primary key from the user table and three additional columns named `ChangeVersion`, `AttemptCount` and `LeastExpirationTime`. If any of the primary key columns happen to have the same name, that will result in an error message listing any conflicts. In this case, the listed primary key columns must be renamed for the trigger to work.
```


```

File: src/TriggerBinding/SqlTriggerBindingProvider.cs [hunk 2]
```diff
@@ -57,6 +57,10 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
             ParameterInfo parameter = context.Parameter;
             SqlTriggerAttribute attribute = parameter.GetCustomAttribute<SqlTriggerAttribute>(inherit: false);
 
+            // During application startup, the WebJobs SDK calls 'TryCreateAsync' method of all registered trigger
+            // binding providers in sequence for each parameter in the user function. A provider that finds the
+            // parameter-attribute that it can handle returns the binding object. Rest of the providers are supposed to
+            // return null. This binding object later gets used for binding before every function invocation.
```
Charles-Gagnon, line 63: I'd suggest also updating it to directly return null instead of `default(ITriggerBinding)` since that's the value we actually want.

JatinSanghvi: This was done to prevent explicitly passing the type-parameter `ITriggerBinding` to call the generic `Task.FromResult` method. Calling this way keeps lines 66 and 88 similar.

Charles-Gagnon: Line 68? I'm not seeing anything in that line that looks similar. 

And I guess I just don't see the point in doing `default(ITriggerBinding)`. Your comment (and the functionality expected) is that null is returned if the provider can't handle the parameter. So we should just return null. Keep it simple and straightforward.

JatinSanghvi: It's line 84 in old revision and line 88 in new revision.

Charles-Gagnon (approved):

JatinSanghvi:
> It would be great if you could make separate issues for each of these things you're fixing and then closing them as you fix them, that'll make it much easier to see how many tasks are left and who's working on them.

I will get individual issues created starting of next week.

PR #298, opened 2022-08-18
Enable PR validation for 	`triggerbindings` branch

JatinSanghvi:
Required to discover any issues in `triggerbindings` branch early on rather than waiting until the branch is merged into `main`.

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -4,6 +4,7 @@ pr:
   branches:
     include:
     - main
+    - triggerbindings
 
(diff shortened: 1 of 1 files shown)


JatinSanghvi:
Need to re-run PR validation for #297 and #296 after this PR is merged.

Charles-Gagnon (approved):
Looks like there's a failing test, please follow up with fixing that.

and when you do fix it, please use descriptive messages for assertions. Right now it's just showing this which isn't helpful at all

```
Assert.Equal() Failure
Expected: 11
Actual:   21
```

PR #299, opened 2022-08-19
Enable strong-name signing

### LINKED ISSUE #295: Make `Microsoft.Azure.WebJobs.Extensions.Sql` a strong-named assembly
The build does not produce a strong-named assembly. This will cause build error for the applications that are themselves strong-named and require all their dependencies to be strong-named.

https://nuget.info/packages/Microsoft.Azure.WebJobs.Extensions.Sql/0.1.346-preview

image

JatinSanghvi:
Fixes #295

If public-keys are not specified for the friend assemblies, the compiler throws:

_error CS1726: Friend assembly reference 'DynamicProxyGenAssembly2' is invalid. Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations._

All friend assemblies (`Microsoft.Azure.WebJobs.Extensions.Sql.Tests`) of a strong name assembly (`Microsoft.Azure.WebJobs.Extensions.Sql`) should be signed, else we get below error during build. Hence, signing was also enabled for the test assembly.

_error CS0281: Friend access was granted by 'Microsoft.Azure.WebJobs.Extensions.Sql, Version=99.99.99.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35', but the public key of the output assembly ('') does not match that specified by the InternalsVisibleTo attribute in the granting assembly. [C:\Source\github.com\Azure\azure-functions-sql-extension\test\Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj]_


Charles-Gagnon:
I'm working on this and will have a PR out shortly.

PR #300, opened 2022-08-19
Always publish artifacts during build

Charles-Gagnon:
Even if tests fail - that can help debugging and just general be useful if someone is doing adhoc builds and doesn't actually care about the results of the tests when verifying the build artifacts.

### DIFF
--- a/builds/azure-pipelines/template-steps-publish.yml
+++ b/builds/azure-pipelines/template-steps-publish.yml
@@ -6,6 +6,7 @@ steps:
   displayName: 'Install .NET Core SDK 5.0.x for CodeSign'
   inputs:
     version: '5.0.x'
+  condition: succeededOrFailed()
 
 - task: EsrpCodeSigning@1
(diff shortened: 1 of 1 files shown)


builds/azure-pipelines/template-steps-publish.yml
```diff
@@ -6,6 +6,7 @@ steps:
   displayName: 'Install .NET Core SDK 5.0.x for CodeSign'
   inputs:
     version: '5.0.x'
+  condition: succeededOrFailed()
```
chlafreniere, line 9: Will this still try to codesign even if there was a compilation error?

Charles-Gagnon: Yes, but given that it's already failing then it doesn't really matter if this step also fails IMO. Trying to determine exactly what failed and only do these steps on certain types of failures isn't really worth it IMO.

chlafreniere (approved):

PR #310, opened 2022-08-23
Use constants for worker table column names

### LINKED ISSUE #308: Replace string literals with constants for worker table column names

I know you said you didn't use constants because of code readability, but I don't like the potential risk of regression here by having these values be separate. So I'd strongly prefer that we make constants for each of these - even if it does decrease the readability of the query slightly.

_Originally posted by @Charles-Gagnon in https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r933617162_

JatinSanghvi:
- Closes #308


JatinSanghvi:
@Charles-Gagnon, we can rename column names to say `_az_func_AttemptCount`, etc. to make conflicts extremely unlikely. This would be a breaking change (where worker table is already created with older schema) but hopefully that will be acceptable by the team dependent on it at this point.

Also, please let me know if you have a different take on how to replace string literals with constants.

There are columns in global state table too that can be converted to constants.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -36,6 +36,10 @@ internal sealed class SqlTableChangeMonitor<T> : IDisposable
         public const int LeaseIntervalInSeconds = 60;
         public const int LeaseRenewalIntervalInSeconds = 15;
 
+        private const string ChangeVersionColumnName = SqlTriggerConstants.WorkerTableChangeVersionColumnName;
```
Charles-Gagnon, line 39: Why are you making new constants here? You can just use the ones from SqlTriggerConstants directly below.

src/TriggerBinding/SqlTriggerConstants.cs
```diff
@@ -10,5 +10,9 @@ internal static class SqlTriggerConstants
         public const string GlobalStateTableName = "[" + SchemaName + "].[GlobalState]";
 
         public const string WorkerTableNameFormat = "[" + SchemaName + "].[Worker_{0}]";
+
+        public const string WorkerTableChangeVersionColumnName = "ChangeVersion";
```
Charles-Gagnon, line 14: I like the idea of prefixing these with `_az_func_` - since this isn't actually published yet there shouldn't be any issues with renaming columns.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -29,6 +29,10 @@ internal sealed class SqlTriggerListener<T> : IListener
         private const int ListenerStopping = 3;
         private const int ListenerStopped = 4;
 
+        private const string ChangeVersionColumnName = SqlTriggerConstants.WorkerTableChangeVersionColumnName;
```
Charles-Gagnon, line 32: Same here - just use the constants directly instead of making new ones here.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: This is done to keep the SQL queries look concise and legible.

src/TriggerBinding/SqlTriggerConstants.cs
JatinSanghvi: Also looking for opinion on `AzFuncAttemptCount`.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: It's fine to be a little verbose IMO - the difference of a dozen letters doesn't really matter.

src/TriggerBinding/SqlTriggerConstants.cs
Charles-Gagnon: What do you mean? I think prefixing all the internal column names with `_az_func_` is a good idea. (so it'd be `_az_func_attemptcount`)

JatinSanghvi:
@Charles-Gagnon, if we are going to change the table schema, I would also like to change the table name from `Worker_*` to `Leases_*` as the table is specific to user function, not to a worker (considering a worker is a function app instance).

lucyzhang929 (approved):

JatinSanghvi:
@lucyzhang929, @Charles-Gagnon, I will issue another PR to rename the worker table to leases table. Cosmos DB extension uses similar name as default name for the collection that is used to maintain leases ([link](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2-trigger?tabs=in-process%2Cfunctionsv2&pivots=programming-language-csharp#:~:text=(Optional)%20The%20name%20of%20the%20collection%20used%20to%20store%20leases.%20When%20not%20set%2C%20the%20value%20leases%20is%20used.)).

PR #311, opened 2022-08-24
Fix DateTime support AsyncEnumerable and AsyncCollector

Whivel:
# AsyncEnumerable
Forcing .ToString() on data read from SqlDataReader causes localization issue with some conversion such as datetime column.
Since SqlDataReader already converts sql columns to the CLR types, it is better to leave the task to the Json deserializator. 

# AsyncCollector
Json default serialization of DateTime properties could be not handled by dbms, so forcing to the standard ISO (without timezone) could be better.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -344,6 +344,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
         public class TableInformation
         {
+            private const string DATETIME_FORMAT = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fff";
+
             public IEnumerable<MemberInfo> PrimaryKeys { get; }
 
             /// <summary>
@@ -388,7 +390,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                 this.JsonSerializerSettings = new JsonSerializerSettings
                 {
-                    ContractResolver = new DynamicPOCOContractResolver(columns, comparer)
+                    ContractResolver = new DynamicPOCOContractResolver(columns, comparer),
+                    DateFormatString = DATETIME_FORMAT
                 };
             }
             public static bool GetCaseSensitivityFromCollation(string collation)
--- a/src/SqlBindingUtilities.cs
+++ b/src/SqlBindingUtilities.cs
@@ -136,9 +136,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         /// </summary>
         /// <param name="reader">Used to determine the columns of the table as well as the next SQL row to process</param>
         /// <returns>The built dictionary</returns>
-        public static IReadOnlyDictionary<string, string> BuildDictionaryFromSqlRow(SqlDataReader reader)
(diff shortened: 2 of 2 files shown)


Charles-Gagnon:
Can you give an example of what this is fixing? Preferably by creating an issue with details so we can properly track it and anyone else experiencing the same issue can easily find it.

lucyzhang929:
We should add some tests for this - maybe add a DateTime field to our Product object. Thoughts? @Charles-Gagnon @MaddyDev

MaddyDev:
> We should add some tests for this - maybe add a DateTime field to our Product object. Thoughts? @Charles-Gagnon @MaddyDev

Appreciate the thoughtfulness @lucyzhang929, we definitely should do that and also think about other edge case data types that can be affected in a similar manner.

Whivel:
> > We should add some tests for this - maybe add a DateTime field to our Product object. Thoughts? @Charles-Gagnon @MaddyDev
> 
> Appreciate the thoughtfulness @lucyzhang929, we definitely should do that and also think about other data types that can be affected in a similar manner.

This link could be useful [https://docs.microsoft.com/it-it/dotnet/framework/data/adonet/sql-server-data-type-mappings](https://docs.microsoft.com/it-it/dotnet/framework/data/adonet/sql-server-data-type-mappings)

Here my primary thoughts:

We already know the issue with **DateTime**, but we should check with **DateTimeOffset**

Using ToString() there should be issues also on **float and double** (CLR types) because of the decimal separator. Moreover if I am not wrong Sql Server has some difficulties to handle conversion from string to decimal/float/double if input string has also thousands separator.
Instead If ToString() is discarded (or any other form of indiscriminate string conversion), the Json serializator can convert the value directly to a number and JSON feature of Sql Server will read in the right way.

ToString() could have some issues also with **byte[]** (the ToString will convert the byte array into System.Byte[]).
In that case JSonSerializer will serialize using Base64 and JSON feature of Sql Server can read the value in the right way
That code returns 0x0102 as expected 
```
SELECT *
FROM OPENJSON('{"a": "AQI="}')
 WITH (
    a varbinary(100) '$.a'
  );
```

And now a list of some potential issues is:

| CLR Type | Sql Db Type | When? |
|-----------|--------------|----------|
| char[] | (n)(var)char |  R/W* |
| DateTime | datetime | R/W** |
| DateTime | smalldatetime | R/W** |
| DateTime | datetime2 | R/W** |
| DateTime | date | R/W** |
| DatetimeOffset | datetimeoffset | |
| TimeSpan | time | R/W** |
| Byte[] | varbinary | R° |
| Byte[] | Timestamp | |

And any valid of number types combinations

*By my tests char[] cannot be used to write (n)(var)char using JSON feature: OPENJSON will return "NULL" 
** Localization
° Default ToString() Implementation

Charles-Gagnon:
@Whivel I opened up a new PR with your fixes + tests, I really appreciate the detail you put into the description and for getting the fixes figured out in the first place! Your contribution means a lot to us and we're glad to see community support for SQL bindings!

I'm going to close this then, but if you have any other comments or suggestions please leave them in that PR!

https://github.com/Azure/azure-functions-sql-extension/pull/329

PR #320, opened 2022-08-27
connection info telemetry data from input bindings

MaddyDev:
Sql version isn't being captured for all the input bindings since we aren capturing the event prior to getting that info. Added  a new event BuildItemFromAttributeAsync.

### DIFF
--- a/src/SqlConverters.cs
+++ b/src/SqlConverters.cs
@@ -158,12 +158,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 using (var adapter = new SqlDataAdapter())
                 using (SqlCommand command = SqlBindingUtilities.BuildCommand(attribute, connection))
                 {
+                    Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps();
+                    TelemetryInstance.TrackConvert(type, props);
                     adapter.SelectCommand = command;
                     await connection.OpenAsync();
-                    Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps();
                     var dataTable = new DataTable();
                     adapter.Fill(dataTable);
-                    TelemetryInstance.TrackConvert(type, props);
                     this._logger.LogInformation($"{dataTable.Rows.Count} row(s) queried from database: {connection.Database} using Command: {command.CommandText}");
                     return JsonConvert.SerializeObject(dataTable);
                 }
--- a/src/SqlConverters.cs
+++ b/src/SqlConverters.cs
@@ -149,6 +149,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <param name="attribute">
             /// The binding attribute that contains the name of the connection string app setting and query.
             /// </param>
+            /// <param name="type">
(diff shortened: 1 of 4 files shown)


src/SqlConverters.cs
```diff
@@ -162,8 +162,10 @@ public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attri
                 {
                     adapter.SelectCommand = command;
                     await connection.OpenAsync();
+                    Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps();
                     var dataTable = new DataTable();
                     adapter.Fill(dataTable);
+                    TelemetryInstance.TrackEvent(TelemetryEventName.BuildItemFromAttributeAsync, props);
```
lucyzhang929, line 168: Maybe we can pass the ConvertType to BuildItemFromAttributeAsync and use `TelemetryInstance.TrackConvert(type, props)` here instead of creating a new event?

Charles-Gagnon: +1 - we already have an event for input bindings so we shouldn't change that.

Charles-Gagnon: @MaddyDev That's what Lucy is saying. Add a parameter for the type of convert being done and then move the TraceConvert call into this method instead

src/SqlConverters.cs
```diff
@@ -152,14 +150,19 @@ public async Task<IEnumerable<T>> ConvertAsync(SqlAttribute attribute, Cancellat
             /// The binding attribute that contains the name of the connection string app setting and query.
             /// </param>
             /// <returns></returns>
-            public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attribute)
+            public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attribute, ConvertType? type = null)
```
Charles-Gagnon, line 153: Why is this optional? Looking at it the only places it's called all call TrackConvert so we should be sending the event with connection props for all of those, right?

Charles-Gagnon, line 153: Update xml doc

Charles-Gagnon (changes requested):

src/SqlConverters.cs
MaddyDev: I made it optional for tests.

Charles-Gagnon: That's generally not a pattern we should be using - instead you should be providing the appropriate mocks/passing the appropriate values as needed to simulate actual behavior as closely as possible.

MaddyDev: I know we have to simulate actual behavior but this is to collect telemetry, do we want data from tests as well?

Charles-Gagnon: We will not be sending the actual event. Instead generally we would mock out the telemetry service so that it just no-ops any events being sent. (I don't know if this is actually being done currently, you may have to do it).

But the fact that this is used for telemetry is just an implementation detail, and not guaranteed to be the case in the future if we decide to use this value somewhere else so we shouldn't be making assumptions for that in the tests unless there's some reason we can't/shouldn't do that right now.

PR #328, opened 2022-08-30
Add more debug logging

lucyzhang929:
Adding some more debug logging that can also be used to analyze perf for input and output bindings.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -514,12 +514,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                 // Get case sensitivity from database collation (default to false if any exception occurs)
                 bool caseSensitive = false;
-                logger.LogDebugWithThreadId("BEGIN GetCaseSensitivity");
                 var tableInfoSw = Stopwatch.StartNew();
                 var caseSensitiveSw = Stopwatch.StartNew();
                 try
                 {
                     string getDatabaseCollationQuery = GetDatabaseCollationQuery(sqlConnection);
+                    logger.LogDebugWithThreadId($"BEGIN GetCaseSensitivity Query=\"{getDatabaseCollationQuery}\"");
                     var cmdCollation = new SqlCommand(getDatabaseCollationQuery, sqlConnection);
                     using (SqlDataReader rdr = await cmdCollation.ExecuteReaderAsync())
                     {
@@ -529,7 +529,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                         }
                         caseSensitiveSw.Stop();
                         TelemetryInstance.TrackDuration(TelemetryEventName.GetCaseSensitivity, caseSensitiveSw.ElapsedMilliseconds, sqlConnProps);
-                        logger.LogDebugWithThreadId($"END GetCaseSensitivity Duration={caseSensitiveSw.ElapsedMilliseconds}ms Query=\"{getDatabaseCollationQuery}\"");
+                        logger.LogDebugWithThreadId($"END GetCaseSensitivity Duration={caseSensitiveSw.ElapsedMilliseconds}ms");
                     }
                 }
                 catch (Exception ex)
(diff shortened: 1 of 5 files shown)


src/Utils.cs
```diff
@@ -93,5 +94,10 @@ public static void LowercasePropertyNames(this JObject obj)
                 property.Replace(new JProperty(property.Name.ToLowerInvariant(), property.Value));
             }
         }
+
+        public static void LogDebugWithThreadId(this ILogger logger, string message, params object[] args)
```
lucyzhang929, line 98: @Charles-Gagnon I included the thread Id but not sure if it is actually useful for input output bindings - it was mainly for row locks in trigger code right?

Charles-Gagnon: Yeah it was, but in general it's not a bad thing to add - especially as we are looking into supporting scalable apps.

src/SqlAsyncCollector.cs
```diff
@@ -189,6 +190,7 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                 }
 
                 TelemetryInstance.TrackEvent(TelemetryEventName.UpsertStart, props);
+                this._logger.LogDebugWithThreadId("BEGIN upsert rows transaction");
```
Charles-Gagnon, line 193: nit: I'd suggest using a single word after BEGIN/END - the idea there was to make it easy to match the two events by searching for that (which is easier to do with just a single word you can double click to select)

src/SqlAsyncCollector.cs
```diff
@@ -216,7 +218,8 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                     { TelemetryMeasureName.CommandDurationMs, commandSw.ElapsedMilliseconds }
                 };
                     TelemetryInstance.TrackEvent(TelemetryEventName.UpsertEnd, props, measures);
-                    this._logger.LogInformation($"Upserted {rows.Count()} row(s) into database: {connection.Database} and table: {fullTableName}.");
+                    this._logger.LogDebugWithThreadId($"END upsert rows transaction Duration={transactionSw.ElapsedMilliseconds}ms Upserted {rows.Count()} row(s) into database: {connection.Database} and table: {fullTableName}.");
```
Charles-Gagnon, line 221: I'd suggest saving off the duration values so that they match what we're sending in telemetry and aren't affected by any calls between the end of the command and us logging this

src/SqlConverters.cs
```diff
@@ -98,10 +106,14 @@ public SqlGenericsConverter(IConfiguration configuration, ILogger logger)
             public async Task<IEnumerable<T>> ConvertAsync(SqlAttribute attribute, CancellationToken cancellationToken)
             {
                 TelemetryInstance.TrackConvert(ConvertType.IEnumerable);
+                this._logger.LogDebugWithThreadId("BEGIN ConvertAsync (IEnumerabe)");
```
Charles-Gagnon, line 109: IEnumerable

src/SqlAsyncCollector.cs
```diff
@@ -504,6 +507,7 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
             {
                 Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps();
                 TelemetryInstance.TrackEvent(TelemetryEventName.GetTableInfoStart, sqlConnProps);
+                logger.LogDebugWithThreadId("BEGIN RetrieveTableInformationAsync");
```
Charles-Gagnon, line 510: We should log all queries too so we can see exactly what is being executed in case there's some edge case we aren't handling correctly

src/SqlAsyncCollector.cs
```diff
@@ -521,6 +529,7 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                         }
                         caseSensitiveSw.Stop();
                         TelemetryInstance.TrackDuration(TelemetryEventName.GetCaseSensitivity, caseSensitiveSw.ElapsedMilliseconds, sqlConnProps);
+                        logger.LogDebugWithThreadId($"END GetCaseSensitivity Duration={caseSensitiveSw.ElapsedMilliseconds}ms Query=\"{getDatabaseCollationQuery}\"");
```
Charles-Gagnon, line 532: You should log the queries at the start - since if they fail they'll never hit these lines

Charles-Gagnon (changes requested):

Charles-Gagnon (approved):

Issue #316, opened 2022-08-25
Localization/serialization Issues on IAsyncEnumerable and IAsyncCollector


Whivel:
# Issue Reproduction Step
**DB**
```
CREATE TABLE [dbo].[LocalizationIssue] (
    [ID] int IDENTITY (1,1) PRIMARY KEY,
    [LastExecutionDate] datetime NULL
)
GO
INSERT INTO [dbo].[LocalizationIssue] ([LastExecutionDate]) VALUES (NULL),(GETUTCDATE())
GO
```
**Azure Function IAsyncEnumerable**
```
public record class ItemDto(int ID, DateTime? LastExecutionDate);
        [FunctionName("Function1")]
        public static async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
            [Sql("SELECT * FROM [dbo].[LocalizationIssue]", ConnectionStringSetting = "SqlServer")] IAsyncEnumerable<ItemDto> items,
            ILogger log)
        {
            log.LogInformation("C# HTTP trigger function processed a request.");
            await foreach (var item in items)
            {
                log.LogInformation(JsonSerializer.Serialize(item));
            }

            return new OkObjectResult("");
        }
```

**Azure Function IAsyncCollector**
```
public record class ItemDto(int ID, DateTime? LastExecutionDate);
        [FunctionName("Function1")]
        public static async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
            [Sql("[dbo].[LocalizationIssue]",CommandType = System.Data.CommandType.TableDirect, ConnectionStringSetting = "SqlServer")] IAsyncCollector<ItemDto> items,
            ILogger log)
        {
            log.LogInformation("C# HTTP trigger function processed a request.");
            await items.AddAsync(new ItemDto(999, DateTime.UtcNow));
            await items.FlushAsync();

            return new OkObjectResult("");
        }
```
CultureInfo.CurrentCulture: **"it-IT"**
CultureInfo.CurrentUICulture: "en-US"

# Result IAsyncEnumerable
```
[2022-08-25T07:31:54.098Z] Executed 'Function1' (Failed, Id=fc64982d-fa38-4427-8778-e7c8d1c9a2d7, Duration=1317ms)
[...]

PR #329, opened 2022-08-30
Fix localization/serialization issues

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/316

Original PR with fixes : https://github.com/Azure/azure-functions-sql-extension/pull/311

The issue has a good description of the issues, the only thing I found was that for the output binding it has nothing to do with the collector so I just use a normal out parameter to keep things simpler. 

The tests are meant to have other types added to them later, keeping them confined to just datetime types for now to keep this smaller. And we may want to make some other tests for different languages or different binding types (like IEnumerable) just to verify all the scenarios. But that can be done as a follow up task down the line - this fixes the immediate issues brought up by the original issue.


test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
```diff
@@ -35,10 +35,10 @@
 
   <Target Name="CopyTestSqlFiles" AfterTargets="Build">
     <ItemGroup>
-      <_CSharpTestSqlFiles Include="Integration\test-csharp\Database\**\*.*" />
```
Charles-Gagnon, line 38: This folder didn't even exist - guessing it was left over from some other work and never cleaned up.

lucyzhang929: This is from https://github.com/Azure/azure-functions-sql-extension/pull/317/files#diff-1ef1922c8368ea733b52db01d803d4da431c6a250fe37ca710169142665df11a. The Integration\test-csharp\Database\ folder is in the trigger branch although we should probably move those sql files into test\Database too. 

I think this copy is not needed since we already copy here https://github.com/Azure/azure-functions-sql-extension/blob/a1a045e25ded463f8466dab994051e646ffef5d0/samples/samples-csharp/Microsoft.Azure.WebJobs.Extensions.Sql.Samples.csproj#L19-L21

test/Integration/test-js/GetProductsColumnTypesDifferentCulture/index.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This function verifies that serializing an item with various data types
+ * works when the language is set to a non-enUS language.
+ */
+module.exports = async function (context, req, products) {
+    context.log(JSON.stringify(products));
```
lucyzhang929, line 6: Do we need to set the culture before calling stringify?

lucyzhang929 (approved):

test/Integration/test-js/GetProductsColumnTypesDifferentCulture/index.js
Charles-Gagnon: Good callout - so I'm not actually aware of any way to set a culture like you can for .NET apps. JSON.stringify has its own default method of serialization which you can override though.

So instead what I'm doing is splitting this into two tests - one that just tests serialization in general with the default language and another that specifically tests with another culture set for languages that support that (i.e. just C# for now).

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
Charles-Gagnon: Ah, thanks for digging into that!

We do need both copies - since they're from separate projects and the paths are relative to that project. But we can remove the custom target and instead just use the None items like the samples project does which simplifies it a bit.

MaddyDev (approved):

MaddyDev:
Looks like there is a test failure, looks great otherwise.

PR #331, opened 2022-09-01
Run PR validation on release branches and rename nuget to *-lapreview.nupkg

Charles-Gagnon:
Currently the packages generated from the LA preview branch will have numbers that are very similar to the "official" nuget.org packages which could be dangerous (if the numbers happened to line up) and confusing (hard to tell where a package came from). Changing it to -lapreview will help avoid these problems. 

Doing directly to release branch since this isn't something we want in main ever - so leaving it out of the triggerbindings branch to keep things clean for eventual merging there.

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -5,6 +5,7 @@ pr:
     include:
     - main
     - triggerbindings
+    - release/*
 
 variables:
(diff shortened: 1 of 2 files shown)


dzsquared (approved):

Charles-Gagnon:
Merging in now to get release package out asap

PR #337, opened 2022-09-07
sql version telemetry data from input bindings

MaddyDev:
Sql version isn't being captured for all the input bindings since we are capturing the event prior to getting that info. Moved it to BuildItemFromAttributeAsync method to capture that info from connection.

I looked into creating a ITelemetryService and how we can pass the mock for tests, but we're mocking the calls where we capture the telemetry in tests already and is not needed for the current code. Will definitely work on it in a separate PR for future releases.


lucyzhang929 (approved):

src/SqlConverters.cs
```diff
@@ -188,10 +191,12 @@ public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attri
 
             IAsyncEnumerable<T> IConverter<SqlAttribute, IAsyncEnumerable<T>>.Convert(SqlAttribute attribute)
             {
-                TelemetryInstance.TrackConvert(ConvertType.IAsyncEnumerable);
                 try
                 {
-                    return new SqlAsyncEnumerable<T>(SqlBindingUtilities.BuildConnection(attribute.ConnectionStringSetting, this._configuration), attribute);
+                    SqlConnection conn = SqlBindingUtilities.BuildConnection(attribute.ConnectionStringSetting, this._configuration);
+                    Dictionary<TelemetryPropertyName, string> props = conn.AsConnectionProps();
```
Charles-Gagnon, line 197: You're trying to access the properties here before the connection is opened which is likely why the tests are failing now.

Charles-Gagnon (changes requested):

src/SqlConverters.cs
Charles-Gagnon: Actually, for that second part I'll make a new issue and we can follow up with that later. No reason to clutter up this PR even more.

MaddyDev: So I can't really collect the server version unless the connection is open?

Charles-Gagnon: https://github.com/Azure/azure-functions-sql-extension/issues/338

Charles-Gagnon: Given that you need to talk to the server to know what version it is, yes that would be extremely difficult to do without an open connection 😄

MaddyDev: But am collecting that info prior to opening the connection in BuildItemFromAttributeAsync, how doeds that work?

src/SqlAsyncEnumerable.cs
```diff
@@ -106,6 +107,8 @@ private async Task<bool> GetNextRowAsync()
                     using (SqlCommand command = SqlBindingUtilities.BuildCommand(this._attribute, this._connection))
                     {
                         await command.Connection.OpenAsync();
+                        Dictionary<TelemetryPropertyName, string> props = command.Connection.AsConnectionProps();
```
Charles-Gagnon, line 110: My concern here is that since we're not sending this event until GetNextRowAsync is called it's going to behave differently than all the other ones (which immediately send the event when the convert is called). 

My assumption based on this code is that if someone made a function which used IAsyncEnumerable but then never actually called MoveNextAsync that we'd never fire this event and thus get a lower count of people using this. That doesn't seem like a super reasonable case that I can think of, but in general we still should avoid knowingly having gaps in our telemetry.

Instead it seems like we should discuss opening the connection as soon as the AsyncEnumerable is created, or passing in an already open connection (but still letting the dispose close it). That way we could send the event as soon as convert is called and still lazily get the rows from the DB. 

@lucyzhang929 @chlafreniere Thoughts here?

lucyzhang929: I think opening the connection when SqlAsyncEnumerable is created makes sense here.

Charles-Gagnon: My suggestion for that would be : 

1. Open the connection in the constructor
2. Making the connection a public readonly property
3. In the code that creates the AsyncEnumerable (`IConverter<SqlAttribute, IAsyncEnumerable<T>>.Convert`) create the enumerable, use the connection from it to send the event and then return that enumerable

This keeps the event logic all in the same place which is easier to work with.

src/SqlAsyncEnumerable.cs
```diff
@@ -60,6 +61,24 @@ public SqlAsyncEnumerator(SqlConnection connection, SqlAttribute attribute)
...
+            /// <summary>
+            /// Gets the data from server into the reader ready for use.
+            /// </summary>
+            private async void GetDataAsync()
+            {
+                if (this._reader == null)
+                {
+                    using (SqlCommand command = SqlBindingUtilities.BuildCommand(this._attribute, this._connection))
+                    {
+                        await command.Connection.OpenAsync();
```
Charles-Gagnon, line 76: 1. We don't need to create the command here. Just open the connection. The command execution can still stay where it was
2. I'd prefer the TrackConvert be done in SqlConverters, see my other comment
3. You can just use Connection.Open() to open the connection - we don't really need to use the async version here for that.

Charles-Gagnon: The idea with not executing the command is that we still want to be as lazy as possible to avoid overhead when we can. Opening the connection is a separate step from executing the command so we can just delay that until we actually need to fetch the rows - that still allows us to get server metadata like the version from the connection itself before we actually execute the command.

src/SqlAsyncEnumerable.cs
```diff
@@ -105,7 +108,10 @@ private async Task<bool> GetNextRowAsync()
                 {
                     using (SqlCommand command = SqlBindingUtilities.BuildCommand(this._attribute, this._connection))
                     {
-                        await command.Connection.OpenAsync();
+                        if (this._connection.State != System.Data.ConnectionState.Open)
+                        {
+                            await this._connection.OpenAsync();
```
Charles-Gagnon, line 113: Is there a reason you're doing this here? The connection shouldn't be closed so I wouldn't expect this to be necessary.

src/SqlAsyncEnumerable.cs
```diff
@@ -26,8 +25,12 @@ internal class SqlAsyncEnumerable<T> : IAsyncEnumerable<T>
         /// </exception>
         public SqlAsyncEnumerable(SqlConnection connection, SqlAttribute attribute)
         {
-            this._connection = connection ?? throw new ArgumentNullException(nameof(connection));
+            this.Connection = connection ?? throw new ArgumentNullException(nameof(connection));
             this._attribute = attribute ?? throw new ArgumentNullException(nameof(attribute));
+            if (!string.IsNullOrEmpty(this.Connection.ConnectionString))
```
Charles-Gagnon, line 30: Is there a reason you're adding this check here? If the connection string is empty then nothing will work - so having it fail the Open call seems fine rather than just silently skipping the open

MaddyDev: TestNullCurrentValueEnumerator calls the constructor without any connection string and it fails, so I added that check.

src/SqlAsyncEnumerable.cs
Charles-Gagnon: If we did want to add that we should add it every time we read a value - since at that point we're saying that we expect the connection could be closed at any point. But if we did that let's do it as a separate task and just keep the behavior as it is now, there's no reasonable reason I can think of for why the connection would have been closed between the creation of the enumerable and querying the rows.

src/SqlAsyncEnumerable.cs
Charles-Gagnon: I would suggest refactoring that test to be like TestNullArgumentsSqlAsyncEnumerableConstructor and assert that it throws. Without a connection we can't do anything so it seems reasonable to me to throw immediately in that case instead of just skipping the open and later on failing.

test/Unit/SqlInputBindingTests.cs
```diff
@@ -60,17 +61,14 @@ public void TestNullCommand()
...
         public void TestNullArgumentsSqlAsyncEnumerableConstructor()
         {
-
             Assert.Throws<ArgumentNullException>(() => new SqlAsyncEnumerable<string>(connection, null));
             Assert.Throws<ArgumentNullException>(() => new SqlAsyncEnumerable<string>(null, new SqlAttribute("")));
         }
 
         [Fact]
-        public void TestNullCurrentValueEnumerator()
+        public void TestInvalidOperationSqlAsyncEnumerableConstructor()
```
Charles-Gagnon, line 69: I'd suggest adding an xml doc comment explaining that this is expected to throw when an invalid/empty connection string setting is given

Charles-Gagnon (approved):

test/Unit/SqlInputBindingTests.cs
```diff
@@ -60,17 +61,16 @@ public void TestNullCommand()
         [Fact]
         public void TestNullArgumentsSqlAsyncEnumerableConstructor()
         {
-
             Assert.Throws<ArgumentNullException>(() => new SqlAsyncEnumerable<string>(connection, null));
             Assert.Throws<ArgumentNullException>(() => new SqlAsyncEnumerable<string>(null, new SqlAttribute("")));
         }
 
+        /// SqlAsyncEnumerable should throw InvalidOperationExcepion when invoked with an invalid connection
```
Charles-Gagnon, line 68: Use actual XML doc (with <summary> tags) here. And I would probably specify that this is testing with an empty connection string setting (since invalid could also mean non-empty but also non-existent setting)

MaddyDev: No other places require the connection open check, am going to get this in and we can prep for a release.

src/SqlAsyncEnumerable.cs
```diff
@@ -105,7 +105,10 @@ private async Task<bool> GetNextRowAsync()
                 {
                     using (SqlCommand command = SqlBindingUtilities.BuildCommand(this._attribute, this._connection))
                     {
-                        await command.Connection.OpenAsync();
+                        if (this._connection.State != System.Data.ConnectionState.Open)
+                        {
+                            await this._connection.OpenAsync();
```
Charles-Gagnon, line 110: I still don't understand why this is needed. Why would the connection be closed between the constructor call and this?

Charles-Gagnon (changes requested):

src/SqlAsyncEnumerable.cs
MaddyDev: I debugged this and am not sure why but the MoveNextAsync is being invoked one more time after DisposeAsync() is called. So the connection is being closed in DisposeAsync and we're trying to load the reader again so the error. I've added connection check and to return false if the connection is closed.

src/SqlAsyncEnumerable.cs
```diff
@@ -77,8 +77,14 @@ public SqlAsyncEnumerator(SqlConnection connection, SqlAttribute attribute)
             public ValueTask DisposeAsync()
             {
                 // Doesn't seem like there's an async version of closing the reader/connection
-                this._reader.Close();
-                this._connection.Close();
+                if (this._reader != null)
+                {
+                    this._reader.Close();
+                }
+                if (this._connection != null && this._connection.State == System.Data.ConnectionState.Open)
```
Charles-Gagnon, line 84: Why are you adding these checks? We create the connection in the constructor so that should never be null, and closing an already-closed connection just no-ops so we should be fine to just leave it like it is currently.

src/SqlAsyncEnumerable.cs
```diff
@@ -101,23 +107,22 @@ public ValueTask<bool> MoveNextAsync()
             /// </returns>
             private async Task<bool> GetNextRowAsync()
             {
-                if (this._reader == null)
+                if (this._connection.State != System.Data.ConnectionState.Closed)
```
Charles-Gagnon, line 110: Could you debug and add a breakpoint in the dispose call to figure out where that's coming from? Because the fact that it's disposing the enumerable while something (the function) is still using it is worrisome to me.

MaddyDev: image

src/SqlAsyncEnumerable.cs
```diff
@@ -77,8 +77,14 @@ public SqlAsyncEnumerator(SqlConnection connection, SqlAttribute attribute)
             public ValueTask DisposeAsync()
             {
                 // Doesn't seem like there's an async version of closing the reader/connection
-                this._reader.Close();
-                this._connection.Close();
+                if (this._reader != null)
+                {
+                    this._reader.Close();
```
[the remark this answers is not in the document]
MaddyDev: I added that at first and the compiler showed me some error with the red underline and I thought it's specific to typescript.

src/SqlAsyncEnumerable.cs
MaddyDev: Forgot to remove this (was debugging and it's throwing null reference, so added all these)

src/SqlAsyncEnumerable.cs
MaddyDev: GetProductsAsyncEnumerable also invokes the same code but it doesn't call  MoveNextAsync after the dispose, probably depends on how you invoke the enumerator.

Charles-Gagnon: Did you run "Show External Code"?

https://stackoverflow.com/questions/11629340/what-does-external-code-in-the-call-stack-mean

(I usually prefer to use VS for this reason since it's a lot better about this stuff)

MaddyDev: image

src/SqlAsyncEnumerable.cs
```diff
@@ -72,13 +74,15 @@ public SqlAsyncEnumerator(SqlConnection connection, SqlAttribute attribute)
 
             /// <summary>
             /// Closes the SQL connection and resources associated with reading the results of the query
+            /// Set _isDisposed to true so any further MoveNextAsync calls will not attempt to read from a disposed reader.
```
Charles-Gagnon, line 77: I'd put a comment down in GetNextRowAsync instead and mention that you were seeing it called after DisposeAsync was called, it'd be easy to miss this comment currently.

Charles-Gagnon (approved):

src/SqlAsyncEnumerable.cs
```diff
@@ -101,23 +100,22 @@ public ValueTask<bool> MoveNextAsync()
             /// </returns>
             private async Task<bool> GetNextRowAsync()
             {
-                if (this._reader == null)
+                if (this._connection.State != System.Data.ConnectionState.Closed)
```
Charles-Gagnon, line 103: Assuming the tests pass please create an issue for investigating this and then add a comment explaining why we're doing this with a link to the issue.

MaddyDev:
Reverted the changes to check on connection state and created the following to track this https://github.com/Azure/azure-functions-sql-extension/issues/350

Charles-Gagnon (approved):

src/SqlAsyncEnumerable.cs
MaddyDev: I ran locally and they passed (fingers crossed), will add the comment, thanks

PR #340, opened 2022-09-09
Add retry when attempting to release leases

### LINKED ISSUE #282: Add retry if releasing the lease fails

A retry here would be preferable - if it fails after 3ish retries then the default behavior of the lease expiring will have to do.

_Originally posted by @Charles-Gagnon in https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r929157776_

AmeyaRele:
Closes #282


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -494,67 +494,69 @@ private async Task ReleaseLeasesAsync(SqlConnection connection, CancellationToke
...
-                            releaseLeasesDurationMs = commandSw.ElapsedMilliseconds;
-                            this._logger.LogDebugWithThreadId($"END ReleaseLeases Duration={releaseLeasesDurationMs}ms");
-                        }
+                    var transactionSw = Stopwatch.StartNew();
+                    long releaseLeasesDurationMs = 0L, updateLastSyncVersionDurationMs = 0L;
 
-                        // Update the global state table if we have processed all changes with ChangeVersion <= newLastSyncVersion,
-                        // and clean up the leases table to remove all rows with ChangeVersion <= newLastSyncVersion.
-                        using (SqlCommand updateTablesPostInvocationCommand = this.BuildUpdateTablesPostInvocation(connection, transaction, newLastSyncVersion))
+                    using (SqlTransaction transaction = connection.BeginTransaction(System.Data.IsolationLevel.RepeatableRead))
```
Charles-Gagnon, line 502: One thing I've been thinking about that we need to be careful of is our use of transactions across the same connection. A couple things I've found looking through the docs and playing around with it myself : 

1. SqlConnection doesn't like it when multiple transactions are opened for the same connection. Although oddly this doesn't seem to matter with multiple threads, I was able to open multiple transactions as long as that was done on another thread...not sure if that's intended or not
2. If you try to execute a command on a connection with an open transaction it'll throw

Hopefully this shouldn't be a huge issue since we're opening a new connection for every worker that's processing changes, but we need to make sure that each worker isn't going to be doing anything in parallel that could cause issues.

Charles-Gagnon (approved):

lucyzhang929 (approved):

MaddyDev (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -494,67 +494,69 @@ private async Task ReleaseLeasesAsync(SqlConnection connection, CancellationToke
...
 
-                        try
-                        {
-                            transaction.Rollback();
+                            TelemetryInstance.TrackEvent(TelemetryEventName.ReleaseLeasesEnd, this._telemetryProps, measures);
+                            // Don't want to loop if no exception occurs.
+                            break;
                         }
-                        catch (Exception ex2)
+                        catch (Exception ex) when (retryCount < MaxRetryReleaseLeases)
```
JatinSanghvi, line 539: IMO, we should log the error and attempt to rollback the transaction event if this is not the last retry. The for-block can be placed outside the outer-most try-block.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -494,67 +494,69 @@ private async Task ReleaseLeasesAsync(SqlConnection connection, CancellationToke
...
+                            {
+                                [TelemetryMeasureName.ReleaseLeasesDurationMs] = releaseLeasesDurationMs,
+                                [TelemetryMeasureName.UpdateLastSyncVersionDurationMs] = updateLastSyncVersionDurationMs,
+                            };
 
-                        try
-                        {
-                            transaction.Rollback();
+                            TelemetryInstance.TrackEvent(TelemetryEventName.ReleaseLeasesEnd, this._telemetryProps, measures);
+                            // Don't want to loop if no exception occurs.
```
JatinSanghvi, line 536: nit: Try to avoid putting C# comments in line immediately following the line with C# expression (just braces do not count as code). Add a blank line above the comments in such cases. Arguably, we can replace `break` with a variable named `retrySucceeded` and check for its value in the for-statement.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Thank for the insights. Here, we are having two loops running in parallel, changed-consumption and lease-renewal, and both use their own connections so no issue there. We will probably be adding a new loop for scale monitoring. I will ensure another connection is created for that one. As it seems, with every new change, we should ensure that the code is correctly committing (or rolling back) the transaction before beginning a new one.

Charles-Gagnon: FYI - got a response on the issue I submitted

https://github.com/dotnet/SqlClient/issues/1758#issuecomment-1243072330

So we need to be careful that we're never opening a connection and then accessing it in another thread. I looked through everything and it seems like we're doing that currently everywhere so we should be safe but if you could also take a look through and make sure I'm not missing anything I'd appreciate it.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: We should definitely be trying to rollback on the last exception. The error and exception event being logged though I think I would prefer to keep separate as they are - we want to make it clear that it failed all the retry attempts and is now going to move on (and from telemetry purposes I want a separate event in case we change the number of retries later - so that we can easily pick out the number of times that a user hit the worst-case scenario where the retries all failed).

So what I'd suggest is to : 

1. Remove the when from this catch
2. Have an if in here that changes what is logged/tracked based on whether it's the last attempt or not
3. Remove the outer catch 
4. You should be able to just remove the finally as well and have it run after the loop - since we'll be catching and handling all the exceptions

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -491,8 +491,9 @@ private async Task ReleaseLeasesAsync(SqlConnection connection, CancellationToke
             await this._rowsLock.WaitAsync(token);
             this._logger.LogDebugWithThreadId("END WaitRowsLock - ReleaseLeases");
             long newLastSyncVersion = this.RecomputeLastSyncVersion();
+            bool retrySucceeded = false;
 
-            try
+            for (int retryCount = 1; retryCount <= MaxRetryReleaseLeases && retrySucceeded == false; retryCount++)
```
JatinSanghvi, line 496: Replace `retrySucceeded == false` with `!retrySucceeded`. In general, the pattern where Boolean value is compared against `true` or `false` is discouraged in all programming languages.

AmeyaRele: Thanks, changed it.

Charles-Gagnon (approved):

PR #343, opened 2022-09-10
Add test for multiple triggered functions

JatinSanghvi:
The test ensures correct functionality with multiple user functions tracking the same table.


Charles-Gagnon (approved):
I'd still prefer that we use a TaskCompletionSource instead of just a straight time delay while waiting to validate that the changes were received (same reasoning as the fix for the error checking). But if you don't want to do that now then I'll just open an issue and come fix it later.

PR #344, opened 2022-09-10
Add test for multiple function hosts

JatinSanghvi:
The test ensures correct functionality with user functions running across multiple functions host processes.


test/Integration/IntegrationTestBase.cs
```diff
@@ -343,14 +370,17 @@ public void Dispose()
...
-                this.FunctionHost?.Kill();
-                this.FunctionHost?.Dispose();
-            }
-            catch (Exception e2)
+            foreach (Process functionHost in this.FunctionHostList)
             {
-                this.TestOutput.WriteLine($"Failed to stop function host, Error: {e2.Message}");
+                try
+                {
+                    functionHost?.Kill();
```
Charles-Gagnon, line 377: This shouldn't need to use the null operator anymore

test/Integration/IntegrationTestBase.cs
```diff
@@ -259,11 +280,17 @@ private static string GetFunctionsCoreToolsPath()
...
+
+        private DataReceivedEventHandler GetTestOutputHandler(int processId)
         {
-            if (e != null && !string.IsNullOrEmpty(e.Data))
+            return TestOutputHandler;
+
+            void TestOutputHandler(object sender, DataReceivedEventArgs e)
             {
-                this.TestOutput.WriteLine(e.Data);
+                if (e != null && !string.IsNullOrEmpty(e.Data))
```
Charles-Gagnon, line 290: `if (!string.IsNullOrEmpty(e?.Data))`

test/Integration/IntegrationTestBase.cs
```diff
@@ -1,29 +1,36 @@
...
 using Xunit.Abstractions;
-using Microsoft.Azure.WebJobs.Extensions.Sql.Samples.Common;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
 {
     public class IntegrationTestBase : IDisposable
     {
         /// <summary>
-        /// Host process for Azure Function CLI
+        /// Host process for Azure Function CLI. Useful when only one host process is involved.
```
Charles-Gagnon, line 26: I'd even just specify that it gets the first - that should be clear enough.

`The first Function Host process that was started. Null if no process has been started yet.`

test/Integration/IntegrationTestBase.cs
JatinSanghvi: I will be removing the check for null as I haven't ever seen a case when the event is set to null and none of the examples in Microsoft Docs ([1](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.errordatareceived?view=net-6.0#:~:text=String.IsNullOrEmpty(outLine.Data)), [2](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.errordatareceived?view=net-6.0#:~:text=String.IsNullOrEmpty(errLine.Data)), [3](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.outputdatareceived?view=net-6.0#:~:text=String.IsNullOrEmpty(e.Data))) check for it.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -97,13 +97,51 @@ public async Task MultiOperationTriggerTest()
...
+
+            var changes = new List<SqlChange<Product>>();
+            DataReceivedEventHandler[] changeHandlers = new[] { this.GetProductChangeHandler(changes, "SQL Changes: ") };
+
+            // Prepare three function host processes
+            this.StartFunctionHost(nameof(ProductsTrigger), Common.SupportedLanguages.CSharp, useTestFolder: false, changeHandlers);
+            this.StartFunctionHost(nameof(ProductsTrigger), Common.SupportedLanguages.CSharp, useTestFolder: false, changeHandlers);
+            this.StartFunctionHost(nameof(ProductsTrigger), Common.SupportedLanguages.CSharp, useTestFolder: false, changeHandlers);
+
+            // Considering the polling interval of 5 seconds and batch-size of 10, it should take around 15 seconds to
```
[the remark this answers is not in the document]
JatinSanghvi: If we update the default batch-size and polling interval, I would like to have those reflected in all of the test cases, as the tests are in a way testing that those defaults are being honored correctly. For example, this test case also ensures that all changes are consumed 3x faster since as there are 3 functions hosts doing the work. Though not the main reason, setting the exact condition for exiting-out the polling will become tricky for 'multiple functions, single host' case (the other PR) where we will need to check that every function has completed consuming the change at its end.

Charles-Gagnon (approved):

PR #349, opened 2022-09-13
Wait for error inside `StartFunctionsHostAndWaitForError` method

### LINKED ISSUE #348: StartFunctionsHostAndWaitForError isn't waiting for the error correctly
https://github.com/Azure/azure-functions-sql-extension/blob/triggerbindings/test/Integration/SqlTriggerBindingIntegrationTests.cs#L251

We create a TaskCompletionSource and complete it, but we're never waiting on the Task. 

This gets a tiny bit more complicated for multiple hosts if that's a consideration, but I don't feel like we need to test the error scenarios with multiple hosts currently since the functionality being tested here is just verifying that it throws immediately on startup and shouldn't really be something that matters at all for multiple functions running at the same time.

JatinSanghvi:
- Closes #348 

It's basically a copy of interim changes that were made in #296 ([commit](https://github.com/Azure/azure-functions-sql-extension/blob/76f4bdb8c1325de5913ee18b7ac577678e2554b4/test/Integration/SqlTriggerBindingIntegrationTests.cs#L265-L276)).

### DIFF
--- a/test/Integration/SqlTriggerBindingIntegrationTests.cs
+++ b/test/Integration/SqlTriggerBindingIntegrationTests.cs
@@ -271,7 +271,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
             this.StartFunctionHost(functionName, Common.SupportedLanguages.CSharp, useTestFolder, OutputHandler);
 
             // The functions host generally logs the error message within a second after starting up.
-            const int BufferTimeForErrorInSeconds = 5;
+            const int BufferTimeForErrorInSeconds = 15;
             bool isCompleted = tcs.Task.Wait(TimeSpan.FromSeconds(BufferTimeForErrorInSeconds));
 
             this.FunctionHost.OutputDataReceived -= OutputHandler;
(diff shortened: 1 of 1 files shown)


test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -269,9 +269,15 @@ void OutputHandler(object sender, DataReceivedEventArgs e)
 
             // All trigger integration tests are only using C# functions for testing at the moment.
             this.StartFunctionHost(functionName, Common.SupportedLanguages.CSharp, useTestFolder, OutputHandler);
+
+            // The functions host generally logs the error message within a second after starting up.
+            const int BufferTimeForErrorInSeconds = 5;
```
Charles-Gagnon, line 274: I'd suggest at least 15 seconds for this, I've had host startups take many seconds before and assuming the test works correctly it doesn't really matter what this value is since it shouldn't be hitting it much.

Charles-Gagnon (approved):

MaddyDev (approved):

PR #357, opened 2022-09-15
Refactor locking logic in class `SqlTableChangeMonitor`

### LINKED ISSUE #286: Improve the design of trigger binding support
This issue includes the work items that will require modifying the trigger binding implementation design.

1. **Check if central locking can be removed in `SqlTableChangeMonitor`**
    Make an issue to look into this please - at the very least I'd like to try and refactor this slightly to make this easier to understand. All this locking stuff I want to go over anyways since right now it seems kind of arbitrary where the locks are acquired/released which could easily lead to issues later on.
    _Originally posted by @Charles-Gagnon in https://github.com/Azure/azure-functions-sql-extension/pull/226#discussion_r933596791_

JatinSanghvi:
Performed well-rounded manual testing (three apps with random long processing time for each batch) to ensure there is no impact.

- Closes #286

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -51,14 +51,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private readonly CancellationTokenSource _cancellationTokenSourceRenewLeases;
         private CancellationTokenSource _cancellationTokenSourceExecutor;
 
-        // The semaphore ensures that mutable class members such as this._rows are accessed by only one thread at a time.
+        // The semaphore gets used by lease-renewal loop to ensure that '_state' stays set to 'ProcessingChanges' while
+        // the leases are being renewed. The change-consumption loop requires to wait for the semaphore before modifying
+        // the value of '_state' back to 'CheckingForChanges'. Since the field '_rows' is only updated if the value of
+        // '_state' is set to 'CheckingForChanges', this guarantees that '_rows' will stay same while it is being
+        // iterated over inside the lease-renewal loop.
         private readonly SemaphoreSlim _rowsLock;
 
         private readonly IDictionary<TelemetryPropertyName, string> _telemetryProps;
 
         private IReadOnlyList<IReadOnlyDictionary<string, string>> _rows;
         private int _leaseRenewalCount;
-        private State _state = State.CheckingForChanges;
+        private State _state;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqlTableChangeMonitor{T}" />> class.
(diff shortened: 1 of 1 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -370,11 +375,9 @@ private async void RunLeaseRenewalLoopAsync()
                     this._logger.LogDebugWithThreadId("BEGIN OpenLeaseRenewalLoopConnection");
                     await connection.OpenAsync(token);
                     this._logger.LogDebugWithThreadId("END OpenLeaseRenewalLoopConnection");
+
                     while (!token.IsCancellationRequested)
                     {
-                        this._logger.LogDebugWithThreadId("BEGIN WaitRowsLock - LeaseRenewal");
-                        await this._rowsLock.WaitAsync(token);
-                        this._logger.LogDebugWithThreadId("END WaitRowsLock - LeaseRenewal");
```
JatinSanghvi, line 377: Waiting on semaphore inside `RenewLeasesAsync` so we have same method where we are waiting and releasing the semaphore.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -398,9 +401,13 @@ private async void RunLeaseRenewalLoopAsync()
 
         private async Task RenewLeasesAsync(SqlConnection connection, CancellationToken token)
         {
-            try
+            this._logger.LogDebugWithThreadId("BEGIN WaitRowsLock - RenewLeases");
+            await this._rowsLock.WaitAsync(token);
+            this._logger.LogDebugWithThreadId("END WaitRowsLock - RenewLeases");
+
+            if (this._state == State.ProcessingChanges)
```
JatinSanghvi, line 408: Have single a `if` statement instead of repetitive checks in try and finally blocks.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -489,10 +490,6 @@ private async Task ClearRowsAsync(bool acquireLock)
         private async Task ReleaseLeasesAsync(SqlConnection connection, CancellationToken token)
         {
             TelemetryInstance.TrackEvent(TelemetryEventName.ReleaseLeasesStart, this._telemetryProps);
-            this._logger.LogDebugWithThreadId("BEGIN WaitRowsLock - ReleaseLeases");
-            // Don't want to change the "_rows" while another thread is attempting to renew leases on them.
-            await this._rowsLock.WaitAsync(token);
-            this._logger.LogDebugWithThreadId("END WaitRowsLock - ReleaseLeases");
             long newLastSyncVersion = this.RecomputeLastSyncVersion();
```
JatinSanghvi, line 493: No need to wait for semaphore here. While the `_state` is set to `ProcessingChanges`, the semaphore should be acquired if and only if either of `_rows`, `_state` or `_leaseRenewalLoop` needs to be modified in change-consumption loop and none of the shared class fields are getting modified. Once the `_state` equals `ProcessingChanges`, it is only `ClearRowsAsync` method that is in charge of modifying the rows, hence it is the only method that should be waiting on semaphore.

Charles-Gagnon (approved):

Issue #330, opened 2022-08-31
Have trigger batch size/polling delay be configurable


Charles-Gagnon:
Right now they are hardcoded to 10 items/5sec respectively. This is quite slow and is something that users will likely want to control based on their needs - this should be something they can pass in through the trigger attribute.

PR #364, opened 2022-09-21
Add configuration for batch size/polling interval

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/330

Adds two configuration values that can be used to control the batch size/polling interval.

I considered adding them as parameters to the trigger attribute as well but decided that was unnecessary overhead - it's simple enough to use an app setting for them.

I've created https://github.com/Azure/azure-functions-sql-extension/issues/330 to follow up with changing the defaults, that will require additional testing so wanted to get this in for now and then we can take our time finding new defaults without worrying about people being blocked.

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -26,16 +26,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
     /// <typeparam name="T">POCO class representing the row in the user table</typeparam>
     internal sealed class SqlTableChangeMonitor<T> : IDisposable
     {
-        public const int MaxAttemptCount = 5;
-        private const string CONFIG_KEY_SQL_TRIGGER_BATCHSIZE = "Sql_Trigger_BatchSize";
-        private const string CONFIG_KEY_SQL_TRIGGER_POLLINGINTERVALMS = "Sql_Trigger_PollingIntervalMs";
-        // Leases are held for approximately (LeaseRenewalIntervalInSeconds * MaxLeaseRenewalCount) seconds. It is
+        #region Constants
+        /// <summary>
+        /// The maximum number of times we'll attempt to process a change before giving up
+        /// </summary>
+        private const int MaxChangeProcessAttemptCount = 5;
+        /// <summary>
+        /// The maximum number of times that we'll attempt to renew a lease be
+        /// </summary>
+        /// <remarks>
+        /// Leases are held for approximately (LeaseRenewalIntervalInSeconds * MaxLeaseRenewalCount) seconds. It is
         // required to have at least one of (LeaseIntervalInSeconds / LeaseRenewalIntervalInSeconds) attempts to
         // renew the lease succeed to prevent it from expiring.
-        public const int MaxLeaseRenewalCount = 10;
-        public const int LeaseIntervalInSeconds = 60;
-        public const int LeaseRenewalIntervalInSeconds = 15;
-        public const int MaxRetryReleaseLeases = 3;
+        // </remarks>
+        private const int MaxLeaseRenewalCount = 10;
+        private const int LeaseIntervalInSeconds = 60;
+        private const int LeaseRenewalIntervalInSeconds = 15;
+        private const int MaxRetryReleaseLeases = 3;
+        #endregion Constants
+
         private readonly string _connectionString;
         private readonly int _userTableId;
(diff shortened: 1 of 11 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -25,10 +26,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
     /// <typeparam name="T">POCO class representing the row in the user table</typeparam>
     internal sealed class SqlTableChangeMonitor<T> : IDisposable
     {
-        public const int BatchSize = 10;
-        public const int PollingIntervalInSeconds = 5;
         public const int MaxAttemptCount = 5;
```
JatinSanghvi, line 29: Can you make `MaxAttemptCount` private too?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -131,7 +142,7 @@ public void Dispose()
         /// </summary>
         private async Task RunChangeConsumptionLoopAsync()
         {
-            this._logger.LogDebugWithThreadId("Starting change consumption loop.");
+            this._logger.LogInformationWithThreadId($"Starting change consumption loop. BatchSize: {this._batchSize} PollingIntervalMs: {this._pollingIntervalInMs}");
```
JatinSanghvi, line 145: I think such information should be logged at debug level as user won't be interested in generally in internal workings of the extension. We can log the evaluated batch size and polling interval as separate message.

JatinSanghvi (approved):

lucyzhang929 (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: I'm still a little unsure about exactly what should be info and what should be debug, but does this sound reasonable?

info - information useful to the user to see during normal execution
debug - detailed information about the inner workings

The reason I changed this to info was because I was thinking that a user would find it useful to know the batch size and polling interval - especially for the default values which they may not know offhand.

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -55,6 +55,49 @@ public async Task SingleOperationTriggerTest()
...
+                { "Sql_Trigger_BatchSize", "20" }
+            });
+
+            var changes = new List<SqlChange<Product>>();
+            this.MonitorProductChanges(changes, "SQL Changes: ");
+
+            // Considering the polling interval of 5 seconds and batch-size of 20, it should take around 10 seconds to
+            // process 40 insert operations.
+            this.InsertProducts(1, 40);
+            await Task.Delay(TimeSpan.FromSeconds(12));
```
Charles-Gagnon, line 76: I plan on coming back and refactoring this logic to not be a hardcoded wait since that's more unstable and less reliable. Will be done separately though since I want to update all the places that are using this kind of logic at once though.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: I'm going to leave this as info for now for the reasons above, and we can come back and do a look over logging in the entire extension pre-GA to make sure everything lines up how we want (I'll make an issue to track)

JatinSanghvi: I already mentioned that we should log batch size and polling interval but not the rest of the message.

Users normally use the logs to check the history of their own function invocations. The extension should in general log all the information that might be useful for the user during host/listener startup only. Logging "Starting change consumption loop" even if for one-time will only confuse users and will prompt them to check what the message means for them.

test/Integration/SqlTriggerBindingIntegrationTests.cs
JatinSanghvi: If you are planning on updating the wait-logic, should I keep the changes in my PR ready to be committed? I have been busy with Hackathon so haven't picked up any work on SQL extension items this week.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Oh sorry, I misread your original comment. What would you suggest the info message with the batch size/polling interval would be then?

Issue #347, opened 2022-09-12
Log message missing primary key details


jpomfret:
Just noticed in the logs that the details of the primary key are missing, imagine it should contain the column details of the PK on the table:

> DB and Table: dbName.[dbo].[tblName]. Primary keys: []. SQL Column and Definitions: [statsid int,collectiondate datetime...]

![image](https://user-images.githubusercontent.com/981370/189634127-63ea8ab2-6f46-4df1-b6af-4c1ea0ec2597.png)

This is being used as an output binding - merging into that table

lucyzhang929:
Hi @jpomfret, the logs should contain the names of the primary key(s) in your table. I tested the AddProducts function in our samples/samples-csharp/ folder and I see the following in my logs:
`DB and Table: test.dbo.Products. Primary keys: [ProductID]. SQL Column and Definitions:  [productid int,name nvarchar(100),cost int]`
Are you using the latest version of the sql extension (v0.1.346)?

jpomfret:
Hi there,
I don't see that outputted anywhere in the logs and I didn't specify a specific version (that I know about), where can I check that to get the latest version?

Thanks

lucyzhang929:
What language is your Functions App in? For CSharp you can check the version of Microsoft.Azure.WebJobs.Extensions.Sql  in your .csproj file. For Python/JavaScript/TypeScript, the sql extension is added through the extension bundle and that should have the latest version.

Would you be able to share a little more about your table schema to help us debug? 
- What are the names and datatypes of your primary keys?
- Is your table case sensitive? 
- Are there any identity columns part of your primary keys?

jpomfret:
Hi @lucyzhang929,

The functions app is written in Python (I would have used PowerShell, but the bindings for Azure SQL DB aren't available yet), hence I don't know what I'm doing :) 

I have this in my `host.json` file:

```JSON
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
    "version": "[4.*, 5.0.0)"
  }
```

definition of Primary Key
```SQL 
ALTER TABLE [dbo].[stats] ADD PRIMARY KEY CLUSTERED 
(
	[statsId] ASC
) WITH (STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ONLINE = OFF, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
GO
```

[...]

PR #367, opened 2022-09-23
log primarykeys

lucyzhang929:
This PR fixes: https://github.com/Azure/azure-functions-sql-extension/issues/347.

We were logging the primary keys in the POCO but this wouldn't always include the primary keys that are identity columns.


Charles-Gagnon (approved):
Oops! Thanks for tracking this down.

MaddyDev (approved):

PR #370, opened 2022-09-26
update samples sdk version to match the extension

MaddyDev:
Building C# samples project errors out with the below on Windows:
image

Updating it to match the sdk version used by the extension fixes it.

### DIFF
--- a/samples/samples-csharp/global.json
+++ b/samples/samples-csharp/global.json
@@ -1,6 +1,0 @@
-{
-  "sdk": {
-    "version": "6.0.300",
-    "rollForward": "latestFeature"
-  }
-}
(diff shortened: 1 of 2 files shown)


samples/samples-csharp/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.419",
+    "version": "6.0.300",
```
Charles-Gagnon, line 3: We should just get rid of the global.json here and let it bubble up to the top one then.

MaddyDev:
oh well, the builds on mac and linux failed, probably why we have a separate sdk version. Will check on my mac again (It passed on my mac before)

MaddyDev:
Closing this since it's specific to my machine.

PR #377, opened 2022-09-27
Sync Logic Apps to latest

Charles-Gagnon:
Get latest changes to configuration in in preparation for releasing new package.


Charles-Gagnon:
Merging in immediately since I'll be kicking off product build manually to generate the new package which will run the tests.

PR #382, opened 2022-10-05
Add sql extension exposing binding metadata for out of proc

MaddyDev:
Added Sql extension for out of proc inside our webjobs extension as suggested from our sync. 

What does the extension have : extension implementation defines the attributes (generally, simply exposing the properties from the WebJobs attributes) and brings an assembly level attribute to indicate what Functions/WebJobs extension they support (here’s an example: https://github.com/Azure/azure-functions-dotnet-worker/blob/d60f190cf30b0bdaaf470cb6494022d9aaba16dc/extensions/Worker.Extensions.SendGrid/src/Properties/AssemblyInfo.cs#L6)

Will be sending a separate PR for samples.


Worker.Extension.Sql/src/Microsoft.Azure.Functions.Worker.Extension.Sql.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<AssemblyName>Microsoft.Azure.Functions.Worker.Extension.Sql</AssemblyName>
+		<RootNamespace>Microsoft.Azure.Functions.Worker.Extension.Sql</RootNamespace>
+		<Description>Sql extension for .NET isolated functions</Description>
+
+		<!--Version information-->
+		<VersionPrefix>1.0.0</VersionPrefix>
```
Charles-Gagnon, line 9: If you're going to put a version in here it should be a dev-only one (such as 99.99.99 like https://github.com/Azure/azure-functions-sql-extension/blob/main/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj)

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extension.Sql
+{
+    public sealed class SqlInputAttribute : InputBindingAttribute
+    {
+        /// <summary>
+        /// Constructs a new instance.
```
Charles-Gagnon, line 11: This isn't a super helpful description - it should describe what the class is actually doing

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+        public SqlInputAttribute(string commandText, System.Data.CommandType commandType, string connectionStringSetting)
+        {
+            this.CommandText = commandText;
+            this.CommandType = commandType;
+            this.ConnectionStringSetting = connectionStringSetting;
+        }
+
+        /// <summary>
+        /// The name of the app setting where the SQL connection string is stored
+        /// (see https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection?view=sqlclient-dotnet-core-2.0).
```
Charles-Gagnon, line 25: Clean up any URLs you have in docs : 

- Remove lang IDs such as en-us
- Remove any unnecessary query parameters

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// The attributes specified in the connection string are listed here
+        /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnection.connectionstring?view=sqlclient-dotnet-core-2.0
+        /// For example, to create a connection to the "TestDB" located at the URL "test.database.windows.net" using a User ID and password,
+        /// create a ConnectionStringSetting with a name like SqlServerAuthentication. The value of the SqlServerAuthentication app setting
+        /// would look like "Data Source=test.database.windows.net;Database=TestDB;User ID={userid};Password={password}".
+        /// </summary>
+        public string ConnectionStringSetting { get; set; }
+
+        /// <summary>
+        /// For an input binding, either a SQL query or stored procedure that will be run in the database referred to in the ConnectionString.
```
Charles-Gagnon, line 35: This is specifically an input binding so it seems unnecessary to call out specifically

(make sure you check all the other doc descriptions for the same kind of thing)

performance/packages.lock.json
```diff
@@ -2088,37 +2088,37 @@
...
         "type": "Project",
         "dependencies": {
-          "Microsoft.ApplicationInsights": "2.17.0",
-          "Microsoft.AspNetCore.Http": "2.1.22",
-          "Microsoft.Azure.WebJobs": "3.0.31",
-          "Microsoft.Data.SqlClient": "3.0.1",
-          "Newtonsoft.Json": "13.0.1",
-          "System.Runtime.Caching": "4.7.0",
-          "morelinq": "3.3.2"
+          "Microsoft.ApplicationInsights": "[2.17.0, )",
```
Charles-Gagnon, line 2091: Why are these changing? Undo them if they aren't necessary.

Charles-Gagnon (changes requested):

Charles-Gagnon:
I would also like to see the README updated in this PR so that we have docs on how to use this functionality.

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extension.Sql
+{
+    public sealed class SqlInputAttribute : InputBindingAttribute
```
Charles-Gagnon, line 8: So why does the out of proc have separate attributes for input/output but in-proc just has a single one that's shared? @fabiocav is there a reason you'd know of why it seems to be mostly done like this across all the extensions?

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Sql", "0.1.457-preview")]
```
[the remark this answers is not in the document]
MaddyDev: This is one of the requirements for pulling the correct package version for out-of-proc functions that @fabiocav mentioned. I'll look into the [AssemblyMetadata](https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2019#assemblymetadata) attribute but it says it only applies to projects using the SDK for .NET 5 (and .NET Core) and later versions.

performance/packages.lock.json
MaddyDev: dotnet restore updated them automatically, will check and undo them if unnecessary.

Worker.Extension.Sql/src/SqlInputAttribute.cs
MaddyDev: I just copy pasted from Sql Attribute, should we clean up there also?

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
Charles-Gagnon, line 6: I would prefer that we use the same version number as the build for this instead of hardcoding it - that way we save ourselves having to update this every time we want to release (and when building/running from source we should be using the local dev copy of the extension anyways - otherwise we end up with situations like the python tests where we're only testing with released code)

So the version here would be "99.99.99" and then during the pipeline builds we'd replace that value with the actual nuget package version we generated.

MaddyDev: I chatted with @fabiocav regarding have it updated during build and he said "You want to, at the very least, lock to a given major version. You can use "x.*" (e.g., 3.*), which would get the highest 3.x release", Am checking with him if we can do 0.1.*-preview

Charles-Gagnon: You can just try it out yourself, right? It sounds like there isn't an issue with using version ranges and I definitely agree locking it to a major version (+ -preview) is what we'd want to do right now. `0.*-preview` should be good if that's allowed.

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Sql", "0.1.*-preview")]
```
Charles-Gagnon, line 6: Please make a follow up issue to make this dynamically generated during build - we still don't want to have to remember to come back and update this later on when we bump version/get out of preview. And it should be pretty straightforward to do this using the pipeline variables I showed you earlier.

Worker.Extension.Sql/src/SqlOutputAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// The table name.
+        /// </summary>
+        public string CommandText { get; }
+
+        /// <summary>
+        /// Specifies <see cref="CommandText"/> which refers to the table name.
+        /// </summary>
+        public System.Data.CommandType CommandType { get; set; } = System.Data.CommandType.Text;
```
Charles-Gagnon, line 41: Is this needed for some reason? We should always be using Text so we shouldn't expose this for users if we can

Worker.Extension.Sql/src/SqlOutputAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// The attributes specified in the connection string are listed here
+        /// https://docs.microsoft.com/dotnet/api/microsoft.data.sqlclient.sqlconnection.connectionstring
+        /// For example, to create a connection to the "TestDB" located at the URL "test.database.windows.net" using a User ID and password,
+        /// create a ConnectionStringSetting with a name like SqlServerAuthentication. The value of the SqlServerAuthentication app setting
+        /// would look like "Data Source=test.database.windows.net;Database=TestDB;User ID={userid};Password={password}".
+        /// </summary>
+        public string ConnectionStringSetting { get; set; }
+
+        /// <summary>
+        /// The table name.
```
Charles-Gagnon, line 34: `The table name to upsert the values to`

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
MaddyDev: https://github.com/Azure/azure-functions-sql-extension/issues/404

Charles-Gagnon, line 6: I don't think we need to include the minor version here too - just `0.*-preview` should be enough (assuming we plan to keep following proper semver versioning)

@chlafreniere @lucyzhang929 thoughts?

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /// <summary>
+        /// Creates an instance of the <see cref="SqlAttribute"/>, specifying the Sql attributes
+        /// the function supports.
+        /// </summary>
+        /// <param name="commandText">The text of the command.</param>
+        /// <param name="commandType">Specifies whether commandText refers to a stored procedure or SQL query string.</param>
+        /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored.</param>
+        public SqlInputAttribute(string commandText, System.Data.CommandType commandType, string connectionStringSetting)
```
lucyzhang929, line 17: Do we need to set the parameters here too?

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
chlafreniere: I don't have a strong opinion here - it's pretty unlikely that we'll go to 0.2.*, but totally fine just to be a bit more inclusive version range-wise in case I'm wrong 😄

Worker.Extension.Sql/src/SqlOutputAttribute.cs
MaddyDev: I was contemplating that as well and added it since we have it defined for all the languages defaulted. If it always be text, I can default it and remove the set.

Worker.Extension.Sql/src/SqlInputAttribute.cs
MaddyDev: what do you mean here too?

Worker.Extension.Sql/src/Properties/AssemblyInfo.cs
MaddyDev: Will update it to 0.*-preview.

Worker.Extension.Sql/src/SqlInputAttribute.cs
Charles-Gagnon: Default parameters I assume - mostly for the commandType where we default it to Text in the property itself.

Worker.Extension.Sql/src/SqlOutputAttribute.cs
Charles-Gagnon: Not likely, and if we do we can add it back then. I'd remove it to reduce potential confusion/mistakes. We have the CommandType for the general binding attribute due to limitations there with that attribute being used for both input and output.

Worker.Extension.Sql/src/SqlInputAttribute.cs
lucyzhang929: Ah I was referring to settings the Parameters field but am now realizing that is not needed (getting things mixed up with the Java classes I have been writing 😅 ). 
Looking at the in proc SqlAttribute constructor, we only take in the commandText. Is there a reason we are doing it differently here?
https://github.com/Azure/azure-functions-sql-extension/blob/66a1d06a2fede766c57842d4f297438d2d500a79/src/SqlAttribute.cs#L23

MaddyDev: @lucyzhang929 I was looking at the samples attribute definition instead of SqlAttribute here and added them, you were right we don't need the other params here since they have getters and setters defined appropriately.

Worker.Extension.Sql/src/SqlOutputAttribute.cs
MaddyDev: alrighty, updated it to text.

Worker.Extension.Sql/src/SqlInputAttribute.cs
Charles-Gagnon: That being said - I honestly don't know why we don't have constructors with the additional parameters to begin with. That just seems useful to have (especially if dynamically creating a binding which is supported). 

Anyways this is fine to keep consistent and we can update later on if we have a reason to.

Worker.Extension.Sql/src/SqlInputAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Microsoft.Azure.Functions.Worker.Extension.Sql
+{
+    public sealed class SqlInputAttribute : InputBindingAttribute
+    {
+        /// <summary>
+        /// Creates an instance of the <see cref="SqlAttribute"/>, specifying the Sql attributes
+        /// the function supports.
+        /// </summary>
+        /// <param name="commandText">The text of the command.</param>
+        /// <param name="commandType">Specifies whether commandText refers to a stored procedure or SQL query string.</param>
```
Charles-Gagnon, line 15: Remove unused params

Worker.Extension.Sql/src/SqlOutputAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Microsoft.Azure.Functions.Worker.Extension.Sql
+{
+    public class SqlOutputAttribute : OutputBindingAttribute
+    {
+        /// <summary>
+        /// Creates an instance of the <see cref="SqlAttribute"/>, specifying the Sql attributes
+        /// the function supports.
+        /// </summary>
+        /// <param name="commandText">The text of the command.</param>
+        /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored.</param>
```
Charles-Gagnon, line 15: Remove unused param

Charles-Gagnon (approved):

lucyzhang929 (approved):

MaddyDev:
> I would also like to see the README updated in this PR so that we have docs on how to use this functionality.

I've been working on a README, will add it along with the samples. Created https://github.com/Azure/azure-functions-sql-extension/issues/407 for tracking this.

Issue #385, opened 2022-10-06
Serialization issue with Date in JavaScript Output Bindings


lucyzhang929:
Original issue (IAsyncCollector section): https://github.com/Azure/azure-functions-sql-extension/issues/316

We added ISO DateTime Format to the serialization settings but as noted [here](https://github.com/Azure/azure-functions-sql-extension/blob/a916d353b0f04dad8f66cbcc1e8944bf512fc8c2/src/SqlAsyncCollector.cs#L339), the serialization settings are ignored in JavaScript.

If the Date passed in is not ISO format, we get the following error
`Conversion failed when converting date and/or time from character string.`

MaddyDev:
@lucyzhang929 Is this still a valid issue in JavaScript since we made some changes overall to default time to ISO format and removed the code mentioned [here](https://github.com/Azure/azure-functions-sql-extension/blob/a916d353b0f04dad8f66cbcc1e8944bf512fc8c2/src/SqlAsyncCollector.cs#L339)

lucyzhang929:
In CSharp, we always convert the date to ISO format but this doesn't happen for non-csharp languages. So if a date is passed in a different format, the database will throw an error.

I'll close this issue for now since we documented the expected format in the readme and all our samples use ISO format for datetime.

PR #386, opened 2022-10-06
Fix folder used for Integration tests

lucyzhang929:
This PR fixes the working directory being used for integration tests. Previously the tests in test/Integration/test-js were not being used.

Also converted the Date in the AddProductColumnTypes function to ISO format. Without ISO format, we get the following error from SQL Database: `Conversion failed when converting date and/or time from character string.`
Issue documented here: https://github.com/Azure/azure-functions-sql-extension/issues/385

### DIFF
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -178,7 +178,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         /// </remarks>
         protected void StartFunctionHost(string functionName, SupportedLanguages language, bool useTestFolder = false, DataReceivedEventHandler customOutputHandler = null)
         {
-            string workingDirectory = useTestFolder ? GetPathToBin() : Path.Combine(GetPathToBin(), "SqlExtensionSamples", Enum.GetName(typeof(SupportedLanguages), language));
+            string workingDirectory = language == SupportedLanguages.CSharp && useTestFolder ? GetPathToBin() : Path.Combine(GetPathToBin(), "SqlExtensionSamples", Enum.GetName(typeof(SupportedLanguages), language));
             if (!Directory.Exists(workingDirectory))
             {
                 throw new FileNotFoundException("Working directory not found at " + workingDirectory);
--- a/test/Integration/SqlOutputBindingIntegrationTests.cs
+++ b/test/Integration/SqlOutputBindingIntegrationTests.cs
@@ -106,7 +106,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         /// <param name="lang">The language to run the test against</param>
         [Theory]
         [SqlInlineData()]
-        [UnsupportedLanguages(SupportedLanguages.JavaScript)] // Temporarily skipping while we look into the datetime conversion issue with JavaScript
         public void AddProductColumnTypesTest(SupportedLanguages lang)
         {
             this.StartFunctionHost(nameof(AddProductColumnTypes), lang, true);
--- a/test/Integration/SqlOutputBindingIntegrationTests.cs
+++ b/test/Integration/SqlOutputBindingIntegrationTests.cs
@@ -106,6 +106,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         /// <param name="lang">The language to run the test against</param>
         [Theory]
         [SqlInlineData()]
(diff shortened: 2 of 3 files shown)


chlafreniere:
> Previously the tests in test/Integration/test-js were not being used.

Can you expand on this, @lucyzhang929?

test/Integration/IntegrationTestBase.cs
```diff
@@ -178,7 +178,7 @@ protected void StartAzurite()
         /// </remarks>
         protected void StartFunctionHost(string functionName, SupportedLanguages language, bool useTestFolder = false, DataReceivedEventHandler customOutputHandler = null)
         {
-            string workingDirectory = useTestFolder ? GetPathToBin() : Path.Combine(GetPathToBin(), "SqlExtensionSamples", Enum.GetName(typeof(SupportedLanguages), language));
+            string workingDirectory = language == SupportedLanguages.CSharp && useTestFolder ? GetPathToBin() : Path.Combine(GetPathToBin(), "SqlExtensionSamples", Enum.GetName(typeof(SupportedLanguages), language));
```
Charles-Gagnon, line 181: Actually - I don't even think we need this anymore if you just update https://github.com/Azure/azure-functions-sql-extension/blob/a20541cfedb4a774395a364454f419f546b5e470/test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj to also copy the CS bindings from `Integration\test-cs` into the SqlExtensionSamples folder. There's a bit of extra duplication there with having multiple copies of the functions, but it makes the test code simpler and running the functions from the same place will make it easier to diagnose issues. 

I would also suggest when you do that updating the folder to just be generically named `TestFunctions`

lucyzhang929:
@chlafreniere Our invalid scenario functions are all in the test/Integration/test-* folders and the way that we were setting up the workingDirectory previously always used the CSharp functions to test the invalid scenarios. Valid scenarios (in samples folder) were being run correctly.

test/Integration/IntegrationTestBase.cs
lucyzhang929: Yes that makes sense - I'll update that. Thanks Charles!

Charles-Gagnon:
Hrm, yeah that's a problem that it was just silently doing that and no one noticed...I wonder if we should update the function names to include the language as well to prevent this in the future....

test/Integration/IntegrationTestBase.cs
lucyzhang929: One issue I’m running into is that we need to copy the built function files (ex. AddProductsColumnTypes folder containing the function.json file) and not the .cs files. The built function folders are in bin/Debug/netcoreapp3.1/ but I’m not sure if there is a way to only copy those folders into the TestFunctions folder.   
Screen Shot 2022-10-06 at 6 30 29 PM

Another option is to move the functions in the test folder into a separate folder in samples ex. Samples/samples-csharp/InvalidSamples. Then all functions can be built in the same project. This will also make it easier for Java so that we don't have to build two java projects: samples/samples-java and test/Integration/test-java. @Charles-Gagnon @MaddyDev thoughts?

Charles-Gagnon: I'd rather avoid having the invalid (test) functions in the samples - that's just confusing for the users and could lead to someone not paying attention and trying to do something that isn't going to work. 

You should be able to just use the same logic we have for adding the items from the sample folder - we just need to special case to not copy over the target folder.

```xml
      <_CSharpCopyItems Include="..\samples\samples-csharp\bin\$(Configuration)\$(TargetFramework)\**\*.*" />
      <_CSharpCopyItems Include="$(OutDir)\**\function.json" />
      <_CSharpCopyItems Remove="$(OutDir)\SqlExtensionSamples\**\*.*" />
```

lucyzhang929: I agree, having the test functions in the samples can be confusing. 

I tried copying over the function.json files but the issue is that the function.json files for CSharp refers to the dll needed which would be different for the Samples vs Test functions:
`"scriptFile": "../bin/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.dll",`.
I also tried copying over the Test.dll but the functions still don't get loaded correctly. I believe it is because of the `functions.deps.json` file which references the target project and would be different for Samples vs Tests.

For now we will have to stick to two separate folders for CSharp tests (I think Java will also need to separate folders).

Charles-Gagnon: Ah dang, yeah ok let's just leave it like it is then with two completely separate folders. Thanks for looking into that.

Issue #383, opened 2022-10-05
auto-generated identity column not getting auto-generated


engenb:
I'm working with:
Microsoft.Azure.WebJobs.Extensions.Sql 0.1.457-preview

I have a table such as:
```sql
CREATE TABLE [dbo].[ReportJobs](
	[Id] [uniqueidentifier] NOT NULL,
	...

ALTER TABLE [dbo].[ReportJobs] ADD  DEFAULT (newsequentialid()) FOR [Id]
```

I have a C# poco:
```csharp
public class ReportJob
{
    public Guid Id { get; set; }
    ...
}
```
and a function with a sql output binding:
```csharp
[Sql($"dbo.{nameof(ReportJob)}s", ConnectionStringSetting = $"{nameof(ReportFactoryOptions.ConnectionStrings)}:{ReportFactoryOptions.ConnectionStrings.Database}")]
IAsyncCollector<ReportJob> reportJobs,
```

I've reviewed #189 and the updates to README.md and I believe I'd fall under scenario 2 of the [Primary Keys and Identity Columns](https://github.com/Azure/azure-functions-sql-extension/blob/main/README.md#primary-keys-and-identity-columns) "If the identity column is included" section:
_"If the identity column is included (even if it's an optional nullable value) then a merge is performed similar to what happens when no identity column is present. This merge will either insert a new row or update an existing row based on the existence of a row that matches the primary keys (including the identity column). See [AddProductWithIdentityColumnIncluded](https://github.com/Azure/azure-functions-sql-extension/blob/main/samples/samples-csharp/OutputBindingSamples/AddProductWithIdentityColumnIncluded.cs) for an example."_

^ the code comments in that linked sample indicate:
_"...the identity column is included in the output object. In this case the identity column is used to match the rows and all non-identity columns are updated if a match is found. Otherwise a new row is inserted (with the identity column being ignored since that will be generated by the engine)."_

The way I interpret those statements is, if I leave my Id as `default` (empty Guid) it won't match any existing rows and the identity column should be ignored and generated by the db.  However, that doesn't seem to be the case.  I've run sql profiler, and I can see that the generated script is not omitting Id, but is instead supplying the Id value with the insert.

```sql
EXEC sp_executesql 
[...]

PR #388, opened 2022-10-10
Allow columns with default values to be excluded from the POCO

lucyzhang929:
This PR fixes: https://github.com/Azure/azure-functions-sql-extension/issues/383

If a table has a primary key with with default values, that column can be excluded from the POCO and will get generated by the engine.


src/SqlAsyncCollector.cs
```diff
@@ -426,13 +430,19 @@ public static string GetPrimaryKeysQuery(SqlObject table)
             {
                 return $@"
                     SELECT
-                        {ColumnName}, c.is_identity
+                        ccu.{ColumnName}, c.is_identity,
```
Charles-Gagnon, line 433: Let's put `c.is_identity` on a separate line to make it easier to read

README.md
```diff
@@ -43,7 +43,7 @@ Azure SQL bindings for Azure Functions are supported for:
       - [ICollector&lt;T&gt;/IAsyncCollector&lt;T&gt;](#icollectortiasynccollectort)
       - [Array](#array)
       - [Single Row](#single-row)
-      - [Primary Keys and Identity Columns](#primary-keys-and-identity-columns)
+      - [Special Cases](#special-cases)
```
Charles-Gagnon, line 46: Maybe `Primary Key Special Cases`

README.md
```diff
@@ -765,11 +765,17 @@ Normally Output Bindings require two things :
 
 If either of these are false then an error will be thrown.
```
Charles-Gagnon, line 766: `Normally if either of these are false then an error will be thrown. Below are the situations in which this is not the case :`

(and then probably reword the first sentence in each section to read a bit better)

Charles-Gagnon (approved):

chlafreniere (approved):

PR #389, opened 2022-10-11
Add runtime driven scaling support

### LINKED ISSUE #351: Elastic premium scaling support

JatinSanghvi:
- Closes #351.


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -159,6 +159,46 @@ public void Dispose()
...
+
+            try
+            {
+                long getUnprocessedChangesDurationMs = 0L;
+
+                using (var connection = new SqlConnection(this._connectionString))
+                {
+                    this._logger.LogDebugWithThreadId("BEGIN GetUnprocessedChangeCountAsync");
+                    await connection.OpenAsync();
+                    this._logger.LogDebugWithThreadId("END GetUnprocessedChangeCountAsync");
```
Charles-Gagnon, line 174: This should go at the end

Although given you already have logs for GetUnprocessedChangeCount below I don't think you even need to keep these ones since this method is only doing one thing.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+                Vote = ScaleVote.None
+            };
+
+            // Do not make a scale decision unless we have enough samples.
+            if (metrics == null || (metrics.Length < minMetricSamplesForScaling))
+            {
+                return status;
+            }
+
+            // Add worker is count of unprocessed changes per worker exceeds the maximum limit.
```
Charles-Gagnon, line 516: `Add worker if...` ?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+            long lastUnprocessedChangeCount = metrics.Last().UnprocessedChangeCount;
+            this._logger.LogInformation($"Unprocessed change count: {lastUnprocessedChangeCount}, worker count: {workerCount}, maximum changes per worker: {maxChangesPerWorker}.");
+
+            if (lastUnprocessedChangeCount > workerCount * maxChangesPerWorker)
+            {
+                status.Vote = ScaleVote.ScaleOut;
+                this._logger.LogInformation($"Requesting scale-out: Found too many unprocessed changes for table: '{this._userTable.FullName}' relative to the number of workers.");
```
Charles-Gagnon, line 523: This seems like a place we should be adding telemetry

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+        {
+            return this.GetScaleStatusCore(context.WorkerCount, context.Metrics?.Cast<SqlTriggerMetrics>().ToArray());
+        }
+
+        public ScaleStatus GetScaleStatus(ScaleStatusContext<SqlTriggerMetrics> context)
+        {
+            return this.GetScaleStatusCore(context.WorkerCount, context.Metrics?.ToArray());
+        }
+
+        private ScaleStatus GetScaleStatusCore(int workerCount, SqlTriggerMetrics[] metrics)
```
Charles-Gagnon, line 500: Please add a comment at the top of this method explaining the rationale behind how the decision to scale out/scale in is made.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+            // Check to see if there are no changes occurring in the user table for a while.
+            if (metrics.All(p => p.UnprocessedChangeCount == 0L))
+            {
+                status.Vote = ScaleVote.ScaleIn;
+                this._logger.LogInformation($"Requesting scale-in: Found table: '{this._userTable.FullName}' to be idle for a while.");
+                return status;
+            }
+
+            // Check if there is a continuous increase in count of unprocessed changes.
+            if (metrics[0].UnprocessedChangeCount > 0)
```
Charles-Gagnon, line 536: Why only the first in the array? Is that going to be the first (earliest) or last (latest) metrics gathered?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+                if (isIncreasing)
+                {
+                    status.Vote = ScaleVote.ScaleOut;
+                    this._logger.LogInformation($"Requesting scale-out: Found the unprocessed changes for table: '{this._userTable.FullName}' to be continuously increasing.");
+                    return status;
+                }
+            }
+
+            // Check if there is a continuous decrease in count of unprocessed changes.
+            bool isDecreasing = true;
```
Charles-Gagnon, line 557: You can combine this into the above loop - there's no reason to loop through the metrics twice

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
             }
         }
 
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await this.GetMetricsAsync();
```
Charles-Gagnon, line 476: Why not just put the code directly in this method?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+
+            if (lastUnprocessedChangeCount > workerCount * maxChangesPerWorker)
+            {
+                status.Vote = ScaleVote.ScaleOut;
+                this._logger.LogInformation($"Requesting scale-out: Found too many unprocessed changes for table: '{this._userTable.FullName}' relative to the number of workers.");
+                return status;
+            }
+
+            // Check to see if there are no changes occurring in the user table for a while.
+            if (metrics.All(p => p.UnprocessedChangeCount == 0L))
```
Charles-Gagnon, line 528: Are the metrics cleared every once in a while then?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -464,6 +471,110 @@ PRIMARY KEY ({primaryKeys})
...
+                    this._logger.LogInformation($"Requesting scale-out: Found the unprocessed changes for table: '{this._userTable.FullName}' to be continuously increasing.");
+                    return status;
+                }
+            }
+
+            // Check if there is a continuous decrease in count of unprocessed changes.
+            bool isDecreasing = true;
+            for (int index = metrics.Length - minMetricSamplesForScaling; index < metrics.Length - 1; index++)
+            {
+                if (metrics[index].UnprocessedChangeCount == 0 || metrics[index].UnprocessedChangeCount > metrics[index + 1].UnprocessedChangeCount)
```
Charles-Gagnon, line 560: Why do we not consider it decreasing if one of the metrics has 0 unprocessed changes? If I have metrics like this 

100 -> 75 -> 50 -> 25 -> 0

then we should scale it down since it's clearly decreasing

Charles-Gagnon (changes requested):
Please also add a section to the README giving a brief overview of the scaling and how it determines whether to scale up or scale down.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: It is going to be the earliest in the window. Most of the logic is copied from the [Azure Queue listener](https://github.com/Azure/azure-webjobs-sdk/blob/919d2d63419691166a6e36f1c1785f4c2c31aeb3/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs#L519). The check against 0 is not very critical. It only ensures that we scale for metrics [10, 20, 30] and not for [0, 10, 20] as in a way, that means the SQL table was not active for the entirety of the window.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: I reversed the condition by mistake. Please ignore it for now. The intention was to recommend a scale-in even for metric samples like [20, 10, 0, 0, 0] where the plain strictly-decreasing condition would return false.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Seems to be so. As said in the other comment thread, the queue listener carries similar logic so I assumed that the metrics should be getting cleared, and though of confirming this while testing the published function app. I will update this thread with my observation. BTW, I think I should remove this check as this should be covered entirely by the 'isDecreasing' check below.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: The listener needs to expose four methods (two for backward compatibility):

```cs
Task<ScaleMetrics> GetMetricsAsync()
Task<SqlTriggerMetrics> GetMetricsAsync()
ScaleStatus GetScaleStatus(ScaleStatusContext context)
ScaleStatus GetScaleStatus(ScaleStatusContext<SqlTriggerMetrics> context)
```
So, both of the methods are required.

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: I'd probably take their IsTrueForLastN over too - that seems useful and will make this easier to understand.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,110 @@ PRIMARY KEY ({primaryKeys})
             }
         }
 
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await this.GetMetricsAsync();
+        }
+
+        public async Task<SqlTriggerMetrics> GetMetricsAsync()
```
Charles-Gagnon, line 478: How often are metrics collected anyways? Is that a static value?

JatinSanghvi: In the video recordings explaining the scale controller, it was mentioned that the polling happens every 10 seconds. I will try to check the code and paste the link to it here but you may not have access to the repo.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: The queue listener needs to loop 4 times so there it makes sense to move the iteration in its own method. The next commit is going to have a single loop anyhow. I am also finding the `if (metrics[0].UnprocessedChangeCount > 0)` to be arbitrary, so that will also go away.

Another problem I see with the queue-listener logic is that if the function app is lightly loaded, since we are doing comparison only among 5 samples, there is possibility that they turn out to be all in an increasing order e.g. [0, 2, 4, 5, 9]. Since the max capacity per worker is 1000, this should not cause a scale out. Similarly, if we find the metric samples like [1900, 1988, 1955, 1944, 1800] with two workers, this should not cause as scale in, since clearly 1 worker is not going to handle these many requests. I think most of the logic should hover around current number of changes, and the trend of increase/decrease should be a secondary consideration.

JatinSanghvi: For trend, a better approach should be to extrapolate and if we find the number of pending changes will go beyond the total capacity of existing workers, then we should scale out. Same logic can determine when one should scale-in. Of course, this will require a couple of unit tests to ensure there are no bugs.

src/TriggerBinding/SqlTableChangeMonitor.cs
JatinSanghvi: Need to rename `GetUnprocessedChangeCountAsync` to `OpenGetUnprocessedChangesConnection`.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: From code walkthrough too, I found it to be 10 seconds. Links to code:
- [RuntimeScaleActivationTask.cs](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Azure/ScaleController/Microsoft.Azure.AppService.ScaleController.Extension/ActivationTasks/RuntimeScaleActivationTask.cs&version=GBdev&line=45&lineEnd=49&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)
- [ScaleControllerConfiguration.cs](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Websites?path=/src/Hosting/Azure/ScaleController/Microsoft.Azure.AppService.ScaleController.Extension/Shared/ScaleControllerConfiguration.cs&version=GBdev&line=96&lineEnd=108&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: So, I found the number of metrics to be varying between 10 and 12. The logic for this is entirely dictated by functions host runtime. All metrics populated in last two minutes are considered, which is consistent with my observation. Code for reference:

- [ScaleOptions.cs](https://github.com/Azure/azure-functions-host/blob/379e98010c1d789369df6516b0eb5cc60556cdc7/src/WebJobs.Script/Config/ScaleOptions.cs#L25)
- [TableStorageScaleMetricsRepository.cs](https://github.com/Azure/azure-functions-host/blob/379e98010c1d789369df6516b0eb5cc60556cdc7/src/WebJobs.Script.WebHost/Scale/TableStorageScaleMetricsRepository.cs#L324)

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: This is done from my end. The unit tests will be added to verify there are no issues in the next PR.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: Added lines to send telemetry events.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+        {
+            return this.GetScaleStatusWithTelemetry(context.WorkerCount, context.Metrics?.ToArray());
+        }
+
+        private ScaleStatus GetScaleStatusWithTelemetry(int workerCount, SqlTriggerMetrics[] metrics)
+        {
+            var measures = new Dictionary<TelemetryMeasureName, double>
+            {
+                [TelemetryMeasureName.WorkerCount] = workerCount,
+                [TelemetryMeasureName.UnprocessedChangeCount] = (metrics is null || metrics.Length == 0) ? -1L : metrics[metrics.Length - 1].UnprocessedChangeCount,
```
Charles-Gagnon, line 504: This doesn't seem super useful to have without having all the metrics being considered. I'd just have this be a stringified array (that we can process on the server side if needed) of the counts that are being considered - e.g. `[10, 20, 30, 40, 50]`

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+
+            if (isIncreasing)
+            {
+                // Scale out only if the expected count of unprocessed changes would exceed the combined limit after 30 seconds.
+                long expectedUnprocessedChangeCount = (2 * metrics[metrics.Length - 1].UnprocessedChangeCount) - metrics[metrics.Length - 4].UnprocessedChangeCount;
+
+                if (expectedUnprocessedChangeCount > workerCount * maxChangesPerWorker)
+                {
+                    status.Vote = ScaleVote.ScaleOut;
+                    this._logger.LogInformation($"Requesting scale-out: Found the unprocessed changes for table: '{this._userTable.FullName}' to be continuously increasing.");
```
Charles-Gagnon, line 571: `to be continuously increasing and will exceed the maximum changes that the workers are able to process`

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+                // Scale in only if the count of unprocessed changes will not exceed the combined limit post the scale-in operation.
+                if (lastUnprocessedChangeCount <= (workerCount - 1) * maxChangesPerWorker)
+                {
+                    status.Vote = ScaleVote.ScaleIn;
+                    this._logger.LogInformation($"Requesting scale-in: Found table: '{this._userTable.FullName}' to be either idle or the unprocessed changes to be continuously decreasing.");
+                    return status;
+                }
+            }
+
+            this._logger.LogInformation($"Requesting no-scaling: Found the number of unprocessed changes for table: '{this._userTable.FullName}' to be steady.");
```
Charles-Gagnon, line 587: "steady" isn't really accurate here - there's a lot of other logic that goes into whether it determines if it scales out or in. 

I'd just say `Requesting no-scaling: Found the number of unprocessed changes for table: '{this._userTable.FullName}' to not require scaling`.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+                isIncreasing = isIncreasing && metrics[index].UnprocessedChangeCount < metrics[index + 1].UnprocessedChangeCount;
+                isDecreasing = isDecreasing && (metrics[index].UnprocessedChangeCount == 0 || metrics[index].UnprocessedChangeCount > metrics[index + 1].UnprocessedChangeCount);
+            }
+
+            if (isIncreasing)
+            {
+                // Scale out only if the expected count of unprocessed changes would exceed the combined limit after 30 seconds.
+                long expectedUnprocessedChangeCount = (2 * metrics[metrics.Length - 1].UnprocessedChangeCount) - metrics[metrics.Length - 4].UnprocessedChangeCount;
+
+                if (expectedUnprocessedChangeCount > workerCount * maxChangesPerWorker)
```
Charles-Gagnon, line 568: I'd add an else with just a logging statement explaining that the metrics were increasing but won't exceed the limit the workers can handle. 

(same thing for the isDecreasing block below)

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+            {
+                this._logger.LogInformation($"Requesting no-scaling: Insufficient metrics for making scale decision for table: '{this._userTable.FullName}'.");
+                return status;
+            }
+
+            // Add worker if the count of unprocessed changes per worker exceeds the maximum limit.
+            long lastUnprocessedChangeCount = metrics.Last().UnprocessedChangeCount;
+            this._logger.LogInformation($"Unprocessed change count: {lastUnprocessedChangeCount}, worker count: {workerCount}, maximum changes per worker: {maxChangesPerWorker}.");
+
+            if (lastUnprocessedChangeCount > workerCount * maxChangesPerWorker)
```
Charles-Gagnon, line 547: Do we really need this extra check? If it truly is increasing at a rate that requires scaling out then this will just be handled with the below checks. Right now if there's a sudden blip in the number of changes but then it immediately settles down do we really want to be scaling out and then immediately scaling back in?

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: The value needs to be of type double. Let me know if you have any suggestion.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: This is a valid concern, but I think this is a primary check and should be retained. If the load not strictly increasing [1100, 1050, 1150] but above the worker capacity, we should always ensure that per-worker load is maintained below their capacity.

The scale controller has complex logic for interval between scaling out. For example, it sometimes launches new instances at the same frequency of polling, which is 10 seconds while other times, especially for the very first scale-out request after a long period of no scaling (guessing based on my observation), it gives time for the newly scaled app to pick up and start processing the events and contribute to reduction in overall load before scaling out the app further.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+        /// made based on both the latest metrics and the trend of increase or decrease in the count of unprocessed
+        /// changes in the user table. In all of the calculations, it is attempted to keep the number of workers minimum
+        /// while also ensuring that the count of unprocessed changes per worker stays under the maximum limit.
+        /// </summary>
+        /// <param name="workerCount">The current worker count for the host application.</param>
+        /// <param name="metrics">The collection of metrics samples to make the scale decision.</param>
+        /// <returns></returns>
+        private ScaleStatus GetScaleStatusCore(int workerCount, SqlTriggerMetrics[] metrics)
+        {
+            const int minSamplesForScaling = 5;
```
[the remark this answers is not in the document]
JatinSanghvi: These are picked from other listeners and no explanation is provided there. We need to have 5 samples to estimate the trend with some reliability. Below 5 will not be enough. They roughly correspond to past 1 minute of load but in some cases, when the scaling is taking place, the scale controller polls less frequently, so the 5 samples can also span up to 2 minutes.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,124 @@ PRIMARY KEY ({primaryKeys})
...
+            bool isDecreasing = true;
+            for (int index = metrics.Length - minSamplesForScaling; index < metrics.Length - 1; index++)
+            {
+                isIncreasing = isIncreasing && metrics[index].UnprocessedChangeCount < metrics[index + 1].UnprocessedChangeCount;
+                isDecreasing = isDecreasing && (metrics[index].UnprocessedChangeCount == 0 || metrics[index].UnprocessedChangeCount > metrics[index + 1].UnprocessedChangeCount);
+            }
+
+            if (isIncreasing)
+            {
+                // Scale out only if the expected count of unprocessed changes would exceed the combined limit after 30 seconds.
```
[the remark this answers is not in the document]
JatinSanghvi: Sure. The 30 seconds will be time in usual cases, but when scaling is happening, the polls on pending change count will be less frequent and the time difference between last three samples can be as huge as 2 minutes. I will need to complicate the algorithm if we need to add dependency on time when the change-counts were collected. Possibly, only compare the latest count with the first sample that was collected 30-seconds before the last sample.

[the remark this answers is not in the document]
Charles-Gagnon: Sounds good - we can probably stick with this simple algorithm for now (especially since that's similar to what other triggers are doing) and then revisit this later if there seems to be a need to have a more complex decision being made.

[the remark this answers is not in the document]
JatinSanghvi: Let me check here. If this does not require more than 3-4 lines of changes, I will want to add it since I found the intervals between sample to be changing arbitrarily while scaling is in action. Other triggers I checked are not having the logic to estimate future scaling. If there is continuous increase, even if a small one, they scale out the application. I personally think the scaling in and out happens much more aggressively there than required.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: We can average across samples here to not react to one-off blips.

src/TriggerBinding/SqlTriggerListener.cs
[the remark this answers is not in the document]
JatinSanghvi: This is done. The logic did not turn out to be complicated.

src/TriggerBinding/SqlTriggerListener.cs
JatinSanghvi: I haven't made any change here. Will wait for your comment.

README.md
```diff
@@ -872,6 +873,12 @@ The trigger binding utilizes SQL [change tracking](https://docs.microsoft.com/sq
...
+
+There are a couple of checks made to decide on whether the host application needs to be scaled in or out. The rationale behind these checks is to ensure that the count of pending changes per application-worker stays below a certain maximum limit, which is defaulted to 1000, while also ensuring that the number of workers running stays minimal. The scaling decision is made based on the latest count of the pending changes and whether the last 5 times we checked the count, we found it to be continuously increasing or decreasing.
```
Charles-Gagnon, line 880: Add a comment to where the 1000 is defined in the code to also update this text if/when that's changed

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: Just have it be a property instead - it doesn't really make a huge difference which bag it ends up in since we manually process all our events anyways for our reporting. 

I'd also simplify this and just have a single event `GetScaleStatus` that's sent at the end. I don't really remember the reason why we have a bunch of start/end events for the other ones but it generally isn't that useful and just makes combining the data to be more tedious. 

And then also wrap the call to get the scale status in a try/catch and track error event if it fails.

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: Alright, I'm starting to lean towards not overthinking this and getting something out there and we can always change it later if needed. Given that the other triggers don't seem to have a problem with very basic logic for this stuff I think what we have is fine enough

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -463,6 +470,140 @@ PRIMARY KEY ({primaryKeys})
...
+            return await this.GetMetricsAsync();
+        }
+
+        public async Task<SqlTriggerMetrics> GetMetricsAsync()
+        {
+            Debug.Assert(!(this._changeMonitor is null));
+
+            return new SqlTriggerMetrics
+            {
+                UnprocessedChangeCount = await this._changeMonitor.GetUnprocessedChangeCountAsync(),
```
Charles-Gagnon, line 484: Wrap the call to get the change count in a try/catch and track error event if it fails.

JatinSanghvi: The method body of `GetUnprocessedChangeCountAsync` is entirely enclosed within a try-block. I am not sure if it will make sense to have extra logging here.

Charles-Gagnon: Ah right, ok yeah that's enough then.

Charles-Gagnon:
Hm, looks like something with my changes to the configuration overrides tests isn't stable. I'll look into it - for now just ignore any failures in those tests.

Charles-Gagnon (approved):

PR #392, opened 2022-10-12
Skip AddProductWithDefaultPKTest for Javascript

lucyzhang929:
The test is failing for JavaScript since the extension bundle does not contain the fix for default columns yet. I manually updated the sql extension dll in my local extension bundle to verify that this test passes for JavaScript. 

Created issue to re-enable: https://github.com/Azure/azure-functions-sql-extension/issues/393.

### DIFF
--- a/test/Integration/SqlOutputBindingIntegrationTests.cs
+++ b/test/Integration/SqlOutputBindingIntegrationTests.cs
@@ -384,6 +384,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         /// </summary>
         [Theory]
         [SqlInlineData()]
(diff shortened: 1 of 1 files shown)


lucyzhang929:
Maybe we can copy the dll into the extension bundle in the pipeline as well. I'll look into that.

Charles-Gagnon:
`func GetExtensionBundlePath` should be useful for that. 

Might also want to ask the AF team to see if there's a better way to do this too.

lucyzhang929:
Opened https://github.com/Azure/azure-functions-sql-extension/pull/395 to address issue.

Charles-Gagnon:
Make sure to delete the branch if you don't need this anymore.

PR #395, opened 2022-10-13
Copy latest sql dll to extension bundle in pipeline

lucyzhang929:
This PR adds a task in the build pipeline to copy the latest sql dll to the extension bundle so that the latest changes can be used for non-CSharp tests. Passing ad hoc build https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=174829&view=results

I also realized that the case sensitive test was failing after this PR: https://github.com/Azure/azure-functions-sql-extension/pull/388 (my bad for not noticing that the PR validation was not running when I merged in the PR). The test failed because we cache the table info for 10 minutes so when we change the table collation back to case insensitive, SqlAsyncCollector still thinks the collation is case sensitive.

I couldn't figure out why the cache was being missed previously (the cache was empty during the second upsert) and I didn't change the caching logic is this PR https://github.com/Azure/azure-functions-sql-extension/pull/388 so not sure what exact change caused the table caching to work.

### DIFF
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -73,14 +73,16 @@ steps:
 - task: CopyFiles@2
   displayName: 'Copy Sql extension dll to Azure Functions extension bundle for Windows'
   inputs:
-    contents: $(Build.SourcesDirectory)\src\bin\${{ parameters.configuration }}\**\Microsoft.Azure.WebJobs.Extensions.Sql.dll
+    sourceFolder: $(Build.SourcesDirectory)\src\bin\${{ parameters.configuration }}\netstandard2.0
+    contents: Microsoft.Azure.WebJobs.Extensions.Sql.dll
     targetFolder: $(azureFunctionsExtensionBundlePath)\bin
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'))
 
 - task: CopyFiles@2
   displayName: 'Copy Sql extension dll to Azure Functions extension bundle for Linux'
   inputs:
-    contents: $(Build.SourcesDirectory)/src/bin/${{ parameters.configuration }}/**/Microsoft.Azure.WebJobs.Extensions.Sql.dll
+    sourceFolder: $(Build.SourcesDirectory)/src/bin/${{ parameters.configuration }}/netstandard2.0
+    contents: Microsoft.Azure.WebJobs.Extensions.Sql.dll
     targetFolder: $(azureFunctionsExtensionBundlePath)/bin
   condition: and(succeeded(), eq(variables['Agent.OS'], 'linux'))
 
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -76,7 +76,7 @@ steps:
   inputs:
     sourceFolder: $(Build.SourcesDirectory)/src/bin/${{ parameters.configuration }}/netstandard2.0
     contents: Microsoft.Azure.WebJobs.Extensions.Sql.dll
-    targetFolder: $(azureFunctionsExtensionBundlePath)\bin
+    targetFolder: $(azureFunctionsExtensionBundlePath)/bin
     overWrite: true
 
 - script: |
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -247,7 +247,6 @@ steps:
     TEST_SERVER: '$(testServer)'
     NODE_MODULES_PATH: '$(nodeModulesPath)'
     AZUREFUNCTIONS_SQLBINDINGS_TELEMETRY_OPTOUT: '1'
-    AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN: '0'
   inputs:
     command: test
     projects: '${{ parameters.solution }}'
@@ -259,7 +258,6 @@ steps:
   env:
     SA_PASSWORD: '$(serverPassword)'
     AZUREFUNCTIONS_SQLBINDINGS_TELEMETRY_OPTOUT: '1'
-    AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN: '0'
   inputs:
     command: test
     projects: '${{ parameters.solution }}'
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -73,8 +73,9 @@ steps:
 - task: CopyFiles@2
   displayName: 'Copy Sql extension dll to Azure Functions extension bundle'
   inputs:
-    contents: '$(Build.SourcesDirectory)\src\bin\${{ parameters.configuration }}\Microsoft.Azure.WebJobs.Extensions.Sql.dll'
-    targetFolder: '$(azureFunctionsExtensionBundlePath)'
+    contents: $(Build.SourcesDirectory)\src\bin\${{ parameters.configuration }}\Microsoft.Azure.WebJobs.Extensions.Sql.dll
+    targetFolder: $(azureFunctionsExtensionBundlePath)
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'))
 
 - script: |
     npm install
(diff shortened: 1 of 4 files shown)


test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -369,7 +369,9 @@ public void AddProductCaseSensitiveTest(SupportedLanguages lang)
             Assert.Throws<AggregateException>(() => this.SendOutputGetRequest("addproduct-params", query).Wait());
 
             // Change database collation back to case insensitive
-            this.ExecuteNonQuery($"ALTER DATABASE {this.DatabaseName} COLLATE Latin1_General_CI_AS");
+            this.ExecuteNonQuery($"ALTER DATABASE {this.DatabaseName} SET Single_User WITH ROLLBACK IMMEDIATE; ALTER DATABASE {this.DatabaseName} COLLATE Latin1_General_CI_AS; ALTER DATABASE {this.DatabaseName} SET Multi_User;");
+
+            Thread.Sleep(660000); // Wait 11 minutes to ensure table info cache is not hit since database collation was changed
```
lucyzhang929, line 374: This increases build time by a lot (especially once more languages are added). Another option is to remove the second get request since the main objective of this test is to test the case sensitive scenario.

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -355,7 +355,7 @@ public void AddProductCaseSensitiveTest(SupportedLanguages lang)
             this.StartFunctionHost(nameof(AddProductParams), lang);
 
             // Change database collation to case sensitive
-            this.ExecuteNonQuery($"ALTER DATABASE {this.DatabaseName} COLLATE Latin1_General_CS_AS");
+            this.ExecuteNonQuery($"ALTER DATABASE {this.DatabaseName} SET Single_User WITH ROLLBACK IMMEDIATE; ALTER DATABASE {this.DatabaseName} COLLATE Latin1_General_CS_AS; ALTER DATABASE {this.DatabaseName} SET Multi_User;");
```
lucyzhang929, line 358: Modified this because I was getting the following error when running this test locally:
`The database could not be exclusively locked to perform the operation`

test/Integration/SqlOutputBindingIntegrationTests.cs
Charles-Gagnon: Instead of this we should add a way for tests to bypass/reduce this cache. I would suggest we add an environment variable for that - something like `AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MS` (so we can set 0 to not have caching at all)

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -67,6 +71,24 @@ steps:
     projects: '${{ parameters.solution }}'
     arguments: '--configuration ${{ parameters.configuration }} -p:GeneratePackageOnBuild=false -p:Version=${{ parameters.binariesVersion }}'
 
+- task: CopyFiles@2
+  displayName: 'Copy Sql extension dll to Azure Functions extension bundle for Windows'
+  inputs:
+    sourceFolder: $(Build.SourcesDirectory)\src\bin\${{ parameters.configuration }}\netstandard2.0
```
Charles-Gagnon, line 77: I didn't think these tasks cared about path separators - they were able to take in either. Is that not actually the case?

lucyzhang929: Oh I thought I saw somewhere that they do but I can verify.

src/SqlAsyncCollector.cs
```diff
@@ -171,6 +172,16 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                 ObjectCache cachedTables = MemoryCache.Default;
                 var tableInfo = cachedTables[cacheKey] as TableInformation;
 
+                int timeout = 10;
+                string timeoutEnvVar = Environment.GetEnvironmentVariable("AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN");
+                this._logger.LogDebugWithThreadId("TIMEOUT STRING HERE " + timeoutEnvVar);
```
Charles-Gagnon, line 177: I'd remove this and then add a log message along the lines of "Overriding default table info cache timeout with new value {0}`

src/SqlAsyncCollector.cs
```diff
@@ -171,6 +172,16 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
...
+
+                if (!string.IsNullOrEmpty(timeoutEnvVar))
+                {
+                    timeout = int.Parse(timeoutEnvVar, CultureInfo.InvariantCulture);
```
Charles-Gagnon, line 181: TryParse - we don't want this to fail if the variable isn't set correctly. Just fall back to the default in that case.

src/SqlAsyncCollector.cs
```diff
@@ -179,7 +190,7 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                     var policy = new CacheItemPolicy
                     {
                         // Re-look up the primary key(s) after 10 minutes (they should not change very often!)
```
Charles-Gagnon, line 192: Update comment

src/SqlAsyncCollector.cs
```diff
@@ -171,6 +172,16 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                 ObjectCache cachedTables = MemoryCache.Default;
                 var tableInfo = cachedTables[cacheKey] as TableInformation;
 
+                int timeout = 10;
```
Charles-Gagnon, line 175: Make a `DEFAULT_TABLE_INFO_CACHE_TIMEOUT_MIN` var and set this to that initially.

src/SqlAsyncCollector.cs
lucyzhang929: Oops that was just for my debugging but yes I will update it.

builds/azure-pipelines/template-steps-build-test.yml
lucyzhang929: Verified that we can just use one task.

src/SqlAsyncCollector.cs
```diff
@@ -171,15 +174,25 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                 ObjectCache cachedTables = MemoryCache.Default;
                 var tableInfo = cachedTables[cacheKey] as TableInformation;
 
+                int timeout = DEFAULT_TABLE_INFO_CACHE_TIMEOUT_MIN;
+                string timeoutEnvVar = Environment.GetEnvironmentVariable("AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN");
+                if (!string.IsNullOrEmpty(timeoutEnvVar))
+                {
+                    if (int.TryParse(timeoutEnvVar, NumberStyles.Integer, CultureInfo.InvariantCulture, out timeout))
```
Charles-Gagnon, line 181: If TryParse fails it will set the result to 0 

https://learn.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-6.0#system-int32-tryparse(system-string-system-int32@)

So you need to reset it to the default in that case.

src/SqlAsyncCollector.cs
```diff
@@ -171,15 +174,25 @@ private async Task UpsertRowsAsync(IEnumerable<T> rows, SqlAttribute attribute,
                 ObjectCache cachedTables = MemoryCache.Default;
                 var tableInfo = cachedTables[cacheKey] as TableInformation;
 
+                int timeout = DEFAULT_TABLE_INFO_CACHE_TIMEOUT_MIN;
+                string timeoutEnvVar = Environment.GetEnvironmentVariable("AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN");
```
Charles-Gagnon, line 178: Let's name this `AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MINUTES` so it isn't confused with minimum

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -235,6 +247,7 @@ steps:
     TEST_SERVER: '$(testServer)'
     NODE_MODULES_PATH: '$(nodeModulesPath)'
     AZUREFUNCTIONS_SQLBINDINGS_TELEMETRY_OPTOUT: '1'
+    AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MIN: '0'
```
Charles-Gagnon, line 250: Let's not set this at the top level here so that other tests use the normal timeout. Instead pass it in the start function host call just for that specific test.

Charles-Gagnon (approved):

PR #401, opened 2022-10-14
Fix polling size override using wrong configuration value

Charles-Gagnon:
This was broken in https://github.com/Azure/azure-functions-sql-extension/pull/376 (copypaste typo)

It wasn't caught by the test because the default values were changed but the tests hadn't been updated with the new values. I fixed them to use the default values defined in the trigger code directly so this doesn't happen again if those are changed in the future. 

I also added another check to watch for the log message we send to indicate what the batchSize/pollingInterval values are as a secondary verification.

### DIFF
--- a/test/Common/TestUtils.cs
+++ b/test/Common/TestUtils.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -75,14 +75,30 @@ public async Task SingleOperationTriggerTest()
         [Fact]
         public async Task BatchSizeOverrideTriggerTest()
         {
-            const int batchSize = 20;
+            // Use enough items for the default batch size to require 4 batches but then
```
MaddyDev, line 78: "// Use enough items to require 4 batches to be processed" is simpler to understand.

test/Common/TestUtils.cs
```diff
@@ -177,5 +178,35 @@ public static async Task<TResult> TimeoutAfter<TResult>(this Task<TResult> task,
...
+
+        /// <summary>
+        /// Creates a DataReceievedEventHandler that will wait for the specified regex and then check that
+        /// the matched group matches the expected value.
+        /// </summary>
+        /// <param name="taskCompletionSource">The task completion source to signal when the value is receieved</param>
+        /// <param name="regex">The regex. This must have a single group match for the specific value being looked for</param>
+        /// <param name="valueName">The name of the value to output if the match fails</param>
+        /// <param name="expectedValue">The value expected to be equal to the matched group from the regex</param>
+        /// <returns></returns>
```
MaddyDev, line 190: returns the event handler

MaddyDev (approved):

test/Common/TestUtils.cs
```diff
@@ -177,5 +178,35 @@ public static async Task<TResult> TimeoutAfter<TResult>(this Task<TResult> task,
                 throw new TimeoutException(message);
             }
         }
+
+        /// <summary>
+        /// Creates a DataReceievedEventHandler that will wait for the specified regex and then check that
+        /// the matched group matches the expected value.
+        /// </summary>
+        /// <param name="taskCompletionSource">The task completion source to signal when the value is receieved</param>
```
lucyzhang929, line 186: typo: received

lucyzhang929 (approved):

test/Common/TestUtils.cs
Charles-Gagnon: This is seriously one of my least favorite words - my fingers just always want to spell it recieved.

chlafreniere (approved):

PR #405, opened 2022-10-18
Use merge query for renewing leases

Charles-Gagnon:
While I was investigating timeouts that were happening during the BatchSizeOverride tests I noticed that the renew leases query seemed to be the main cause of the issue - because of recent changes to the tests we were greatly increasing the number of items being processed at once which meant that the query being generated was a huge number (400 in this case) of INSERT/UPDATE statements that were all being executed sequentially with their own table locks. 

When there's only 10 items this isn't too bad - but with 400 items I started seeing it take >1sec to process all those statements. This then caused the test to fail because it never got the trigger events it was waiting on in the time it expected.

To fix this I'm switching to use a merge statement like we do elsewhere so that we can update all the rows in a single statement. It takes in the serialized row data (containing the list of rows to renew leases for) and creates a table from that and then merges that in with the leases table, either creating or updating the values as needed.

Now I'm seeing <50ms updates for 400 rows.

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -745,15 +745,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 // These are the internal column values that we use. Note that we use SYS_CHANGE_VERSION because that's
                 // the new version - the _az_func_ChangeVersion has the old version 
                 .Concat(new string[] { "SYS_CHANGE_VERSION bigint", "_az_func_AttemptCount int" });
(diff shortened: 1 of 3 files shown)


JatinSanghvi:
The changes look good to me. Can you ask someone who understands usage of MERGE INTO statement to do the PR review?

Charles-Gagnon:
I'll see who I can find - I don't know anyone off hand. I'm going to plan on merging this in in the meantime so we can be testing it ourselves and with the automated tests and then if I can find someone have them take a look at it later.

dzsquared (approved):

Issue #410, opened 2022-10-19
[Java] Input binding for tables with datetime type columns throws an error


lucyzhang929:
The java worker throws the following error when converting the datetime json value to a Date object because it expects a time zone and the sql datetime type does not include time zones.
```
Exception: IllegalArgumentException: No time zone indicator
[2022-10-18T20:55:01.008Z] Stack: com.google.gson.JsonSyntaxException: Failed parsing '2022-10-17T14:39:51' as Date; at path $[0].Datetime
```

PR #412, opened 2022-10-20
Serialize datetime objects in utc format for input binding

lucyzhang929:
This PR fixes https://github.com/Azure/azure-functions-sql-extension/issues/410.


src/SqlConverters.cs
```diff
@@ -186,7 +186,12 @@ public virtual async Task<string> BuildItemFromAttributeAsync(SqlAttribute attri
                     var dataTable = new DataTable();
                     adapter.Fill(dataTable);
                     this._logger.LogInformation($"{dataTable.Rows.Count} row(s) queried from database: {connection.Database} using Command: {command.CommandText}");
-                    return JsonConvert.SerializeObject(dataTable);
+                    // Serialize any DateTime objects in UTC format
+                    var jsonSerializerSettings = new JsonSerializerSettings()
+                    {
+                        DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ"
```
Charles-Gagnon, line 192: So we do a slightly different format in the AsyncCollector - https://github.com/Azure/azure-functions-sql-extension/blob/b966b42f3d74d872d5c5b63b40f35daf97669b8c/src/SqlAsyncCollector.cs#L380

But it looks like both that and with the Z are supported formats : 

```
The ISO 8601 date with time format is as follows:

YYYY-MM-DDThh:mm:ss[.nnnnnnn][{+|-}hh:mm]

YYYY-MM-DDThh:mm:ss[.nnnnnnn]Z (UTC, Coordinated Universal Time)
```

We definitely need to be consistent across the entire codebase with how we treat datetimes so I would suggest pulling that constant out of that class and putting it in a general constants file, then using that both there and here. 

In addition, please make a follow up issue to investigate all serialization/deserialization to ensure we're doing it correctly (including in the triggerbinding branch).

Charles-Gagnon (changes requested):
Please add a test to verify dates are being deserialized correctly (or modify the existing test we have to verify that the deserialized format is correct)

src/SqlConverters.cs
lucyzhang929: I will create a constants file and include Z in the format since gson expects the time zone. 
Created an issue to track here: https://github.com/Azure/azure-functions-sql-extension/issues/413

test/Integration/test-csharp/GetProductColumnTypesSerializationAsyncEnumerable.cs
```diff
@@ -13,30 +12,29 @@
...
-            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "getproducts-columntypesserializationdifferentculture")]
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "getproducts-columntypesserializationasyncenumerable")]
             HttpRequest req,
             [Sql("SELECT * FROM [dbo].[ProductsColumnTypes]",
                 CommandType = System.Data.CommandType.Text,
                 ConnectionStringSetting = "SqlConnectionString")]
             IAsyncEnumerable<ProductColumnTypes> products,
             ILogger log)
         {
-            CultureInfo.CurrentCulture = new CultureInfo("it-IT", false);
```
lucyzhang929, line 34: @Charles-Gagnon When I checked the logs, the CurrentCulture didn't affect how the datetime was serialized (I believe it only affects strings). I removed the CurrentCulture and am using this sample to test IAsyncEnumerable and the GetProductsColumnTypes sample to test IEnumerable. Please let me know if there is something I am missing here.

Charles-Gagnon: So you're correct in that it doesn't affect how it's serialized - the issue that we ran into was that we were calling ToString on the Date objects in our extension code - which does adhere to the CurrentCulture. Then when that was converted back into a DateTime for the POCO object it failed at that point with this error 

`[2022-08-25T07:31:54.100Z] System.Private.CoreLib: Exception while executing function: Function1. Newtonsoft.Json: Could not convert string to DateTime: 25/08/2022 07:19:20. Path 'LastExecutionDate', line 1, position 51.`

The issue was fixed so removing the CurrentCulture being set here isn't going to change anything (since I removed the ToString completely) - but it means that if we were to revert that fix we wouldn't catch it.

Basically this test is just verifying that changing the culture of the function doesn't break any of the serialization/deserialization stuff. 

What I'd suggest is having the function take in an HTTP parameter that tells it what language to use (with no value no changing anything). Then you should make multiple versions of the test using this that specify other languages to use (for now just one we know fails like it-IT is fine in addition to default en-us)

lucyzhang929: Added back the culture changes - thanks for the explanation!

test/Integration/test-csharp/GetProductColumnTypesSerializationAsyncEnumerable.cs
```diff
@@ -13,30 +13,35 @@
...
+            string language = req.Query["culture"];
+            if (string.IsNullOrEmpty(language))
```
Charles-Gagnon, line 33: `!string.IsNullOrEmpty`

And add an explanation comment here about why we do this (basically summarize what I put before)

test/Integration/test-csharp/GetProductColumnTypesSerialization.cs
```diff
@@ -27,13 +23,8 @@ public static class GetProductsColumnTypesSerialization
             [Sql("SELECT * FROM [dbo].[ProductsColumnTypes]",
                 CommandType = System.Data.CommandType.Text,
                 ConnectionStringSetting = "SqlConnectionString")]
-            IAsyncEnumerable<ProductColumnTypes> products,
-            ILogger log)
+            IEnumerable<ProductColumnTypes> products)
```
Charles-Gagnon, line 26: We should be consistent and call Serialize on the items in the IEnumerable as well

Charles-Gagnon (approved):

PR #425, opened 2022-10-26
Add pipeline for building Java library

lucyzhang929:


builds/azure-pipelines/build-release-java.yml
```diff
@@ -0,0 +1,55 @@
+trigger: none
+
+pr: none
+
+variables:
+  versionMajor: 0 # TODO When this is bumped have versionPatch counter use versionMajorMinor instead https://github.com/Azure/azure-functions-sql-extension/issues/173
```
MaddyDev, line 6: Charles updated the major version after bumping the sql client version in the other PR, might want to update here as well?

Charles-Gagnon: We shouldn't be having duplicate stuff like that - especially for something as important as a version.

Instead you should make a top level build-release pipeline (that defines the version) with separate stages for each of the different things we build. See ADS with the mac/windows/linux pipelines we have for an example.

lucyzhang929: Since the java library is separate from the c# extension and likely won't be needing updates very often, I plan on keeping the version separate from the nuget package. Are there concerns with this @Charles-Gagnon @MaddyDev ?

Charles-Gagnon: Ok yeah that makes sense given that these are completely separate from the C# extension core. 

I even wonder if we want to auto-increment the version for this one...maybe for now we can skip doing that and just hardcode the version given that I agree it seems extremely unlikely that we'll be updating this much.

builds/azure-pipelines/build-release-java.yml
```diff
@@ -0,0 +1,51 @@
+trigger: none
+
+pr: none
+
+variables:
+  version: '0.1.0'
```
Charles-Gagnon, line 6: I'd suggest just putting this directly into the pom.xml - that's more expected for where to find the current version. (and then remove the set library version step)

lucyzhang929:
@Charles-Gagnon Do you mean add the java build step to the build-release.yml instead of having a separate pipeline?

Charles-Gagnon:
Having the separate pipeline file is fine - just add it as a stage in the overall build/release pipeline (like what we do in https://github.com/microsoft/azuredatastudio/blob/main/build/azure-pipelines/sql-product-build.yml#L7)

lucyzhang929:
Does that mean the java library will be built every day as well? Do we want to do that if it won't be changed often? Right now I have it as a separate pipeline that is only triggered manually. https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build?definitionId=589

Charles-Gagnon:
Yes, because we want to make sure that everything builds successfully every day. You never know when some other change in the repo might affect the build and we want to catch that stuff as early as we can.

builds/azure-pipelines/build-release-java.yml
```diff
@@ -0,0 +1,30 @@
+steps:
+  - task: Maven@3
+    displayName: Build library
+    inputs:
+      mavenPomFile: java-library/pom.xml
+      options: --batch-mode --define=gpg.skip --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn --update-snapshots
+
+  - powershell: |
+      $prefix = 'azure-functions-java-library-sql-0.1.0'
```
Charles-Gagnon, line 9: Instead of hardcoding the version here what about just using the copy task with star patterns?

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/copy-files?view=azure-devops&tabs=yaml

(and I don't think you need to make a version-named target folder for this. Just name it something generic like java-library or something)

builds/azure-pipelines/build-release-java.yml
```diff
@@ -0,0 +1,30 @@
...
+      Copy-Item "$source/target/$prefix-javadoc.jar" "$destination/$prefix-javadoc.jar"
+      Copy-Item "$source/target/$prefix-sources.jar" "$destination/$prefix-sources.jar"
+    displayName: Copy output files
+
+  - task: ManifestGeneratorTask@0
+    displayName: Generate SBOM manifest
+    inputs:
+      buildDropPath: $(Build.ArtifactStagingDirectory)/0.1.0
+      packageName: Azure Functions SQL Java Bindings
+      packageVersion: '0.1.0'
```
Charles-Gagnon, line 24: Do we actually need to specify the version? I don't see that in ADS

lucyzhang929: I think it's not required - I'll remove it

Charles-Gagnon (approved):

PR #414, opened 2022-10-21
Add unit tests for runtime driven scaling

JatinSanghvi:

PR #428, opened 2022-10-28
Add trigger scaling configuration

### LINKED ISSUE #419: Trigger scaling configuration
Currently trigger scaling is based off of a hardcoded items/worker (1000). This should be a user-configurable option similar to the throughput options.

AmeyaRele:
Closes #419

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ This controls the delay in milliseconds between processing each batch of changes
 
 #### Sql_Trigger_MaxChangesPerWorker
 
-This controls the upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#Scaling) section for more information.
+This controls the upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#scaling) section for more information.
 
 ## More Samples
 
--- a/README.md
+++ b/README.md
@@ -578,6 +578,10 @@ This controls the number of changes processed at once before being sent to the t
 
 This controls the delay in milliseconds between processing each batch of changes.
 
+#### Sql_Trigger_MaxChangesPerWorker
+
+This controls the max number of changes handled by a worker, used for scaling decision.
+
 ## More Samples
 
 ### Input Binding
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Azure SQL bindings for Azure Functions are supported for:
     - [Trigger Binding Configuration](#trigger-binding-configuration)
       - [Sql_Trigger_BatchSize](#sql_trigger_batchsize)
       - [Sql_Trigger_PollingIntervalMs](#sql_trigger_pollingintervalms)
+      - [Sql_Trigger_MaxChangesPerWorker](#sql_trigger_maxchangesperworker)
   - [More Samples](#more-samples)
     - [Input Binding](#input-binding)
       - [Query String](#query-string)
@@ -580,7 +581,7 @@ This controls the delay in milliseconds between processing each batch of changes
 
 #### Sql_Trigger_MaxChangesPerWorker
 
-This controls the max number of changes handled by a worker, used for scaling decision.
+This controls the upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#Scaling) section for more information.
 
 ## More Samples
 
--- a/src/TriggerBinding/SqlTriggerListener.cs
+++ b/src/TriggerBinding/SqlTriggerListener.cs
@@ -31,6 +31,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private const int ListenerStarted = 2;
         private const int ListenerStopping = 3;
         private const int ListenerStopped = 4;
+        public const int DefaultMaxChangesPerWorker = 1000;
 
         private readonly SqlObject _userTable;
         private readonly string _connectionString;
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -578,6 +578,10 @@ This controls the number of changes processed at once before being sent to the t
 
 This controls the delay in milliseconds between processing each batch of changes.
 
+#### Sql_Trigger_MaxChangesPerWorker
+
+This controls the max number of changes handled by a worker, used for scaling decision.
```
JatinSanghvi, line 583: Maybe we should expand on this one. Adding @Charles-Gagnon.

'This controls the upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#scaling) section of more information.'

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -41,6 +42,7 @@ internal sealed class SqlTriggerListener<T> : IListener, IScaleMonitor<SqlTrigge
         private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
 
         private readonly IDictionary<TelemetryPropertyName, string> _telemetryProps = new Dictionary<TelemetryPropertyName, string>();
+        private readonly int _maxChangesPerWorker = DefaultMaxChangesPerWorker;
```
JatinSanghvi, line 45: Better not have two places for initialization. Keep `_maxChangesPerWorker` uninitialized here and inside the constructor, change to `this._maxChangesPerWorker = configuredMaxChangesPerWorker ?? DefaultMaxChangesPerWorker;`.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -31,6 +31,7 @@ internal sealed class SqlTriggerListener<T> : IListener, IScaleMonitor<SqlTrigge
         private const int ListenerStarted = 2;
         private const int ListenerStopping = 3;
         private const int ListenerStopped = 4;
+        public const int DefaultMaxChangesPerWorker = 1000;
```
JatinSanghvi, line 34: I had added the below code comment to address review suggestion. Can you add it back to top of `DefaultMaxChangesPerWorker` definition?

```
            // NOTE: please ensure the Readme file and other public documentation are also updated if this value ever
            // needs to be changed.
```

README.md
Charles-Gagnon: That updated text looks great to me!

(except it should be `See the Scaling section for more information`

Charles-Gagnon (commented):
Please unit tests similar to ones added in https://github.com/Azure/azure-functions-sql-extension/pull/414

(and fix the broken tests)

README.md
```diff
@@ -578,6 +579,10 @@ This controls the number of changes processed at once before being sent to the t
 
 This controls the delay in milliseconds between processing each batch of changes.
 
+#### Sql_Trigger_MaxChangesPerWorker
+
+This controls the upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#Scaling) section for more information.
```
JatinSanghvi, line 584: Change link to lowercase i.e., `#scaling`. If you go to https://github.com/Azure/azure-functions-sql-extension/tree/ameyarele/scalingConfiguration#change-tracking and click on the 'link' icon besides 'Scaling' heading, the URL will have the lowercase fragment name. A different case will work for compatibility reasons, but the convention for link generation is to convert the section heading to lower case and replace the spaces with hyphens.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -64,10 +69,21 @@ public SqlTriggerListener(string connectionString, string tableName, string user
             this._executor = executor ?? throw new ArgumentNullException(nameof(executor));
             this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
             this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
-
+            int? configuredMaxChangesPerWorker;
             // Do not convert the scale-monitor ID to lower-case string since SQL table names can be case-sensitive
             // depending on the collation of the current database.
             this._scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{userFunctionId}-SqlTrigger-{tableName}");
+
+            //In case converting from string to int is not possible from the user input.
```
JatinSanghvi, line 77: nit: Put space after `//`.

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -64,10 +69,21 @@ public SqlTriggerListener(string connectionString, string tableName, string user
...
+            try
+            {
+                configuredMaxChangesPerWorker = configuration.GetValue<int?>(SqlTriggerConstants.ConfigKey_SqlTrigger_MaxChangesPerWorker);
```
JatinSanghvi, line 80: Can you use `configuration.GetValue<int>` instead? If the configuration is absent, you should get default of int i.e., `0` as result. That will simplify the check in line 86. You will still require try-catch block for non-integer configuration values.

test/Unit/TriggerBinding/SqlTriggerListenerTests.cs
```diff
@@ -201,24 +201,55 @@ public void ScaleMonitorGetScaleStatus_CountNotIncreasingOrDecreasing_ReturnsNon
...
+        [InlineData("10")]
+        [InlineData("10000")]
+        public void ScaleMonitorGetScaleStatus_VerifyUserConfiguredTriggerConfig(string maxChangesPerWorker)
+        {
+            (IScaleMonitor<SqlTriggerMetrics> monitor, List<string> logMessages) = GetScaleMonitor(maxChangesPerWorker);
+            ScaleStatusContext context = GetScaleStatusContext(new int[] { 100, 150, 130, 120, 90 }, 10);
+
+            monitor.GetScaleStatus(context);
+
+            Assert.Contains(logMessages, message => message.Contains($"maximum changes per worker: {maxChangesPerWorker}"));
```
JatinSanghvi, line 215: We should check beyond what is being logged. Please do not edit the tests for now. I will try pushing a commit for tests by EOD.

JatinSanghvi (changes requested):

JatinSanghvi (approved):

Charles-Gagnon (approved):

PR #429, opened 2022-10-28
Merge main -> triggerbindings

Charles-Gagnon:
Primarily this is bringing in the SqlClient 5.0 update.

Note that I had to downgrade the Microsoft.NET.Sdk.Functions package from 4.1.3 -> 4.1.1 since starting with 4.1.2 it requires Newtonsoft >= 13.0.1 so isn't compatible with this branch with is still using Newtonsoft 11


Charles-Gagnon:
I'm going to leave this open until we can discuss with the LA team about the potential side effects of this. Just wanted to get it out  for people to look at (and use for testing) ahead of time.

Charles-Gagnon:
Merging this in now since we'll be going out in public preview with it. Will be following up with Logic Apps team to ensure that there aren't any issues on their end using this.

Issue #438, opened 2022-11-02
Enable debug logging during test runs


Charles-Gagnon:
Currently test runs have the normal default logging level, but this doesn't have much information. It would be helpful (especially for investigating lab run failures) to enable debug logging since that is a lot more useful.

The downside is that this adds a not-insignificant amount of overhead having to log all these extra messages, so enabling this currently causes timeouts in some tests since there's so much extra time spent writing these logging messages.

PR #439, opened 2022-11-02
Perf test fixes/improvements

Charles-Gagnon:
1. Add editorconfig with same stuff we ignore in test projects since we don't need to be as strict here
2. Add MemoryDiagnoser to trigger test
3. Move common stuff to base class - I'll be adding more trigger perf tests here soon which will reuse those
4. Disable/Re-enable change tracking on the table after/before each test so that we start with a clean slate each time 
5. Delete Leases table after each run, otherwise we run into an issue where the functions are killed before they can release their leases and so end up skipping over the rows since the IDs are reused across multiple runs. This is kind of messy and would be worth revisiting in the future 
6. Adding logging statement to tests when they start waiting for product changes to come through, this can help figure out timeout issues since it tells you when it's actually started watching so you can see what it was doing before it timed out
7. Add README section on troubleshooting tests on enabling debug logging. Made https://github.com/Azure/azure-functions-sql-extension/issues/438 for tracking enabling this for all test runs

### DIFF
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -394,18 +394,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
                 this.LogOutput($"Failed to close connection. Error: {e1.Message}");
             }
 
-            foreach (Process functionHost in this.FunctionHostList)
-            {
-                try
-                {
-                    functionHost.Kill();
-                    functionHost.Dispose();
-                }
-                catch (Exception e2)
-                {
-                    this.LogOutput($"Failed to stop function host, Error: {e2.Message}");
-                }
-            }
+            this.DisposeFunctionHosts();
 
             try
             {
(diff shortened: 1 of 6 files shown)


test/Integration/IntegrationTestBase.cs
```diff
@@ -432,6 +421,26 @@ public void Dispose()
...
+        protected void DisposeFunctionHosts()
+        {
+            foreach (Process functionHost in this.FunctionHostList)
+            {
+                try
+                {
+                    functionHost.Kill();
+                    functionHost.Dispose();
+                }
+                catch (Exception e2)
```
chlafreniere, line 436: nit: does this need to be `e2`?

chlafreniere (approved):

test/Integration/IntegrationTestBase.cs
```diff
@@ -432,6 +421,26 @@ public void Dispose()
...
+                {
+                    this.LogOutput($"Failed to stop function host, Error: {e2.Message}");
```
Charles-Gagnon, line 438: @chlafreniere Great example where using the suggestion feature is useful 😄 
```suggestion
                catch (Exception ex)
                {
                    this.LogOutput($"Failed to stop function host, Error: {ex.Message}");
```

PR #441, opened 2022-11-02
samples for out of proc sql bindings

MaddyDev:
Samples for out of proc functions. I have replicated all applicable samples that are in samples-csharp. 
Excluded the following two samples examples: 1. AddProductsAsyncCollector, 2. AddProductsCollector since .NET isolated projects run in a separate worker process, bindings can't take advantage of rich binding classes, such as ICollector<T>, IAsyncCollector<T>, and CloudBlockBlob. (as stated in https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#output-bindings)


samples/samples-outofproc/.editorconfig
```diff
@@ -0,0 +1,6 @@
+# See https://github.com/dotnet/roslyn-analyzers/blob/main/.editorconfig for an example on different settings and how they're used
+
+[*.cs]
+# Disable these until they can be fixed
```
Charles-Gagnon, line 4: Why can't you fix this one? Unnecessary imports should be easy to fix

samples/samples-outofproc/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
+## files generated by popular Visual Studio add-ons.
+
+# Azure Functions localsettings file
+local.settings.json
+
+# User-specific files
+*.suo
```
Charles-Gagnon, line 8: Do you need this? Most of these are duplicated in the top level gitignore already. If there's ones that are relevant to just this folder then you can keep them, but delete the rest.

samples/samples-outofproc/OutputBindingSamples/AddProductParams.cs
```diff
@@ -0,0 +1,33 @@
...
+        [SqlOutput("dbo.Products", ConnectionStringSetting = "SqlConnectionString")]
+        public static Product Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "addproduct-params")]
+            HttpRequestData req)
+        {
+            if (req != null)
+            {
+                var product = new Product()
+                {
+                    Name = HttpUtility.ParseQueryString(req.Url.Query)["name"],
```
Charles-Gagnon, line 24: Only parse the query string once and save the result

Charles-Gagnon (commented):
Looks like the lock files aren't fully updated, make sure you run `dotnet restore --force-evaluate` from the root directory.

samples/samples-outofproc/InputBindingSamples/GetProductNamesView.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// This shows an example of a SQL Input binding that queries from a SQL View named ProductNames.
+    /// </summary>
+    public static class GetProductNamesView
+    {
+        [Function("GetProductNamesView")]
+        public static IActionResult Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "getproduct-namesview/")]
+            HttpRequest req,
+            [SqlInput("SELECT DISTINCT Name FROM Products",
```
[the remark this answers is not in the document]
MaddyDev: I was going to clarify with you and Vasu on that, skipped my mind! Couldn't find ProductNames table and am not sure why ProductNames would return all products like we're querying from Products, the name function name is confusing.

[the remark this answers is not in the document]
MaddyDev: I haven't created the view before, it works:
image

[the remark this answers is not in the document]
MaddyDev: I haven't created all the scripts on my local test db and was not able to run it before. I had a discussion with Vasu regarding this test and we didn't realize that its querying a view. I just created the view and tested with the above results. 
Yes I've created all the tests same except this one, I was meaning to check with Lucy/Vasu as I stated above.

.vscode/settings.json
```diff
@@ -6,5 +6,6 @@
     "debug.internalConsoleOptions": "neverOpen",
     "azureFunctions.preDeployTask": "publish (functions)",
     "omnisharp.enableEditorConfigSupport": true,
-    "omnisharp.enableRoslynAnalyzers": true
+    "omnisharp.enableRoslynAnalyzers": true,
+    "java.configuration.updateBuildConfiguration": "interactive"
```
chlafreniere, line 10: why are java changes showing up?

MaddyDev: I merged main, not sure why these are showing up, checking

Charles-Gagnon: Remove this change too please

Charles-Gagnon (approved):

lucyzhang929:
Why was the mvc package removed?

MaddyDev:
> Why was the mvc package removed?

Didn't want to add another package dependency for the samples being exposed to users when we can avoid them. (Simplest way to implement them without additional stuff is the key)

lucyzhang929 (approved):

PR #453, opened 2022-11-04
Add more perf benchmarks for Trigger

Charles-Gagnon:
Getting in the latest tests I've been working on. I will likely be coming back and cleaning these up more later (for example combining the override tests all into one file) but for now splitting them out like this makes it easier/faster to run since the tests can take a long time with having to spin up a new host each iteration.


Charles-Gagnon:
The test failure is from the multi-host test, which is likely due to the deadlock issue I'm fixing in the other PR. Given that this PR isn't touching any of the existing tests I'm ignoring the failure for now and will keep an eye on that once we have the deadlock fix in.

chlafreniere (approved):

PR #454, opened 2022-11-04
Deserialize objects to compare integration tests

VasuBhog:
Found issues when I ran tests in PowerShell specifically comparing strings together was not working as we should not expect the returned string to be in order for instance. So we can compare objects instead.

### DIFF
--- a/test/Integration/SqlInputBindingIntegrationTests.cs
+++ b/test/Integration/SqlInputBindingIntegrationTests.cs
@@ -35,10 +35,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
             HttpResponseMessage response = await this.SendInputRequest("getproducts", cost.ToString());
 
             // Verify result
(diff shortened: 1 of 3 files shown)


chlafreniere (approved):
Looks fine to me, with the caveat that I haven't been super involved in the weeds around what's necessary on the PS side.

test/Integration/SqlInputBindingIntegrationTests.cs
```diff
@@ -35,10 +35,11 @@ public async void GetProductsTest(int n, int cost, SupportedLanguages lang)
             HttpResponseMessage response = await this.SendInputRequest("getproducts", cost.ToString());
 
             // Verify result
-            string expectedResponse = JsonConvert.SerializeObject(products);
+            Product[] expectedResponse = products;
```
Charles-Gagnon, line 38: Don't really need this extra var - can just use products below directly (same for similar situations below)

Charles-Gagnon (approved):

MaddyDev (approved):

Issue #406, opened 2022-10-18
CREATE SCHEMA query can fail with existing object error


Charles-Gagnon:
Saw this in a run of the MultiFunctionTriggerTest test

```
[6572] [2022-10-17T23:19:25.550Z] Failed to start SQL trigger listener for table: 'dbo.Products', function ID: '57bfa3d6e7fc45ea'. Exception: Microsoft.Data.SqlClient.SqlException (0x80131904): There is already an object named 'az_func' in the database.
[6572] [2022-10-17T23:19:25.550Z] CREATE SCHEMA failed due to previous errors.
[6572] [2022-10-17T23:19:25.550Z]    at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
[6572] [2022-10-17T23:19:25.550Z]    at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
[6572] [2022-10-17T23:19:25.550Z]    at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
[6572] [2022-10-17T23:19:25.551Z]    at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
[6572] [2022-10-17T23:19:25.551Z]    at Microsoft.Data.SqlClient.SqlCommand.InternalEndExecuteNonQuery(IAsyncResult asyncResult, Boolean isInternal, String endMethod)
[6572] [2022-10-17T23:19:25.551Z]    at Microsoft.Data.SqlClient.SqlCommand.EndExecuteNonQueryInternal(IAsyncResult asyncResult)
[6572] [2022-10-17T23:19:25.551Z]    at Microsoft.Data.SqlClient.SqlCommand.EndExecuteNonQueryAsync(IAsyncResult asyncResult)
[...]

PR #455, opened 2022-11-04
Use app locks for all transactions

Charles-Gagnon:
## Description
This is to address deadlocks that I started seeing when running benchmarks with multiple hosts. Even just 2 hosts running in parallel would very consistently get in a deadlock after a couple of minutes. I was able to track the deadlock down to the [GetChanges ](https://github.com/Azure/azure-functions-sql-extension/blob/triggerbindings/src/TriggerBinding/SqlTableChangeMonitor.cs#L750) and [UpdateTablesPostInvocation](https://github.com/Azure/azure-functions-sql-extension/blob/triggerbindings/src/TriggerBinding/SqlTableChangeMonitor.cs#L916) - although I did not dig into it further enough to know exactly what table it was deadlocked on.

Explanation copied from the comment : 

```
/// The trigger heavily uses transactions to ensure atomic changes, that way if an error occurs during any step of a process we aren't left
/// with an incomplete state. Because of this locks are placed on rows that are read/modified during the transaction, but the lock isn't
/// applied until the statement itself is executed. Some transactions have many statements executed in a row that touch a number of different
/// tables so it's very easy for two transactions to get in a deadlock depending on the speed they execute their statements and the order they
/// are processed in.
///
/// So to avoid this we use application locks to ensure that anytime we enter a transaction we first guarantee that we're the only transaction
/// currently making any changes to the tables, which means that we're guaranteed not to have any deadlocks - albeit at the cost of speed. This
/// is acceptable for now, although further investigation could be done into using multiple resources to lock on (such as a different one for each
/// table) to increase the parallelization of the transactions.
```

I looked at various ways to change the existing locking to remove the possibility of a deadlock, but in the end I came to the conclusion that with the transaction-based nature of most of our statements and how many things each transaction does that trying to do so would be very risky and maybe even impossible.

The app locking approach is very heavy handed - but it's a very straightforward way to ensure application consistency. 

The main downside with this is that we will take a perf hit when multiple functions are running in parallel. But right now focusing on consistency and correctness is more important to me so I wanted to start with this and then selectively make improvements later so we can be sure they're tested properly. 

## Testing

First part was just using the benchmark parallelization suite. Before this change I would very consistently hit a deadlock within a couple of minutes of running with even just two hosts running in parallel. After the change I was able to run the full suite with both 2 hosts and 5 hosts (which runs each benchmark test 50+ times to generate the histogram)

(note that the overall times include the startup time for the host processes, which is why we're seeing such huge values. I'll be following up with fixing that so it only measures the time taken to get all the change notifications but for purposes of this fix that number isn't directly relevant)

```
SqlTriggerBindingPerformance_Parallelization.MultiHost: Job-FLKWWV(InvocationCount=1, UnrollFactor=1) [hostCount=2]
Runtime = .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT; GC = Concurrent Workstation
Mean = 16.049 s, StdErr = 0.094 s (0.58%), N = 81, StdDev = 0.843 s
Min = 14.720 s, Q1 = 15.258 s, Median = 15.976 s, Q3 = 16.769 s, Max = 17.892 s
IQR = 1.510 s, LowerFence = 12.993 s, UpperFence = 19.035 s
ConfidenceInterval = [15.729 s; 16.369 s] (CI 99.9%), Margin = 0.320 s (1.99% of Mean)
Skewness = 0.37, Kurtosis = 1.9, MValue = 2.65
-------------------- Histogram --------------------
[14.464 s ; 15.014 s) | @@@
[15.014 s ; 15.526 s) | @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
[15.526 s ; 16.044 s) | @@@@@@@
[16.044 s ; 16.555 s) | @@@@@@@@@@@@@@@@@
[16.555 s ; 17.233 s) | @@@@@@@@@@@@@@@
[17.233 s ; 17.762 s) | @@@@@@@
[17.762 s ; 18.148 s) | @
---------------------------------------------------

SqlTriggerBindingPerformance_Parallelization.MultiHost: Job-FLKWWV(InvocationCount=1, UnrollFactor=1) [hostCount=5]
Runtime = .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT; GC = Concurrent Workstation
Mean = 38.582 s, StdErr = 0.221 s (0.57%), N = 57, StdDev = 1.671 s
Min = 35.312 s, Q1 = 37.624 s, Median = 38.795 s, Q3 = 39.644 s, Max = 42.237 s
IQR = 2.020 s, LowerFence = 34.594 s, UpperFence = 42.674 s
ConfidenceInterval = [37.813 s; 39.350 s] (CI 99.9%), Margin = 0.769 s (1.99% of Mean)
Skewness = -0.06, Kurtosis = 2.37, MValue = 2
-------------------- Histogram --------------------
[34.742 s ; 36.062 s) | @@@@@
[36.062 s ; 37.202 s) | @@@@@@@@
[37.202 s ; 38.602 s) | @@@@@@@@@@@@@@
[38.602 s ; 40.013 s) | @@@@@@@@@@@@@@@@@@@@@
[40.013 s ; 41.561 s) | @@@@@@@
[41.561 s ; 42.807 s) | @@
---------------------------------------------------
```

After getting that to pass I compared the execution times of the single host to make sure the app locking didn't have a huge impact on that : 

Without App Locking:
```
|              Method | count |     Mean |    Error |   StdDev | Allocated |
|-------------------- |------ |---------:|---------:|---------:|----------:|
| ProductsTriggerTest |     1 |  1.040 s | 0.0094 s | 0.0088 s |     12 KB |
| ProductsTriggerTest |    10 |  1.052 s | 0.0142 s | 0.0133 s |     26 KB |
| ProductsTriggerTest |   100 |  1.153 s | 0.0085 s | 0.0076 s |    174 KB |
| ProductsTriggerTest |  1000 | 10.565 s | 0.0350 s | 0.0310 s |  1,737 KB |
```

With App Locking:
```
|              Method | count |     Mean |    Error |   StdDev | Allocated |
|-------------------- |------ |---------:|---------:|---------:|----------:|
| ProductsTriggerTest |     1 |  1.042 s | 0.0099 s | 0.0092 s |     11 KB |
| ProductsTriggerTest |    10 |  1.056 s | 0.0098 s | 0.0087 s |     26 KB |
| ProductsTriggerTest |   100 |  1.170 s | 0.0147 s | 0.0137 s |    177 KB |
| ProductsTriggerTest |  1000 | 10.612 s | 0.0396 s | 0.0310 s |  1,718 KB |
```

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -718,6 +718,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         private SqlCommand BuildUpdateTablesPreInvocation(SqlConnection connection, SqlTransaction transaction)
         {
             string updateTablesPreInvocationQuery = $@"
+                {AppLockStatements}
+
                 DECLARE @min_valid_version bigint;
                 SET @min_valid_version = CHANGE_TRACKING_MIN_VALID_VERSION({this._userTableId});
 
@@ -748,6 +750,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             string leasesTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = l.{col.name.AsBracketQuotedString()}"));
 
             string getChangesQuery = $@"
+                {AppLockStatements}
+
                 DECLARE @last_sync_version bigint;
                 SELECT @last_sync_version = LastSyncVersion
                 FROM {GlobalStateTableName}
@@ -761,7 +765,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     l.{LeasesTableAttemptCountColumnName},
                     l.{LeasesTableLeaseExpirationTimeColumnName}
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
-                LEFT OUTER JOIN {this._leasesTableName} AS l WITH (TABLOCKX) ON {leasesTableJoinCondition}
+                LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
                 LEFT OUTER JOIN {this._userTable.BracketQuotedFullName} AS u ON {userTableJoinCondition}
                 WHERE
                     (l.{LeasesTableLeaseExpirationTimeColumnName} IS NULL AND
@@ -785,6 +789,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             string leasesTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = l.{col.name.AsBracketQuotedString()}"));
 
             string getUnprocessedChangesQuery = $@"
+                {AppLockStatements}
+
                 DECLARE @last_sync_version bigint;
                 SELECT @last_sync_version = LastSyncVersion
                 FROM {GlobalStateTableName}
@@ -792,7 +798,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                 SELECT COUNT_BIG(*)
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
-                LEFT OUTER JOIN {this._leasesTableName} AS l WITH (TABLOCKX) ON {leasesTableJoinCondition}
+                LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
                 WHERE
                     (l.{LeasesTableLeaseExpirationTimeColumnName} IS NULL AND
                        (l.{LeasesTableChangeVersionColumnName} IS NULL OR l.{LeasesTableChangeVersionColumnName} < c.{SysChangeVersionColumnName}) OR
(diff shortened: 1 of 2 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -761,7 +765,7 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
                     l.{LeasesTableAttemptCountColumnName},
                     l.{LeasesTableLeaseExpirationTimeColumnName}
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
-                LEFT OUTER JOIN {this._leasesTableName} AS l WITH (TABLOCKX) ON {leasesTableJoinCondition}
+                LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
```
Charles-Gagnon, line 768: Since we're using the application lock the table locks are unnecessary

chlafreniere (commented):
What does this do for perf when there's only one function running at a time? Is calling `sp_getapplock` an expensive operation?

Charles-Gagnon:
> What does this do for perf when there's only one function running at a time? Is calling `sp_getapplock` an expensive operation?

Added results from single-host perf tests to description (there didn't seem to be any noticeable impact)

AmeyaRele (approved):

chlafreniere:
Thanks for setting up the benchmarks with/without app locking. Really glad that the runtime impact time is very minor.

JatinSanghvi:
Should we add app lock in queries inside `SqlTriggerListener.cs` too? That might fix https://github.com/Azure/azure-functions-sql-extension/issues/406.

PR #458, opened 2022-11-07
Add java samples

lucyzhang929:
Will create a separate PR for Java integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,7 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a 
 
 #### Output Binding Tutorial
 
-Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](#Create-a-SQL-Server), and that you have the 'Employee.cs' class from the [Input Binding Tutorial](#Input-Binding-Tutorial).
+Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](#Create-a-SQL-Server), and that you have the 'Employee.java' class from the [Input Binding Tutorial](#Input-Binding-Tutorial).
 
 - Open your app in VSCode
 - Press 'F1' and search for 'Azure Functions: Create Function'
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Azure SQL bindings for Azure Functions are supported for:
 - .NET functions (C# in-process)
 - NodeJS functions (JavaScript/TypeScript)
 - Python functions
+- Java functions
 
 ## Table of Contents
 
@@ -32,6 +33,7 @@ Azure SQL bindings for Azure Functions are supported for:
     - [.NET functions](#net-functions)
     - [JavaScript functions](#javascript-functions)
     - [Python functions](#python-functions)
+    - [Java functions](#java-functions)
   - [More Samples](#more-samples)
     - [Input Binding](#input-binding)
       - [Query String](#query-string)
@@ -100,7 +102,7 @@ These steps can be done in the Terminal/CLI or with PowerShell.
 
 1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
 
-2. Create a function app for .NET, JavaScript, TypeScript or Python.
+2. Create a function app for .NET, JavaScript, TypeScript, Python, or Java.
 
     **.NET**
     ```bash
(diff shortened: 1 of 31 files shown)


samples/samples-java/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library</artifactId>
+            <version>${azure.functions.java.library.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library-sql</artifactId>
+            <version>0.1.0</version>
```
Charles-Gagnon, line 30: Add comment next to the library POM value for this version that reminds us to update the samples version as well if that is changed

samples/samples-java/pom.xml
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.function</groupId>
+    <artifactId>samples-java</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <name>Azure Java Functions</name>
```
Charles-Gagnon, line 10: Maybe this?

```suggestion
    <name>Java Azure Functions with SQL Bindings</name>
```

samples/samples-java/src/main/java/com/function/AddProduct.java
```diff
@@ -0,0 +1,41 @@
...
+            @HttpTrigger(
+                name = "req",
+                methods = {HttpMethod.POST},
+                authLevel = AuthorizationLevel.ANONYMOUS,
+                route = "addproduct")
+                HttpRequestMessage<Optional<String>> request,
+            @SQLOutput(
+                commandText = "Products",
+                connectionStringSetting = "SqlConnectionString")
+                OutputBinding<Product> product) throws JsonParseException, JsonMappingException, IOException {
```
Charles-Gagnon, line 32: Out of curiosity where did you get this list of exceptions? I have to imagine that isn't all of them - for example if there's a SqlException in the function itself what kind of exception is thrown in the worker?

Not a huge deal so we can leave it like this for now - just want to know what to expect here.

samples/samples-java/src/main/java/com/function/AddProduct.java
```diff
@@ -0,0 +1,41 @@
+package com.function;
```
Charles-Gagnon, line 1: Copyright for these

Can you create a follow up task to investigate linting options for Java?

Charles-Gagnon (commented):
Update the README too

samples/samples-java/src/main/java/com/function/AddProduct.java
lucyzhang929: The exceptions come from mapper.readValue() in line 36. If I don't add the throws, there are build errors. Ex: `Unhandled exception type JsonProcessingException`

samples/samples-java/src/main/java/com/function/AddProduct.java
lucyzhang929: Tracking here: https://github.com/Azure/azure-functions-sql-extension/issues/462

README.md
```diff
@@ -502,6 +529,148 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a
...
+- Navigate back to your HttpTriggerJava file.
+- Open the local.settings.json file, and in the brackets for "Values," verify there is a 'SqlConnectionString.' If not, add it.
+- Hit 'F5' to run your code. This will start up the Functions Host with a local HTTP Trigger and SQL Input Binding.
+- Click the link that appears in your terminal.
+- You should see your database output in the browser window.
+- Congratulations! You have successfully created your first SQL input binding! Checkout [Input Binding](#Input-Binding) for more information on how to use it and explore on your own!
+
+#### Output Binding Tutorial
+
+Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](#Create-a-SQL-Server), and that you have the 'Employee.cs' class from the [Input Binding Tutorial](#Input-Binding-Tutorial).
```
Charles-Gagnon, line 639: `have the 'Employee.java' file`

Charles-Gagnon (approved):

PR #460, opened 2022-11-07
Fix perf tests

Charles-Gagnon:
My last PR broke the non-multi function tests since they weren't starting up a new host every iteration. While looking at this I decided to also fix up the problem where we were starting up the host as part of the test which gave very inflated numbers for the execution time of each iteration.

Now each test will start the host once and then run through its iterations (cleaning up the tables after every run). This makes them run significantly faster and gives more accurate numbers. 

I also modified some of the values we use to make the runtimes a bit more reasonable. We still have some that run for quite a while that I want to get back to, but it's at least not 30+min for some of the runs now.


Charles-Gagnon:
Merging this in now to get other PR out - leave any comments here and I'll follow up later

PR #469, opened 2022-11-08
Trigger bindings -> main

Charles-Gagnon:
(new branch because I made one small change to a pipeline and didn't want to send out a whole separate PR for just that)


JatinSanghvi:
Should we merge-commit this branch into main? That will retain individual contributions, and will help with git-blame.

Charles-Gagnon:
Yeah that's a good suggestion. Wish we had been more diligent about squash-merging into triggerbindings but a little bit of a messy commit history is worth being able to easily track down the history of a change.

PR #471, opened 2022-11-08
Move and split README docs

Charles-Gagnon:
Note: Currently targeting triggerbindings branch for reviewing changes while the other PR is still open. This should be changed to target main after that's in and then merged in once things look good

Splitting up README content into separate files in the docs folder. Very minimal changes to actual content - that will be done as a follow up task.


Charles-Gagnon:
Let me think about it - the problem I see with that is that there'd be a lot of duplicated content between all the languages for all the "common" setup and reference stuff in each binding type (which was something I was going to consolidate in a follow up PR). Maybe we can have a top level one for each binding type and then one for each language that it can link to.

Charles-Gagnon:
Going with per-language pages for now, will follow up with seeing what we can do to organize it better later on

docs/QuickStart.md
```diff
@@ -0,0 +1,201 @@
...
+
+    ```powershell
+    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --prerelease
+    ```
+
+    **JavaScript and TypeScript:** Update the `host.json` file to the preview extension bundle.
+    ```json
+    "extensionBundle": {
+        "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
+        "version": "[4.*, 5.0.0)"
```
VasuBhog, line 106: Shouldn't this be included in each of the language's Binding Guides? 
Users can now directly click on the language details and skip over this part.

README.md
```diff
@@ -1,14 +1,21 @@
...
+- **Input Binding**: takes a SQL query to run and returns the output of the query in the function.
+- **Output Binding**: takes a list of rows and upserts them into the user table (i.e. If a row doesn't already exist, it is added. If it does, it is updated).
+- **Trigger Binding**: monitors the user table for changes (i.e., row inserts, updates, and deletes) and invokes the function with updated rows.
+
+See the language specific guides for further details on support and usage of these bindings in the supported languages.
+
+- [.NET](./docs/BindingsGuide_Dotnet.md)
+- [Java](./docs/BindingsGuide_Java.md)
+- [Javascript](./docs/BindingsGuide_Javascript.md)
+- [Python](./docs/BindingsGuide_Python.md)
```
VasuBhog, line 18: Can we just condense what we have here with that on line 22 https://github.com/Azure/azure-functions-sql-extension/pull/471/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R22

README.md
```diff
@@ -20,1091 +27,13 @@ Azure SQL bindings for Azure Functions are supported for:
 
 ## Table of Contents
```
VasuBhog, line 28: The table of contents should be at the top of the ReadMe.

docs/QuickStart.md
VasuBhog, line 106: Shouldn't this be included in each of the language's Binding Guides?
Users can now directly click on the language details and skip over this part.

README.md
VasuBhog, line 18: Can we just condense what we have here with that on line 22 https://github.com/Azure/azure-functions-sql-extension/pull/471/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R22

README.md
VasuBhog, line 28: The table of contents should be at the top of the ReadMe.

docs/BindingsGuide_Javascript.md
```diff
@@ -0,0 +1,166 @@
+# Azure SQL bindings for Azure Functions - Javascript
+
+## Input Binding
+
+### function.json Properties for Input Bindings
+
+_TODO_
```
VasuBhog, line 7: What do you mean to put in this section exactly as it is prior to the setup for Input Bindings

Charles-Gagnon: Same as your powershell examples - it will be a table of the properties that the user should have in their function.json

https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsGuide_PowerShell.md#functionjson-properties-for-input-bindings

docs/QuickStart.md
Charles-Gagnon: Yeah, definitely. We can have this be general QuickStart stuff and then each language guide have a language-specific one-time setup section at the top

Issue #476, opened 2022-11-10
Finish Doc TODOs - PowerShell


VasuBhog:
Finish TODO sections in https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsGuide_PowerShell.md

PR #477, opened 2022-11-10
Add powershell docs

VasuBhog:
This adds PowerShell specific docs to the repo.

Follow up for TODO items: https://github.com/Azure/azure-functions-sql-extension/issues/476

### DIFF
--- a/docs/BindingsGuide_PowerShell.md
+++ b/docs/BindingsGuide_PowerShell.md
@@ -114,7 +114,7 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a 
 - Open your app in VS Code
 - Press 'F1' and search for 'Azure Functions: Create Function'
 - Choose HttpTrigger ->  (Provide a function name) -> anonymous
-- In the file that opens (`index.js`), replace the `module.exports = async function (context, req)` block with the below code.
+- In the file that opens (`run.ps1`), replace the code within the file the below code.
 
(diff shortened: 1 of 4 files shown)


docs/BindingsGuide_PowerShell.md
```diff
@@ -0,0 +1,206 @@
+# Azure SQL bindings for Azure Functions - PowerShell
+
+## Input Binding
+
+### function.json Properties for Input Bindings
+```json
```
Charles-Gagnon, line 6: The intention I had for this section was to make it a table of properties and what they mean like this : https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-input?tabs=in-process&pivots=programming-language-javascript#configuration

You can do it now if you want since that should be pretty straightforward (just copy the table since it should be the same content for the most part) but we could also save it and do them for all the languages at the same time so we're sure they have the same content.

docs/BindingsGuide_PowerShell.md
```diff
@@ -0,0 +1,206 @@
...
+
+#### Stored Procedure
+
+_TODO_
+
+## Output Binding
+
+### function.json Properties for Output Bindings
+
+```json
```
Charles-Gagnon, line 93: See above

Charles-Gagnon (approved):

PR #479, opened 2022-11-10
Log event sends as Trace

Charles-Gagnon:
Information is too verbose for these - we don't need them showing up in the default logging.

### DIFF
--- a/src/Telemetry/Telemetry.cs
+++ b/src/Telemetry/Telemetry.cs
@@ -96,7 +96,7 @@ This extension collect usage data in order to help us improve your experience. T
                 {
                     return;
                 }
-                this._logger.LogInformation($"Sending event {eventName}");
+                this._logger.LogTrace($"Sending event {eventName}");
 
                 //continue task in existing parallel thread
                 this._trackEventTask = this._trackEventTask.ContinueWith(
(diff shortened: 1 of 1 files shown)


src/Telemetry/Telemetry.cs
```diff
@@ -96,7 +96,7 @@ private void InitializeTelemetry(string productVersion, IConfiguration config)
                 {
                     return;
                 }
-                this._logger.LogInformation($"Sending event {eventName}");
+                this._logger.LogTrace($"Sending event {eventName}");
```
MaddyDev, line 99: Can this still be left as LogInformation?

Charles-Gagnon: Why do you think this should be left as is? Currently having these as info fills up the logs with tons of these events (especially for the trigger binding) - and customers don't generally care about us sending telemetry events. 

(I might just completely remove these later on but at least having them as Trace for right now hides them in most cases)

MaddyDev: I was coming from the differentiation point between trace and information. I see your point, thanks for clarifying.

MaddyDev (approved):

lucyzhang929 (approved):

PR #480, opened 2022-11-10
Update docs to add function.json attributes and sql attributes for Java

VasuBhog:
Added Setup function app instructions prior to input/output setup as users can come directly to language files. 

I also added attributes and function.json table for each input/output language.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ This project has adopted the Microsoft Open Source Code of Conduct. For more inf
 
 ### SQL Setup
 
-In order to test changes it is suggested that you have a SQL server set up to connect to and run queries against. Instructions to set this up can be found in the [Quick Start Guide](./docs/QuickStart.md)
+In order to test changes it is suggested that you have a SQL server set up to connect to and run queries against. Instructions to set this up can be found in the [Quick Start Guide](./docs/GeneralSetup.md)
 
 ### Set Up Development Environment
 
@@ -28,6 +28,6 @@ code .
 3. Install extensions when prompted after VS Code opens
    - Note: This includes the Azure Functions, C#, and editorconfig extensions
 
-4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/QuickStart.md#configure-function-app)
+4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/GeneralSetup.md#configure-function-app)
 
 5. Press F5 to run SQL bindings samples that are included in this repo. The output window should display startup information as well as the function endpoints that were started.
--- a/docs/BindingsGuide_Dotnet.md
+++ b/docs/BindingsGuide_Dotnet.md
@@ -1,5 +1,23 @@
 # Azure SQL bindings for Azure Functions - .NET
 
+## Setup Function App
+1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
+
+2. Create a function app for .NET:
+    ```bash
+    mkdir MyApp
+    cd MyApp
+    func init --worker-runtime dotnet
+    ```
+
+3. Enable SQL bindings on the function app. More information can be found [in Microsoft Docs](https://docs.microsoft.com/azure/azure-functions/functions-bindings-azure-sql).
+
+    Install the extension.
+
+    ```powershell
+    dotnet add package Microsoft.Azure.WebJobs.Extensions.Sql --prerelease
+    ```
+
 ## Input Binding
 
 ### SqlAttribute for Input Bindings
--- a/docs/BindingsGuide_Dotnet.md
+++ b/docs/BindingsGuide_Dotnet.md
@@ -2,7 +2,7 @@
 
 ## Setup Function App
 
-These instructions will guide you through creating your Function App and adding the SQL binding extension. This only needs to be done once for every function app you create.
+These instructions will guide you through creating your Function App and adding the SQL binding extension. This only needs to be done once for every function app you create. If you have one created already you can skip this step.
 
 1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
 
@@ -43,9 +43,9 @@ The repo contains examples of each of these binding types [here](https://github.
 
 ### Setup for Input Bindings
 
-Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./QuickStart.md#create-a-sql-server).
+Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./GeneralSetup.md#create-a-sql-server).
 
-- Open your app that you created in [Create a Function App](./QuickStart.md#create-a-function-app) in VS Code
+- Open your app that you created in [Create a Function App](./GeneralSetup.md#create-a-function-app) in VS Code
 - Press 'F1' and search for 'Azure Functions: Create Function'
 - Choose HttpTrigger -> (Provide a function name) -> Company.namespace -> anonymous
 - In the file that opens, replace the `public static async Task<IActionResult> Run` block with the below code.
@@ -235,7 +235,7 @@ The repo contains examples of each of these binding types [here](https://github.
 
 ### Setup for Output Bindings
 
-Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./QuickStart.md#create-a-sql-server), and that you have the 'Employee.cs' class from the [Setup for Input Bindings](#setup-for-input-bindings) section.
+Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./GeneralSetup.md#create-a-sql-server), and that you have the 'Employee.cs' class from the [Setup for Input Bindings](#setup-for-input-bindings) section.
 
 - Open your app in VS Code
 - Press 'F1' and search for 'Azure Functions: Create Function'
@@ -483,7 +483,7 @@ public static void Run(
 }
 ```
 
-Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./QuickStart.md#create-a-sql-server), and that you have the 'Employee.cs' file from the [Setup for Input Bindings](#setup-for-input-bindings) section.
+Note: This tutorial requires that a SQL database is setup as shown in [Create a SQL Server](./GeneralSetup.md#create-a-sql-server), and that you have the 'Employee.cs' file from the [Setup for Input Bindings](#setup-for-input-bindings) section.
 
 - Create a new file with the following content:
 
--- a/docs/BindingsGuide_Dotnet.md
+++ b/docs/BindingsGuide_Dotnet.md
@@ -1,6 +1,9 @@
 # Azure SQL bindings for Azure Functions - .NET
 
 ## Setup Function App
+
+These instructions will guide you through creating your Function App and adding the SQL binding extension. This only needs to be done once for every function app you create.
+
 1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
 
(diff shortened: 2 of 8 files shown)


docs/BindingsGuide_Java.md
```diff
@@ -1,10 +1,39 @@
...
 ## Input Binding
 
 ### SQLInput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLInput annotation (com.microsoft.azure.functions.sql.annotation.SQLInput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
+
+| Element |Description|
+|---------|---------|
+| **commandText** | Required. The Transact-SQL query command or name of the stored procedure executed by the binding.  |
```
Charles-Gagnon, line 31: Sync up with @lucyzhang929 on adding the name property

docs/BindingsGuide_Java.md
```diff
@@ -1,10 +1,39 @@
...
+    ```
+
 ## Input Binding
 
 ### SQLInput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLInput annotation (com.microsoft.azure.functions.sql.annotation.SQLInput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
+
+| Element |Description|
```
Charles-Gagnon, line 29: I believe this should be Attribute, not Element @lucyzhang929 

```suggestion
| Attribute | Description |
```

docs/BindingsGuide_Java.md
```diff
@@ -1,10 +1,39 @@
...
+|---------|---------|
+| **commandText** | Required. The Transact-SQL query command or name of the stored procedure executed by the binding.  |
+| **connectionStringSetting** | The name of an app setting that contains the connection string for the database against which the query or stored procedure is being executed. This value isn't the actual connection string and must instead resolve to an environment variable name. | 
+| **commandType** | A [CommandType](/dotnet/api/system.data.commandtype) value, which is [Text](/dotnet/api/system.data.commandtype#fields) for a query and [StoredProcedure](/dotnet/api/system.data.commandtype#fields) for a stored procedure. |
```
Charles-Gagnon, line 33: These links don't work (elsewhere in this PR + in your Powershell docs)

docs/BindingsGuide_Java.md
```diff
@@ -1,10 +1,39 @@
...
+        "version": "[4.*, 5.0.0)"
+    }
+    ```
+
 ## Input Binding
 
 ### SQLInput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLInput annotation (com.microsoft.azure.functions.sql.annotation.SQLInput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
```
Charles-Gagnon, line 27:
```suggestion
In the Java functions runtime library, use the @SQLInput annotation (com.microsoft.azure.functions.sql.annotation.SQLInput) on parameters whose value comes from the query specified by commandText. This annotation supports the following elements:
```

docs/BindingsGuide_Java.md
```diff
@@ -121,7 +150,14 @@ _TODO_
 
 ### SQLOutput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLOutput annotation (com.microsoft.azure.functions.sql.annotation.SQLOutput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
```
Charles-Gagnon, line 153:
```suggestion
In the Java functions runtime library, use the @SQLOutput annotation (com.microsoft.azure.functions.sql.annotation.SQLOutput) on parameters whose values you want to upsert into the target table
```

docs/BindingsGuide_Java.md
```diff
@@ -121,7 +150,14 @@ _TODO_
...
+
+| Element |Description|
+|---------|---------|
+| **commandText** | Required. The name of the table being written to by the binding.  |
```
Charles-Gagnon, line 157: Add name attribute

docs/BindingsGuide_Java.md
```diff
@@ -121,7 +150,14 @@ _TODO_
 
 ### SQLOutput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLOutput annotation (com.microsoft.azure.functions.sql.annotation.SQLOutput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
+
+| Element |Description|
```
Charles-Gagnon, line 155:
```suggestion
| Attribute | Description |
```

docs/BindingsGuide_Java.md
```diff
@@ -1,10 +1,39 @@
...
 
 ### SQLInput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLInput annotation (com.microsoft.azure.functions.sql.annotation.SQLInput) on parameters whose value would come from Azure SQL. This annotation supports the following elements:
+
+| Element |Description|
+|---------|---------|
+| **commandText** | Required. The Transact-SQL query command or name of the stored procedure executed by the binding.  |
+| **connectionStringSetting** | The name of an app setting that contains the connection string for the database against which the query or stored procedure is being executed. This value isn't the actual connection string and must instead resolve to an environment variable name. |
```
Charles-Gagnon, line 32: This is different than the other connection string setting descriptions you have, let's make sure these are all the same

(note the other ones - including your powershell one - have broken links for the extra connection string settings. So make sure to fix that too)

docs/QuickStart.md
```diff
@@ -52,7 +52,7 @@ These steps can be done in the Terminal/CLI or with PowerShell.
 
 1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
```
Charles-Gagnon, line 53: Remove the language-specific setup now that it's in each separate guide. This one should only contain general setup

(and it wouldn't hurt to include a link to this in each of those guides as well to make sure someone doesn't go directly there and miss doing these)

docs/BindingsGuide_Dotnet.md
```diff
@@ -1,5 +1,23 @@
 # Azure SQL bindings for Azure Functions - .NET
 
+## Setup Function App
+1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
```
Charles-Gagnon, line 4: Let's be explicit that this is a one-time setup (do similar for the other pages)

```suggestion
## Setup Function App

These instructions will guide you through creating your Function App and adding the SQL binding extension. This only needs to be done once for every function app you create.

1. Install [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local)
```

docs/BindingsGuide_Java.md
lucyzhang929: Looking through the docs - some bindings use Property and some use Element. Looks like Drew used Element for SQL Bindings https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-output?tabs=in-process&pivots=programming-language-java#annotations

Charles-Gagnon: I guess I was wrong - these docs mention "elements" 

https://docs.oracle.com/javase/tutorial/java/annotations/basics.html

`The annotation can include elements, which can be named or unnamed, and there are values for those elements:`

So nevermind then - Element sounds good to me!

docs/BindingsGuide_Java.md
VasuBhog: as discussed will leave it as Element

docs/BindingsGuide_Java.md
```diff
@@ -121,7 +153,15 @@ _TODO_
 
 ### SQLOutput Attribute
 
-_TODO_
+In the Java functions runtime library, use the @SQLOutput annotation (com.microsoft.azure.functions.sql.annotation.SQLOutput) on parameters whose values you want to upsert into the target table. This annotation supports the following elements:
+
+| Element |Description|
+|---------|---------|
+|**name** |  Required. The name of the variable that represents the query results in function code. |
```
Charles-Gagnon, line 160: This isn't correct (there are no "query results" in output bindings)

@lucyzhang929 We should probably just use the same descriptions you added in your PR

https://github.com/Azure/azure-functions-sql-extension/pull/481/files#diff-07386b4163c1406052b8b902fb2322202c7a235a720c958e540ea4f76fe53f0aR21

That sound good to you?

docs/QuickStart.md
```diff
@@ -48,126 +48,13 @@ ALTER TABLE ['{table_name}'] ADD CONSTRAINT PKey PRIMARY KEY CLUSTERED (['{prima
 
 Now you will need a Function App to add the binding to. If you have one created already you can skip this step.
```
Charles-Gagnon, line 49: This sentence doesn't make sense anymore. And the flow is messed up now - users will need to : 

1. Start here to configure their table
2. Somehow figure out that they need to go into the language specific guides to create their function app
3. Come back here to finish configuring it

I'm not sure I like the idea of duplicating all the content in this file in each language guide - please see if you can come up with something better (and make sure you're walking through the guide yourself to verify things make sense and flow well)

docs/QuickStart.md
```diff
@@ -185,15 +72,15 @@ Once you have your Function App you need to configure it for use with Azure SQL
...
+3. Open the generated `local.settings.json` file and in the `Values` section verify you have the below. If not, add the below and replace `{connection_string}` with the your connection string from the previous step:
 
     ```json
     "AzureWebJobsDashboard": "UseDevelopmentStorage=true",
     "SqlConnectionString": "{connection_string}"
     ```
 
-1. Verify your `host.json` looks like the below:
+4. Verify your `host.json` looks like the below:
```
Charles-Gagnon, line 83: While you're here - could you remove this? It isn't necessary since we aren't actually changing anything and we don't want to confuse people if the default host.json generated changes.

Charles-Gagnon (changes requested):

docs/QuickStart.md
VasuBhog: I think I will keep the stuff in the Quickstart for now as each language has a section that indicates users to go to the quickstart guide to go setup function app further. I just believe that not having the host.json preview bundle in the language is also not a great user experience. Duplicating information is better than having a confused user.

Charles-Gagnon: The bottom of quickstart still references having the user go to the language specific guides. So if you think it makes sense like that then let's fix everything up to follow logically with appropriate links at each stage. I would also suggest we rename this from `QuickStart.md` to something more general like `GeneralSetup.md` since QuickStart to me implies the user should be starting there.

VasuBhog: reverted the fix for now. I believe having the information in both places is fine as we include `If you have one created already you can skip this step.` so that if a user started in language-specific doc or general startup they could understand that they would not need it.

docs/BindingsGuide_Java.md
VasuBhog: Followed up and made it `Required. The variable name used in function.json.` as reflected in the PR Lucy added.

CONTRIBUTING.md
```diff
@@ -12,7 +12,7 @@ This project has adopted the Microsoft Open Source Code of Conduct. For more inf
 
 ### SQL Setup
 
-In order to test changes it is suggested that you have a SQL server set up to connect to and run queries against. Instructions to set this up can be found in the [Quick Start Guide](./docs/QuickStart.md)
+In order to test changes it is suggested that you have a SQL server set up to connect to and run queries against. Instructions to set this up can be found in the [Quick Start Guide](./docs/GeneralSetup.md)
```
Charles-Gagnon, line 15:
```suggestion
In order to test changes it is suggested that you have a SQL server set up to connect to and run queries against. Instructions to set this up can be found in the [General Setup Guide](./docs/GeneralSetup.md)
```

Charles-Gagnon (approved):

PR #481, opened 2022-11-10
Add name to java binding interface

lucyzhang929:
Previously, we could not set the name for the bindings which meant that we would get the following error if we tried to create a function with multiple sql bindings:
`"The 'HttpTriggerJava1' function is in error: Multiple bindings with name 'sql' discovered. Binding names must be unique."`

With this change, name will be required and the current samples have to be updated with a name value to work.

### DIFF
--- a/java-library/pom.xml
+++ b/java-library/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.microsoft.azure.functions</groupId>
   <artifactId>azure-functions-java-library-sql</artifactId>
-  <version>0.1.0</version> <!-- Update the dependency version in samples-java/pom.xml too if this version is updated. -->
+  <version>0.1.1</version> <!-- Update the dependency version in samples-java/pom.xml too if this version is updated. -->
   <packaging>jar</packaging>
 
   <parent>
--- a/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
+++ b/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
@@ -15,12 +15,12 @@ import com.microsoft.azure.functions.annotation.CustomBinding;
 
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.PARAMETER)
-@CustomBinding(direction = "in", name = "sql", type = "sql")
+@CustomBinding(direction = "in", name = "", type = "sql")
 public @interface SQLInput {
     /**
      * The variable name used in function.json.
      */
-    String name() default "sql";
+    String name();
 
     /**
      * Query string or name of stored procedure to be run.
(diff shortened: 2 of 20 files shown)


java-library/pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.microsoft.azure.functions</groupId>
   <artifactId>azure-functions-java-library-sql</artifactId>
-  <version>0.1.0</version> <!-- Update the dependency version in samples-java/pom.xml too if this version is updated. -->
+  <version>0.1.1</version> <!-- Update the dependency version in samples-java/pom.xml too if this version is updated. -->
```
Charles-Gagnon, line 7: See inline comment 😄

Charles-Gagnon (approved):

java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
```diff
@@ -17,6 +17,11 @@
 @Target(ElementType.PARAMETER)
 @CustomBinding(direction = "in", name = "sql", type = "sql")
```
Charles-Gagnon, line 18: Wait - isn't the name here what you need to update? Or does having a name property in the interface override the one set here?

java-library/pom.xml
lucyzhang929: I was planning on sending another PR out to update the samples pom.xml + add new sample after the new library is released. But I guess releasing the library is a quick process so do you think it'd be better to just update the version in this PR too?

java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
```diff
@@ -17,6 +17,11 @@
...
 public @interface SQLInput {
+    /**
+     * The variable name used in function.json.
```
Charles-Gagnon, line 21: This isn't very clear to me - maybe something more like 

```suggestion
     * The unique name to give the binding in the function.json
```

java-library/pom.xml
Charles-Gagnon: Yeah, that's a bit tricky. But since it's quick I think we should get in the habit of updating them in step so we don't forget to do it. If that ends up being too annoying then we can remove this comment and just update our release docs to add a reminder to update the samples as well.

java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
lucyzhang929: Yes, having the name property without a default will require users to always specify a name.

java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
lucyzhang929: Wanted to keep the comments the same as other binding libraries to avoid confusion for users working with multiple bindings:
https://github.com/Azure/azure-functions-java-library/blob/5fe66b0dbf0e6973abfd618ac100813e8ac48bde/src/main/java/com/microsoft/azure/functions/annotation/CosmosDBInput.java#L57

lucyzhang929:
Updated the PR description and also added a sample with two output bindings.

Charles-Gagnon (approved):

MaddyDev (approved):

chlafreniere (approved):

PR #482, opened 2022-11-11
Cleanup samples folders

Charles-Gagnon:
- Add .vscode files for being able to launch/debug the samples directly from VS Code (note for non-C# in proc functions you'll need to open the target samples folder directly and run from there)
- Updated C# in proc samples to target correct runtimes
- Consolidate out of proc runtime from `net6.0` -> `net6`


lucyzhang929:
Should we remove the .vscode folder at the root?

lucyzhang929 (approved):

Charles-Gagnon:
I was thinking that but I'm fine leaving it like that so we can easily run *something* from the root.

I want to come back and have a top-level debug target for each of the samples, but that'll be a later improvement.

Issue #488, opened 2022-11-11
Renew Leases and Unprocessed change count queries are failing


Charles-Gagnon:
Seeing this error when the lease renewal logic runs. 

Note that this only shows up very occasionally because right now the lease renewal loop does not run very often at all due to a potential other issue with the locking logic there. So most of the time this exception won't even be seen currently. 

`Failed to renew leases due to exception: Microsoft.Data.SqlClient.SqlException. Exception message: Unable to acquire exclusive lock on _az_func_Trigger. Result = -999
[1012] [2022-11-11T21:50:31.399Z] You attempted to acquire a transactional application lock without an active transaction.
`

PR #489, opened 2022-11-11
Use transaction for all queries

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/488

The default owner is transaction, which means it'll return an error if the session isn't in a transaction at the time. 

Another possible fix would be to have the session be the owner for these ones, but if we did that then we'd have to add the logic (and error handling) to ensure that we released the lock. Using transactions is simpler and shouldn't add that much extra overhead given that we're already doing the app locking in the first place.


Charles-Gagnon:
Verified that this is now working correctly on deployed app : 

```
2022-11-12T00:31:47.542 [Debug] Computing scale status (WorkerCount=2)
2022-11-12T00:31:47.542 [Debug] 1 scale monitors to sample
2022-11-12T00:31:47.569 [Information] Unprocessed change counts: [0, 0, 0, 0, 0], worker count: 2, maximum changes per worker: 1000.
2022-11-12T00:31:47.570 [Information] Requesting scale-in: Found table: 'dbo.Products' to be either idle or the unprocessed changes to be continuously decreasing.
2022-11-12T00:31:47.570 [Debug] Monitor '50e10ce727225d2a-SqlTrigger-[dbo].[Products]' voted 'ScaleIn'
2022-11-12T00:31:47.574 [Debug] Scaling in based on votes
2022-11-12T00:31:47.640 [Debug] [HostMonitor] Host process CPU stats (PID 5032): History=(3,0,0,6,0), AvgCpuLoad=2, MaxCpuLoad=6
2022-11-12T00:31:47.650 [Debug] [HostMonitor] Host aggregate CPU load 2
2022-11-12T00:31:47.655 [Information] Executing StatusCodeResult, setting HTTP status code 200
2022-11-12T00:31:47.643 [Debug] [HostMonitor] Host process CPU stats (PID 1996): History=(0,0,0,0,12), AvgCpuLoad=2, MaxCpuLoad=12
```

chlafreniere (approved):

JatinSanghvi (approved):
LGTM.. Did we hit any error when testing scaling or something that involves multiple app instances? I remember the rowlock prevents getting into race condition inside the same process, but we still require SQL transactions for multi-process scenario.

Charles-Gagnon:
The issue meant that the query would always fail - the app lock SP requires that there's an active transaction if you don't specify the owner as session and if it doesn't then it returns -999. So we'd never get any metrics which meant scaling would never happen.

I haven't been testing the scaling stuff since right now all the benchmark tests have been local - and we don't have any integration tests for the scaling support (I'll be following up with you on that)

I noticed this during benchmark testing on the renew leases query and then went through the rest of the code and found this one as well (just looked for app lock statements that were being executed outside a transaction) and verified it was broken with an actual deployed app (and then followed up with verifying that it was fixed with this change).

Issue #457, opened 2022-11-07
Ensure README Includes Supported SQL Versions


chlafreniere:
We've had a couple of questions around what versions of SQL that we support. Some of our dependencies (e.g. ScriptDOM) are explicit in that they only support versions of SQL that are still under support (which makes sense 😄).

Would be good to call out:
- SQL box/VM versions
- Any known limitations for Azure SQL DB/MI/Arc
- Being explicit around compat level as well

PR #493, opened 2022-11-14
Add section on supported server versions to README

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/457

I have to actually go through and test out MI - but according to the docs for supported features it should work so I'm preemptively assuming it will and will merge this in once I've verified that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,9 @@ Further information on the Azure SQL binding for Azure Functions is also availab
 
 ## Supported SQL Server Versions
 
-This extension uses the [OPENJSON](https://learn.microsoft.com/sql/t-sql/functions/openjson-transact-sql) statement which requires SQL Server 2016 or higher and a database compatibility level of 130+. Older versions will currently fail to execute properly.
+This extension uses the [OPENJSON](https://learn.microsoft.com/sql/t-sql/functions/openjson-transact-sql) statement which requires a database compatibility level of 130 or higher (2016 or higher). To view or change the compatibility level of your database, see [this documentation article](https://learn.microsoft.com/sql/relational-databases/databases/view-or-change-the-compatibility-level-of-a-database) for more information.
 
-For more information about options to upgrade versions see the [SQL Server end of support options](https://learn.microsoft.com/sql/sql-server/end-of-support/sql-server-end-of-support-overview) page.
-
-Azure SQL Databases are fully supported.
+Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which meet the compatibility level requirement above are supported.
 
 ## Known Issues
 
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ See the language-specific guides for further details on support and usage of the
 
 Further information on the Azure SQL binding for Azure Functions is also available in the [Azure Functions docs](https://docs.microsoft.com/azure/azure-functions/functions-bindings-azure-sql).
 
+## Supported SQL Server Versions
+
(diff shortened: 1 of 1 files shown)


lucyzhang929 (approved):

README.md
```diff
@@ -26,11 +26,18 @@ See the language-specific guides for further details on support and usage of the
 
 Further information on the Azure SQL binding for Azure Functions is also available in the [Azure Functions docs](https://docs.microsoft.com/azure/azure-functions/functions-bindings-azure-sql).
 
+## Supported SQL Server Versions
+
+This extension uses the [OPENJSON](https://learn.microsoft.com/sql/t-sql/functions/openjson-transact-sql) statement which requires SQL Server 2016 or higher and a database compatibility level of 130+. Older versions will currently fail to execute properly.
+
+For more information about options to upgrade versions see the [SQL Server end of support options](https://learn.microsoft.com/sql/sql-server/end-of-support/sql-server-end-of-support-overview) page.
```
chlafreniere, line 33: This feels a little verbose to me.

Maybe something like `For more information about version upgrade options`

README.md
```diff
@@ -26,11 +26,18 @@ See the language-specific guides for further details on support and usage of the
...
+
+Azure SQL Databases are fully supported.
```
chlafreniere, line 35: @dzsquared is "Azure SQL Databases" what we should be calling the Azure SQL suite of products?

README.md
```diff
@@ -26,11 +26,18 @@ See the language-specific guides for further details on support and usage of the
 
 Further information on the Azure SQL binding for Azure Functions is also available in the [Azure Functions docs](https://docs.microsoft.com/azure/azure-functions/functions-bindings-azure-sql).
 
+## Supported SQL Server Versions
+
+This extension uses the [OPENJSON](https://learn.microsoft.com/sql/t-sql/functions/openjson-transact-sql) statement which requires SQL Server 2016 or higher and a database compatibility level of 130+. Older versions will currently fail to execute properly.
```
dzsquared, line 31: This extension uses the [OPENJSON](https://learn.microsoft.com/sql/t-sql/functions/openjson-transact-sql) statement which requires a database compatibility level of 130 or higher (2016 or higher).

README.md
dzsquared: I'd prefer this whole statement gets omitted, it's not about a discussion of what versions of the engine are in support - the SQL extension library has a requirement based on its architecture.

README.md
dzsquared: our triad of 3 products that we support is mentioned in my comment above - "Azure SQL Database" is 1 product, then "Azure SQL Managed Instance" is another.
I'd prefer to avoid lumping it as "Azure SQL" lest it opens confusion to whether we support Synapse, which we do not.

Charles-Gagnon: Yeah - to be clear this is only supposed to be for Azure SQL Database. I haven't had a chance to test out MI yet so wanted to hold off on saying anything until I can verify it (although it should work according to the docs). 

So given that do you think leaving it as Azure SQL Databases is fine or should I be more specific? (or even hold off until I can test MI so we can have an accurate list of supported engines?)

Charles-Gagnon: Updated it with the suggestion Drew had above - I'll test out MI before getting this in so that we have everything here be complete.

Charles-Gagnon:
@dzsquared @chlafreniere I'm going to merge this in so we have it out there. I'm having issues with getting the MI instance set up but I'll be following up to verify that when I can.

Leave any other comments in this PR and I'll address them later

chlafreniere:
> 

Looks good to me now. Thanks @Charles-Gagnon.

Issue #500, opened 2022-11-14
Add Integration tests for scaling support


Charles-Gagnon:
Currently only have unit tests, we should be actually verifying it works as expected (at the very least ensure the queries can run)

Charles-Gagnon:
@JatinSanghvi Do you have any ideas here? My current plan is just to have a test that runs the scaling queries so we can avoid breaks like happened recently - but that won't actually be verifying scaling itself works. But is that something we can even test with the function core tools? Or will we need to deploy a function to actually verify that full end-to-end scenario?

JatinSanghvi:
The Functions core tools (along with WebJobs SDK) contains the logic to:
- Hosting the HTTP APIs that ScaleController can query to fetch scaling metrics.
- Forwarding the request for metrics to the respective extension.
- Caching the metrics (in storage account table if I am correct) so it can supply a list of metrics every time to the extension and it can know if the load is trending up or down.

So, there should be a way to test the above behaviors with just the core tools. We will need to get help from engineers in Azure Functions team to know if this is feasible without requiring a change in Functions runtime host, and if there is another extension already performing integration testing like this. I am not sure if this will be able to be a significant improvement over unit tests.

The ScaleController is the component responsible for actual scaling the Function App (with or without runtime driven scaling enabled). So, if we want to perform a complete end-to-end integration testing, we will need to have a Function App published, and for each test run, we will need to publish the code to this Function App and generate the load on an external SQL server to ensure that Function App is able to scale in and out. If we have a fixed Function App, we will need to make such test part of periodic (e.g., nightly) testing to ensure that the only one test is running at a time. If we make it part of CI test, then even Function App will need to be created every time, which can add to complexity and test completion time.

Note that the ScaleController does not have deterministic behavior (in my experience). It can sometimes take long intervals between polling for metrics, and sometimes there is a significant time delay after a scale-out is requested by the extension and it actually scales out the app, so we will possibly not be able to add precise assertion-statements (i.e., app scaled out to 4 instances after 2 minutes) in such tests.

Charles-Gagnon:
That's great info, thanks. I'm planning on just keeping what we have for now since it should be fairly certain that the function runtime scaling logic is working correctly (and we have tests that our logic is working) and given the complexity of setting something like that up but I'll keep that in mind for when we're looking to increase automated coverage in the future.

PR #509, opened 2022-11-14
Add metrics integration test

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/500

Help prevent issues like https://github.com/Azure/azure-functions-sql-extension/issues/488

Right now I'm just calling the GetMetrics method directly since there isn't a way (I know of) to test scaling locally. But given how simple the process is this seems fine - it does what it intends and will make sure that any future changes don't break this.

I also changed it to throw an exception if the query fails - silently failing makes it harder to track this stuff down and if a user has scaling enabled and we aren't actually getting valid values (it defaulted to 0) then we shouldn't act like there's no unprocessed changes. 

(note that we already have unit tests to test the other parts of the scaling logic, specifically the stuff that determines scale-in vs scale-out. So that isn't necessary to test here since that stuff is all easily unit-testable)

### DIFF
--- a/test/Integration/SqlTriggerBindingIntegrationTests.cs
+++ b/test/Integration/SqlTriggerBindingIntegrationTests.cs
@@ -14,6 +14,7 @@ using Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
+using Moq;
 using Newtonsoft.Json;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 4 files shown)


test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -471,6 +475,27 @@ public void ChangeTrackingNotEnabledTriggerTest()
...
+        /// <summary>
+        /// Tests that the GetMetrics call works correctly.
+        /// </summary>
+        /// <remarks>We call this directly since there isn't a way to test scaling locally - with this we at least verify the methods called don't throw unexpectedly.</remarks>
+        [Fact]
+        public async void GetMetricsTest()
+        {
+            this.SetChangeTrackingForTable("Products");
+            IConfiguration configuration = new ConfigurationBuilder().Build();
+            var listener = new SqlTriggerListener<Product>(this.DbConnectionString, "dbo.Products", "func-id", new Moq.Mock<ITriggeredFunctionExecutor>().Object, new Moq.Mock<ILogger>().Object, configuration);
```
JatinSanghvi, line 487: Can add namespace `using Moq`. Also `Mock<X>.Object` can be simplified to `Mock.Of<X>()`.

JatinSanghvi (approved):

AmeyaRele (approved):

PR #511, opened 2022-11-16
Start Azurite once before performance tests

Charles-Gagnon:
Was seeing some issues with Azurite getting stuck and causing the performance tests to fail in the pipeline. So to fix that we start up Azurite once before all the tests run (the tests will still each try to start up Azurite but then fail because the ports are in use, but this is fine as long as the first one is running because we currently don't actually check that it started up successfully)

This is just a temporary workaround while I investigate the root cause and other solutions


lucyzhang929 (approved):

Charles-Gagnon:
Going to further investigate the test failure separately - doesn't seem related to this change (and is a failure I've seen before)

Issue #442, opened 2022-11-02
Dotnet Test Fails if integration test fails at Function Host


VasuBhog:
Found this when testing on MacOS and Windows.

If a Integration Test fails due to a Azure Function issue then the process that hosts the Azure Function does not dispose / kill the process prior to starting the next test. 

Therefore, it will fail the rest of the tests after the first failure it finds.

Workaround:
- Comment out tests and focus on one test that is failing
- Run Ad-Hoc test to see all failures

PR #513, opened 2022-11-16
Kill entire process tree after tests

Charles-Gagnon:
We weren't killing the entire process tree of the processes that we start up during tests (azurite and func). This meant that we'd be left with the child processes from those still running (for example, azurite starts as a script which spawns a node.js process) which could impact later tests. 

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill?view=net-7.0#system-diagnostics-process-kill(system-boolean)

### DIFF
--- a/performance/SqlBindingBenchmarks.cs
+++ b/performance/SqlBindingBenchmarks.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Performance
             }
             finally
             {
-                azuriteHost.Kill();
+                azuriteHost.Kill(true);
                 azuriteHost.Dispose();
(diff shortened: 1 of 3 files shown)


Charles-Gagnon:
@VasuBhog This should hopefully help with the issues you were seeing with the function host not being cleaned up properly

lucyzhang929 (approved):

VasuBhog (approved):

VasuBhog:
Will be sure to verify the fix once merged in and then close #442. Thanks, Charles!!

Charles-Gagnon:
Still some more investigation to do with the failures, will keep looking into it.

Issue #507, opened 2022-11-14
Add compat level check on startup


Charles-Gagnon:
We only support compat level 130+, for lower versions users will get weird errors about incorrect syntax which isn't very helpful. We should be doing a check on startup for verifying the current compat level and throwing an error saying that we only support 130+ if a lower version is detected.

PR #514, opened 2022-11-16
Verify database compat level on startup

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/507

This will check the compat level on startup for Trigger and Output bindings (since those use OPENJSON/MERGE statements that aren't supported in earlier versions).

Note that while the trigger binding will error immediately on startup, the Output binding doesn't error until the function is actually triggered since that's the earlier we get the actual attribute that contains the connectionstringsetting value. 

I'm not doing this for input bindings since those queries are controlled by the user so may use whatever syntax they want.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -11,7 +11,6 @@ using System.Runtime.Caching;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using static Microsoft.Azure.WebJobs.Extensions.Sql.Telemetry.Telemetry;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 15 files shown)


src/SqlAsyncCollector.cs
```diff
@@ -91,6 +92,13 @@ public SqlAsyncCollector(IConfiguration configuration, SqlAttribute attribute, I
             this._attribute = attribute ?? throw new ArgumentNullException(nameof(attribute));
             this._logger = logger;
             TelemetryInstance.TrackCreate(CreateType.SqlAsyncCollector);
+            using (SqlConnection connection = BuildConnection(attribute.ConnectionStringSetting, configuration))
```
Charles-Gagnon, line 95: I would have liked to avoid doing this here for perf reasons (avoid opening a new connection if possible) but we don't open the connection until the rows are flushed which may be well after the function actually starts running. But something to consider moving if we think that's not worth the extra overhead.

test/Unit/SqlOutputBindingTests.cs
```diff
@@ -24,23 +24,6 @@ public void TestNullCollectorConstructorArguments()
             Assert.Throws<ArgumentNullException>(() => new SqlAsyncCollector<string>(null, arg, logger.Object));
         }
 
-        [Fact]
-        public async Task TestAddAsync()
```
Charles-Gagnon, line 28: This isn't true anymore so removing the test

PR #516, opened 2022-11-16
tests for out of proc samples

MaddyDev:
I've added the tests under the worker project currently and open to suggestions on a better place. Right now I just have the tests for all the out of proc samples and am still working on the negative test scenarios, will add them in a separate PR. 
I had to duplicate a lot of code from test folder since out of proc needs to be an independent project without the references to the azure web jobs and use the worker stuff instead. I want to refactor them in a way that all the common database and utility code resides in a common project that we can reference in both these test projects but wanted to do it separately to reduce the PR size.


Charles-Gagnon (changes requested):
I would have expected that the tests stay in the test folder - the tests themselves are run separately from the function host so it doesn't matter that it's out of proc. 

Calling `func start` from the output folder of samples-outofproc should get you the function process you need to run the tests against. It'll find the samples DLL with the function definitions and run it from there.

MaddyDev:
> 

For creating the negative scenario tests, I'll need to reference the worker project and I cannot add it to the existing tests since functions throw an error if I have both Microsoft.NET.Sdk.Functions and Microsoft.Azure.Functions.Worker in the references. I need to create a separate project for them. What do you suggest, have the postive test cases along with the others and create a separate project dotnet-isolated-tests under test/ ?

Charles-Gagnon:
Yeah, that sounds reasonable to me. Definitely much easier to work with having all the positive and negative test cases near each other.

Charles-Gagnon:
And to make things easier you could start with getting a PR out for adding the positive cases since those are more straightforward and then do a separate PR for just adding the negative cases. Up to you what you prefer.

Charles-Gagnon (approved):

chlafreniere (approved):

PR #518, opened 2022-11-17
Add integration tests for Java bindings

lucyzhang929:
There are a couple of failing tests that I'm skipping temporarily and will be following up on.


test/Integration/IntegrationTestBase.cs
```diff
@@ -164,6 +166,59 @@ private void ExecuteAllScriptsInFolder(string folder)
...
+            this.LogOutput($"Building Java function app...");
+
+            const int buildJavaAppTimeoutInSeconds = 60;
+            bool isCompleted = taskCompletionSource.Task.Wait(TimeSpan.FromSeconds(buildJavaAppTimeoutInSeconds));
+            Assert.True(isCompleted, "Java function app did not build successfully");
+
+            this.LogOutput("Java build successful!");
+            this.Mvn.OutputDataReceived -= SignalStartupHandler;
+
+            void SignalStartupHandler(object sender, DataReceivedEventArgs e)
```
Charles-Gagnon, line 206: I'd suggest moving this up before you attach it - it's confusing to have it declared after it's used

test/Integration/IntegrationTestBase.cs
```diff
@@ -164,6 +166,59 @@ private void ExecuteAllScriptsInFolder(string folder)
...
+            {
+                if (e.Data?.Contains("BUILD SUCCESS") == true)
+                {
+                    taskCompletionSource.SetResult(true);
+                }
+            };
+            taskCompletionSource.Task.Wait(6000);
```
Charles-Gagnon, line 213: Is this necessary? You're already waiting above for it to finish

test/Integration/IntegrationTestBase.cs
```diff
@@ -40,6 +40,8 @@ public class IntegrationTestBase : IDisposable
         /// </summary>
         private Process AzuriteHost;
 
+        private Process Mvn;
```
Charles-Gagnon, line 43: Let's give this the full name `Maven`

test/Integration/IntegrationTestBase.cs
```diff
@@ -235,14 +311,14 @@ protected void StartFunctionHost(string functionName, SupportedLanguages languag
...
-            functionHost.ErrorDataReceived += this.GetTestOutputHandler(functionHost.Id);
+            functionHost.OutputDataReceived += this.GetTestOutputHandler(functionHost.Id.ToString());
+            functionHost.ErrorDataReceived += this.GetTestOutputHandler(functionHost.Id.ToString());
             functionHost.BeginOutputReadLine();
             functionHost.BeginErrorReadLine();
 
             this.LogOutput("Waiting for Azure Function host to start...");
 
-            const int FunctionHostStartupTimeoutInSeconds = 60;
+            const int FunctionHostStartupTimeoutInSeconds = 120;
```
Charles-Gagnon, line 321: Why is this being increased?

test/Integration/IntegrationTestBase.cs
```diff
@@ -263,7 +339,7 @@ void SignalStartupHandler(object sender, DataReceivedEventArgs e)
                     taskCompletionSource.SetResult(true);
                 }
             };
-            taskCompletionSource.Task.Wait(60000);
+            taskCompletionSource.Task.Wait(120000);
```
Charles-Gagnon, line 342: This is duplicated - we already wait on line 322

test/Integration/IntegrationTestBase.cs
```diff
@@ -311,7 +387,7 @@ private void LogOutput(string output)
             }
         }
 
-        private DataReceivedEventHandler GetTestOutputHandler(int processId)
+        private DataReceivedEventHandler GetTestOutputHandler(string processId = null)
```
Charles-Gagnon, line 390: Why are you not including the pid for the maven process?

And right now this would print out `[] <data here>` for those messages, so even if we did want to allow this to skip the ID then we should also be skipping adding the PID to the message output. (but I think it'd be best to just include the PID for all of them anyways)

test/Integration/test-java/pom.xml
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.function</groupId>
+    <artifactId>test-java</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <name>Azure Java Functions</name>
```
Charles-Gagnon, line 10:
```suggestion
    <name>Azure Java Functions - Integration Tests</name>
```

test/Integration/test-java/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library</artifactId>
+            <version>${azure.functions.java.library.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.microsoft.azure.functions</groupId>
+            <artifactId>azure-functions-java-library-sql</artifactId>
+            <version>0.1.1</version>
```
Charles-Gagnon, line 30: Update the comment next to the version to indicate this needs to be updated as well when we bump that

(also it would definitely be good to be finding a way to just reference the local version of the library instead to avoid having to do that and make it so changes to the library are actually being tested)

test/Integration/IntegrationTestBase.cs
```diff
@@ -164,6 +167,72 @@ private void ExecuteAllScriptsInFolder(string folder)
...
+                this.BuildJavaFunctionApp(mavenPath, testJavaPath);
+            }
+        }
+
+        /// <summary>
+        /// Run `mvn clean package` to build the Java function app.
+        /// </summary>
+        private void BuildJavaFunctionApp(string mavenPath, string workingDirectory)
+        {
+            this.Mvn = new Process()
```
Charles-Gagnon, line 196: This only seems to be used within this method, so is there any reason you need to store it as a local var?

test/Integration/IntegrationTestBase.cs
```diff
@@ -164,6 +167,72 @@ private void ExecuteAllScriptsInFolder(string folder)
...
+            this.Mvn.Start();
+            this.Mvn.OutputDataReceived += this.GetTestOutputHandler();
+            this.Mvn.ErrorDataReceived += this.GetTestOutputHandler();
+            this.Mvn.BeginOutputReadLine();
+            this.Mvn.BeginErrorReadLine();
+
+            this.LogOutput($"Building Java function app...");
+
+            const int buildJavaAppTimeoutInSeconds = 60;
+            bool isCompleted = taskCompletionSource.Task.Wait(TimeSpan.FromSeconds(buildJavaAppTimeoutInSeconds));
```
Charles-Gagnon, line 229: Instead of waiting for a build message couldn't you just wait for the process to exit and then check the exit code? That'd be easier/safer in case the message changes for some reason (and also guarantees that the process will have actually exited).

You can still have a timeout on the wait to ensure we aren't waiting for a ridiculously long time.

This also would mean you don't have to redirect the stdout which means you can use shell execute and get rid of the requirement to have MAVEN_PATH defined.

Charles-Gagnon (changes requested):

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -248,6 +248,18 @@ steps:
     projects: '${{ parameters.solution }}'
     arguments: '--configuration ${{ parameters.configuration }} --output $(Build.ArtifactStagingDirectory) --no-build -p:PackageVersion="${{ parameters.nugetVersion }}" -p:GeneratePackageOnBuild=false'
 
+- task: Maven@3
+  displayName: Build Java Samples
+  inputs:
+    mavenPomFile: samples/samples-java/pom.xml
+    options: --batch-mode --define=gpg.skip --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn --update-snapshots
```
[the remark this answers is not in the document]
lucyzhang929: These were options I copied over from the RabbitMq's build pipeline (https://github.com/Azure/azure-functions-rabbitmq-extension/blob/aa40eb2341cb1d861601a8ddbdfdd1cc03f1c343/azure-pipelines.yml#L179). These options aren't needed when building locally, apart from `--update-snapshots` or `-U` which tells maven not to use the cached dependencies.

test/Integration/IntegrationTestBase.cs
lucyzhang929: Whoops, meant to change that back.

test/Integration/test-java/pom.xml
lucyzhang929: Tracking issue here: https://github.com/Azure/azure-functions-sql-extension/issues/525

test/Integration/IntegrationTestBase.cs
```diff
@@ -76,6 +76,7 @@ public IntegrationTestBase(ITestOutputHelper output = null)
             this.TestOutput = output;
             this.SetupDatabase();
             this.StartAzurite();
+            this.BuildJavaFunctionApps();
```
Charles-Gagnon, line 79: We shouldn't do this for every test class - you can add it to the test fixture I'm adding in https://github.com/Azure/azure-functions-sql-extension/pull/527 once that's in

test/Integration/IntegrationTestBase.cs
```diff
@@ -164,6 +165,56 @@ private void ExecuteAllScriptsInFolder(string folder)
...
+        /// <summary>
+        /// Run `mvn clean package` to build the Java function app.
+        /// </summary>
+        private void BuildJavaFunctionApp(string mavenPath, string workingDirectory)
+        {
+            var maven = new Process()
+            {
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = mavenPath,
```
Charles-Gagnon, line 198: You shouldn't need to specify the full path to maven now that you're using shell execute. (hopefully at least, this is assuming it's being added to the path and whatnot...)

Try it out and see if we can get rid of all this extra path logic and make things a lot simpler.

builds/azure-pipelines/template-steps-build-test.yml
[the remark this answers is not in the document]
lucyzhang929: Actually I don't think the `--define=gpg.skip` option is needed at all. gpg is a signing plugin (https://maven.apache.org/plugins/maven-gpg-plugin/usage.html) but we aren't using it. I'll remove that here and in the java-library and run an ad hoc build to make sure nothing breaks.

Based on this [post](https://stackoverflow.com/questions/21638697/disable-maven-download-progress-indication), --batch-mode and the simpleLogger are used to make the build logging less verbose.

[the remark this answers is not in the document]
lucyzhang929: https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=180202&view=logs&s=843cb16e-c119-5eb8-2e63-396776413ea7

The java build was able to complete successfully after removing `--define=gpg.skip`

builds/azure-pipelines/build-release-java.yml
```diff
@@ -3,7 +3,7 @@ steps:
     displayName: Build library
     inputs:
       mavenPomFile: java-library/pom.xml
-      options: --batch-mode --define=gpg.skip --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn --update-snapshots
+      options: --batch-mode --define=org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn --update-snapshots
```
Charles-Gagnon, line 6: Would suggest adding a comment here explaining what the options specified are doing (in the other places too). It's a bit verbose but will make sure there isn't any confusion later on about why they're being used.

test/Integration/IntegrationTestFixture.cs
```diff
@@ -41,6 +45,44 @@ protected void StartAzurite()
...
+            string testJavaPath = Path.Combine(TestUtils.GetPathToBin(), "..", "..", "..", "Integration", "test-java");
+            BuildJavaFunctionApp(testJavaPath);
+        }
+
+        /// <summary>
+        /// Run `mvn clean package` to build the Java function app.
+        /// </summary>
+        private static void BuildJavaFunctionApp(string workingDirectory)
+        {
+            var maven = new Process()
```
Charles-Gagnon, line 65: And actually do we need this if the pipelines are already doing the build? If we want to have a way for people to easily build latest + run tests locally then generally it'd be good to do that in a script instead (since building can take a while and often times isn't actually needed)

Charles-Gagnon (approved):
Few last things to look at but things mostly look fine!

PR #526, opened 2022-11-18
Copy sql dll to extension bundle during Integration tests

lucyzhang929:
Moving this step from the pipeline to IntegrationTestBase so that we don't have to manually copy the dll to the bundle when running integration tests locally.

I couldn't find a way to specify the extension bundle path when running func start. The instructions for testing local extension bundles also use the copy method: https://github.com/Azure/azure-functions-extension-bundles#test-an-extension-bundle

### DIFF
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -121,6 +121,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
                 }
             };
             getExtensionBundlePath.OutputDataReceived += ExtensionBundlePathHandler;
+            getExtensionBundlePath.OutputDataReceived += this.GetTestOutputHandler(maven.Id);
+            getExtensionBundlePath.ErrorDataReceived += this.GetTestOutputHandler(maven.Id);
             getExtensionBundlePath.Start();
             getExtensionBundlePath.BeginOutputReadLine();
             this.LogOutput("Getting extension bundle path...");
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -74,8 +74,65 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         public IntegrationTestBase(ITestOutputHelper output = null)
         {
             this.TestOutput = output;
+            this.CopyLatestSqlExtensionToBundle();
             this.SetupDatabase();
             this.StartAzurite();
+        }
+
+        /// <summary>
+        /// Copy the latest build of the SqlExtension dll to the Azure Functions extension bundle.
+        /// </summary>
+        private void CopyLatestSqlExtensionToBundle()
+        {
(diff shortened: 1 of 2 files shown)


test/Integration/IntegrationTestBase.cs
```diff
@@ -74,10 +74,69 @@ public class IntegrationTestBase : IDisposable
...
         }
 
+        /// <summary>
+        /// Copy the latest build of the SqlExtension dll to the Azure Functions extension bundle.
+        /// </summary>
+        private void CopyLatestSqlExtensionToBundle()
+        {
+            string extensionBundleBinPath = Path.Combine(this.GetExtensionBundlePath(), "bin");
+            string sqlDllPath = Path.Combine(GetPathToBin(), "Microsoft.Azure.WebJobs.Extensions.Sql.dll");
+            File.Copy(sqlDllPath, Path.Combine(extensionBundleBinPath, "Microsoft.Azure.WebJobs.Extensions.Sql.dll"), true);
```
Charles-Gagnon, line 89: The more I think about it the less I like doing this - silently overwriting bundle contents could easily cause issues when trying to test/verify stuff locally. Especially for people who may not know this happened - if someone comes in and runs the tests they could easily get in a bad state and not have any idea what happened. 

What about instead talking with the core tools team and seeing if they'd be open to adding an environment  variable and/or command line parameter to specify what bundle to use? This would even be helpful for testing bundles in general so seems like an overall good thing to add.

And in the meantime I feel like having the manual instructions might actually be fine. Or another couple options that would make this better : 

1. Have a script that users can run that copies it over - this way it's a deliberate action so they're aware its happening
2. Regardless, having it copy the original DLL to *.bak so it's easy to go back isn't a bad idea (you can also just delete the entire folder - that's what I do now - but it wouldn't be a bad idea regardless so it's easy to see when it's been replaced)

Charles-Gagnon (changes requested):

test/Integration/IntegrationTestBase.cs
lucyzhang929: I agree, I'll reach out to the core tools team.

In the meantime I'll add a section to our test readme and create another PR for adding a script.

VasuBhog:
I want to say that I believe this would be quite beneficial to local development. As I am currently having to run the script constantly every time I make changes to the C# code. Would like to know why we would not generally test the latest dll version for non-C# languages? Having this would just enhance our developing experience while the core tools team may potentially add an environment variable.

lucyzhang929:
I think the concern is for folks who are unaware of the extension bundle being modified while running tests and causing confusion with behavior differences between their local sql dll and the sql dll shipped with the bundle.

Charles-Gagnon:
Correct. And since we don't revert it (and couldn't 100% guarantee it would be reverted) this could mean that people running the tests end up with a potentially broken version of the extension in their bundle forever until they manually fix it. 

My ideal scenario would be that the func CLI allow specifying a custom location for the bundle package so that we could use a temp one during tests. But that isn't currently supported unfortunately.

Issue #459, opened 2022-11-07
Automatically copy locally built DLLs 


VasuBhog:
In order to test scripting languages (PowerShell, Javascript and etc) with the latest Dll builds, one must (**currently**) do the following:

1. Run dotnet build in the root of the repo
2. Copy the DLL in src/bin/Debug/netstandard2.0 to the extension bundle (samples you want to test - step 3).
3. Go to the directory of the extension bundle you want to change - for example azure-functions-sql-extension/samples/samples-powershelland usefunc GetExtensionBundlePath` to see where to copy the DLL found in step 2 into.
4. Copy the DLL from step 2 to the path found in step 3 and run Dotnet test at the root of the repo as it should use the latest builds then.

It would be better to make this an automated task - possibly as embedded as a startup task in the tests themselves.

Charles-Gagnon:
@VasuBhog Please add the manual steps to https://github.com/Azure/azure-functions-sql-extension/blob/main/test/README.md

PR #529, opened 2022-11-21
Add script to copy sql dll to extension bundle

lucyzhang929:
fixes https://github.com/Azure/azure-functions-sql-extension/issues/459

### DIFF
--- a/scripts/CopySqlDllToExtensionBundle.ps1
+++ b/scripts/CopySqlDllToExtensionBundle.ps1
@@ -1,0 +1,11 @@
+# Run this script from the root of the repository to copy the latest
+# SQL extension dll to the extension bundle.
+
+# Run `func GetExtensionBundlePath` from samples-js to get the correct path to the extension bundle.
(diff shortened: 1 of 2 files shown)


scripts/CopySqlDllToExtensionBundle.ps1
```diff
@@ -0,0 +1,11 @@
+# Run this script from the root of the repository to copy the latest
+# SQL extension dll to the extension bundle.
```
Charles-Gagnon, line 2:
```suggestion
# Run this script from the root of the repository to copy the latest
# SQL extension dll to the extension bundle. This is required to test
# the latest extension changes when running samples/test.
#
# To revert this fun `func GetExtensionBundlePath` manually from the
# `samples-js` folder and delete the folder it returns. The next time a
# function referencing the bundle is launched the host will download 
# a clean copy of the bundle to use.
```

Charles-Gagnon (approved):

Issue #443, opened 2022-11-02
PowerShell issue - empty string is upserted as NULL in DB


VasuBhog:
Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration.SqlOutputBindingIntegrationTests.AddProductParamsTest(id: 0, name: "", cost: 0, lang: PowerShell) [FAIL]

When executing scalar select Name from Products where ProductId=0
Expected:  (System.String)
Actual:    (System.DBNull)

This seems to be due PowerShell worker that treats empty string as Null.

VasuBhog:
Fixed by a workaround, but further tracked here: https://github.com/Azure/azure-functions-sql-extension/issues/535

PR #531, opened 2022-11-22
PowerShell NULL and function file updates

VasuBhog:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/443. As I found a current workaround using `$TriggerMetadata` and followed up with a tracking issue in the powershell worker repo [here](https://github.com/Azure/azure-functions-powershell-worker/issues/895).

I also discussed with Francisco from the PowerShell team that we should use the latest worker version and include `$TriggerMetadata` in our samples as it can be used instead of $Request as it is given as default in powershell azure functions. I added casting to some of the PowerShell functions in order to match precisely the same way we set them up from C# samples.

I have root caused the issue with the `AddProductWithIdentity` tests as there seems to be an issue with the GenerateDataQueryForMerge as the tests seem to generate the `INSERT INTO` values in the wrong order as how they are defined in the table. As such, there is a failure during the dotnet test. I will address this in the next PR.


chlafreniere:
@VasuBhog please split this out into 2 different PRs

VasuBhog:
> @VasuBhog please split this out into 2 different PRs

Mac PR: https://github.com/Azure/azure-functions-sql-extension/pull/533

samples/samples-powershell/AddProductParams/run.ps1
```diff
@@ -1,16 +1,18 @@
 using namespace System.Net
 
 # Trigger binding data passed in via param block
-param($Request)
+param($Request, $TriggerMetadata)
```
Charles-Gagnon, line 4: Can you just remove $Request from the parameters? Since we're not even using them that would be best if that's allowed.

samples/samples-powershell/AddProductWithIdentityColumnIncluded/run.ps1
```diff
@@ -1,16 +1,16 @@
...
-param($Request)
+param($Request, $TriggerMetadata)
 
 # Write to the Azure Functions log stream.
 Write-Host "PowerShell function with SQL Output Binding processed a request."
 
 # Update req_query with the body of the request
 $req_query = @{
-    productId= if($Request.QUERY.productId) { $Request.QUERY.productId } else { $null };
+    productId= if($Request.QUERY.productId) { [int]$Request.QUERY.productId } else { $null };
```
Charles-Gagnon, line 11: Why are you casting these to ints? Your description doesn't say anything about these changes.

test/Integration/test-powershell/AddProductColumnTypes/run.ps1
```diff
@@ -10,9 +10,9 @@ Write-Host "PowerShell function with SQL Output Binding processed a request."
 # Note that this expects the body to be a JSON object or array of objects 
 # which have a property matching each of the columns in the table to upsert to.
 $req_query = @{ 
-    ProductId=$Request.QUERY.productId;
-    Datetime=Get-Date -AsUTC;
-    Datetime2=Get-Date -AsUTC;
+    ProductId=[int]$Request.QUERY.productId;
+    Datetime=Get-Date;
```
Charles-Gagnon, line 14: Why are you removing AsUTC?

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -47,9 +47,9 @@ public void AddProductTest(int id, string name, int cost, SupportedLanguages lan
         [SqlInlineData(1, "Test", 5)]
         [SqlInlineData(0, "", 0)]
         [SqlInlineData(-500, "ABCD", 580)]
-        // Currently PowerShell and Java functions return null when the parameter for name is an empty string
-        // Issue link: https://github.com/Azure/azure-functions-sql-extension/issues/443
-        [UnsupportedLanguages(SupportedLanguages.PowerShell, SupportedLanguages.Java)]
+        // Add Support to nullable types in query string to be added in the powershell worker
+        // Currently use a workaround through the TriggerMetadata for PowerShell Azure Functions
+        // Issue link: https://github.com/Azure/azure-functions-powershell-worker/issues/895
```
Charles-Gagnon, line 52: We can just remove this since you already have the comment in the function itself
```suggestion
```

samples/samples-powershell/AddProductParams/run.ps1
```diff
@@ -1,16 +1,18 @@
...
 
 # Write to the Azure Functions log stream.
 Write-Host "PowerShell function with SQL Output Binding processed a request."
 
 # Update req_query with the query of the request
+# Currently use a workaround through the TriggerMetadata for PowerShell Azure Functions
+# Issue link: https://github.com/Azure/azure-functions-powershell-worker/issues/895
```
Charles-Gagnon, line 11: Please also make a tracking issue in our repo to remove this workaround once they've fixed it in the worker.

samples/samples-powershell/AddProductParams/run.ps1
```diff
@@ -1,16 +1,18 @@
...
 
 # Trigger binding data passed in via param block
-param($Request)
+param($Request, $TriggerMetadata)
 
 # Write to the Azure Functions log stream.
 Write-Host "PowerShell function with SQL Output Binding processed a request."
 
 # Update req_query with the query of the request
+# Currently use a workaround through the TriggerMetadata for PowerShell Azure Functions
```
Charles-Gagnon, line 10:
```suggestion
# Currently the Powershell worker does not allow empty/null values to be passed through the
# query parameters. We use TriggerMetadata here as a workaround for that issue. 
```

samples/samples-powershell/AddProductWithIdentityColumnIncluded/run.ps1
VasuBhog: Updated description to address this.

samples/samples-powershell/AddProductParams/run.ps1
VasuBhog: Tracked here: https://github.com/Azure/azure-functions-sql-extension/issues/535

test/Integration/test-powershell/AddProductColumnTypes/run.ps1
VasuBhog: Forgot to revert this for now. Will update.

samples/samples-powershell/AddProductWithIdentityColumnIncluded/run.ps1
Charles-Gagnon: So just to be clear - this isn't actually necessary? You're just doing it to be consistent?

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -47,9 +47,6 @@ public void AddProductTest(int id, string name, int cost, SupportedLanguages lan
         [SqlInlineData(1, "Test", 5)]
         [SqlInlineData(0, "", 0)]
         [SqlInlineData(-500, "ABCD", 580)]
-        // Currently PowerShell and Java functions return null when the parameter for name is an empty string
-        // Issue link: https://github.com/Azure/azure-functions-sql-extension/issues/443
-        [UnsupportedLanguages(SupportedLanguages.PowerShell, SupportedLanguages.Java)]
```
Charles-Gagnon, line 52: These are still failing for Java it seems. 

@lucyzhang929 - possibly a similar issue for their worker. So something to follow up with them about when you get around to investigating this.

Charles-Gagnon (commented):
You still have one failing powershell test too

samples/samples-powershell/AddProductWithIdentityColumnIncluded/run.ps1
VasuBhog: Yes, this is to be consistent with testing. These tests are currently not being run due to them being unstable. For now, I would like to have it be exactly the same as C# in order to reduce the variables of why it was not working.

lucyzhang929 (approved):

Charles-Gagnon (approved):

PR #533, opened 2022-11-22
Fix mac integration function tools path

VasuBhog:
If users on Mac have Azure Function core tools installed via [Brew](https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v4%2Cmacos%2Ccsharp%2Cportal%2Cbash#install-the-azure-functions-core-tools). Then dotnet test will constantly fail.

### DIFF
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -275,10 +275,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
                     }
                     throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath} or {usrBinFuncPath}");
                 }
-                else
-                {
-                    throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath}");
-                }
                 throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath}");
             }
 
--- a/test/Integration/IntegrationTestBase.cs
+++ b/test/Integration/IntegrationTestBase.cs
@@ -265,6 +265,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
                     }
                     throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath} or {programFilesFuncPath}");
                 }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    // Search Mac to see if brew installed location has azure function core tools
+                    string usrBinFuncPath = Path.Combine("/usr", "local", "bin", "func");
+                    if (File.Exists(usrBinFuncPath))
+                    {
+                        return usrBinFuncPath;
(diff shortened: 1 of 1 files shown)


test/Integration/IntegrationTestBase.cs
```diff
@@ -265,6 +265,20 @@ private static string GetFunctionsCoreToolsPath()
...
+                    if (File.Exists(usrBinFuncPath))
+                    {
+                        return usrBinFuncPath;
+                    }
+                    throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath} or {usrBinFuncPath}");
+                }
+                else
+                {
+                    throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath}");
+                }
```
Charles-Gagnon, line 281: This isn't needed, line 282 already throws if none of the if's work
```suggestion
```

test/Integration/IntegrationTestBase.cs
```diff
@@ -265,6 +265,20 @@ private static string GetFunctionsCoreToolsPath()
                     }
                     throw new FileNotFoundException($"Azure Function Core Tools not found at {funcPath} or {programFilesFuncPath}");
                 }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
chlafreniere, line 268: Does OSPlatform only has OSX for mac? Just making sure that it's inclusive for macos11/12/etc

VasuBhog: There seems to be `FreeBSD, Linux, OSX, Windows, and Create OS` as options for OSPlatform. 

I believe that OSX should be inclusive of any version of Mac.

chlafreniere:
@VasuBhog I tried pulling down the branch, had a connection string that used SQL Auth, and got an error about kerberos:

```
Cannot authenticate using Kerberos. Ensure Kerberos has been initialized on the client with 'kinit' and a Service Principal Name has been registered for the SQL Server to allow Kerberos authentication.
```

Is this a known issue?

VasuBhog:
> ```
> Cannot authenticate using Kerberos. Ensure Kerberos has been initialized on the client with 'kinit' and a Service Principal Name has been registered for the SQL Server to allow Kerberos authentication.
> ```

Does not seem to be an issue that's tracked in the powershell worker repo. So please create an issue and then can follow up on investigating it.

chlafreniere:
> > ```
> > Cannot authenticate using Kerberos. Ensure Kerberos has been initialized on the client with 'kinit' and a Service Principal Name has been registered for the SQL Server to allow Kerberos authentication.
> > ```
> 
> 
> Does not seem to be an issue that's tracked in the powershell worker repo. So please create an issue and then can follow up on investigating it.

Ah, user error, didn't have the SA_PASSWORD environment variable set.

Charles-Gagnon (approved):

chlafreniere (approved):

PR #538, opened 2022-11-28
Remove java build step in integration test

lucyzhang929:


MaddyDev (approved):

test/README.md
```diff
@@ -35,6 +35,9 @@ Our integration tests are based on functions from the samples project. To run in
 ### Extension Bundle
    If you've made any changes to the CSharp extension, you will need to run the CopySqlDllToExtensionBundle.Ps1 to update your local extension bundle so that the tests can run against your latest changes.
 
+### Running Java Tests
+   Run `mvn clean package` from the samples/samples-java and test/Integration/test-java folders to build the Java Azure Functions before running the tests.
```
Charles-Gagnon, line 39: It'd be great to add a script that does this to make it a one step process (I'd even have it build both and then run the tests to make it more straightforward)

test/Integration/IntegrationTestFixture.cs
```diff
@@ -28,13 +28,9 @@ public class BaseTestFixture : IDisposable
         /// </summary>
         private Process AzuriteHost;
 
-        public BaseTestFixture(bool buildJava)
+        public BaseTestFixture()
```
Charles-Gagnon, line 31: If you have time - I want to revert the changes made to split up the fixture classes since it's not necessary anymore (so just one IntegrationTestFixture in this file). If you're busy though go ahead with this and I'll fix it up myself after this gets in.

lucyzhang929: I can remove that

Charles-Gagnon (approved):

PR #541, opened 2022-11-29
Add debugging tests section

Charles-Gagnon:
Add section with instructions for how to debug tests. I've tested it all using just VS Code so it should be applicable to everybody - although I personally prefer to use VS for debugging C# code and running tests.

There's probably a lot of improvements we could make here - for example looking into getting top-level launch targets for running the functions so we don't have to open a new instance of VS Code. So if anyone has time to look into that while you're investigating your own tests (@VasuBhog) that would be fantastic. Otherwise I'll make other improvements as I have time.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,6 +13,12 @@
             "request": "attach",
             "port": 9229,
             "preLaunchTask": "func: host start"
+        },
+        {
+            "name": "Attach to Function Host",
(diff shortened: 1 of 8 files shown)


Charles-Gagnon:
@VasuBhog @MaddyDev @lucyzhang929 Please try these instructions out when you get a chance to make sure I didn't miss anything

Issue #542, opened 2022-11-30
Foreign keys require specific output binding ordering


jnsvd:
When 2 records are created through an output binding where a foreign key constraint exists between their tables, an exception is thrown if the order of the `bindings` does not respect the order in which the records should be created.

**Test setup**

Using the following tables:

```
CREATE TABLE TransactionData
(
    TransactionDataID uniqueidentifier NOT NULL,
    Data              nvarchar(max)    NOT NULL,
    CONSTRAINT PK_TransactionData PRIMARY KEY (TransactionDataID)
);

CREATE TABLE Transaction
(
    TransactionID     uniqueidentifier NOT NULL,
    Timestamp         datetime         NOT NULL,
    TransactionDataID uniqueidentifier NOT NULL,
    CONSTRAINT PK_Transaction PRIMARY KEY (TransactionID),
    CONSTRAINT FK_Transaction_TransactionData FOREIGN KEY (TransactionDataID) REFERENCES TransactionData (TransactionDataID)
)
```

with the following in `function.json`:

```
{
  "bindings": [
    {
      "name": "transaction",
      "type": "sql",
      "direction": "out",
      "commandText": "dbo.[Transaction]",
      "connectionStringSetting": "SqlConnectionString"
    },
    {
      "name": "transactionData",
      "type": "sql",
      "direction": "out",
      "commandText": "dbo.[TransactionData]",
      "connectionStringSetting": "SqlConnectionString"
    }
  ]
}
```

and the following in `index.js`:

```
const crypto = require('crypto');

module.exports = async function (context) {
  const TransactionDataID = crypto.randomUUID();
  context.bindings.transactionData = {
    TransactionDataID,
    Data: 'Some transaction data',
  };
  context.bindings.transaction = {
    TransactionID: crypto.randomUUID(),
    Timestamp: new Date().toISOString().slice(0, 19).replace('T', ' '),
    TransactionDataID,
  }
}
```

**Observations**

The following exception is thrown:

```
[2022-11-30T14:47:18.776Z] Executed 'Functions.SendTransaction' (Failed, Id=7a9cfde4-0bac-4f1b-8714-1769af1ca772, Duration=4345ms)
[...]

PR #551, opened 2022-12-01
Add output binding order note to known issues

Charles-Gagnon:
Closes https://github.com/Azure/azure-functions-sql-extension/issues/542

I can't think of any way we could fix this in our extension so just adding it to the known issues list. Couple of potential workarounds, although the best ones involve using a .NET function.


lucyzhang929 (approved):

chlafreniere:
@dzsquared just making sure that you saw this as FYI

dzsquared:
@chlafreniere thank you, I will get it added to the MS Learn docs as well

Issue #521, opened 2022-11-17
java.sql.Timestamp object gets serialized with an extra comma on Linux


lucyzhang929:
When using java.sql.Timestamp, AddProductColumnTypes test fails on the linux pipeline because the timestamp object being sent to the database had an extra comma between the date and the time.
Mac/Windows: "Dec 1, 2022 11:55:45 AM"
Linux: "Dec 1, 2022, 4:45:52 PM"

`System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Microsoft.Azure.WebJobs.Extensions.Sql: Unexpected error upserting rows. Core Microsoft SqlClient Data Provider: The conversion of a nvarchar data type to a datetime data type resulted in an out-of-range value.`

This depends on the Java worker adding support for customizing the GSON serializer: https://github.com/Azure/azure-functions-java-worker/issues/424.

PR #552, opened 2022-12-01
Fix AddProductColumnTypesTest for Java

lucyzhang929:
Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/521

The test was failing on the linux pipeline because the timestamp object being sent to the database had an extra comma between the date and the time.
Mac/Windows: "Dec 1, 2022 11:55:45 AM"
Linux: "Dec 1, 2022, 4:45:52 PM"

Using java.sql.Date instead of java.sql.Timestamp fixes the issue.


Charles-Gagnon (approved):
It'd be good to add this to the known issues section of the README

MaddyDev (approved):

PR #553, opened 2022-12-01
Bump packages

Charles-Gagnon:
Just minor bumps where available to keep up with latest fixes/improvements. Main driver here was to get the latest Newtonsoft package so there aren't any potential conflicts with the Logic Apps usage. 

Small change to the telemetry config too - directly setting the key is now deprecated : https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560

### DIFF
--- a/samples/samples-csharp/local.settings.json
+++ b/samples/samples-csharp/local.settings.json
@@ -3,7 +3,7 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "SqlConnectionString": "",
+    "SqlConnectionString": "Data Source=localhost;Initial Catalog=Products;User ID=sa;Password=Yukon900;Encrypt=false",
     "Sp_SelectCost": "SelectProductsCost",
     "ProductCost": 100
   }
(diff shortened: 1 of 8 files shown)


samples/samples-csharp/local.settings.json
```diff
@@ -3,7 +3,7 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "SqlConnectionString": "",
+    "SqlConnectionString": "Data Source=localhost;Initial Catalog=Products;User ID=sa;Password=Yukon900;Encrypt=false",
```
lucyzhang929, line 6: Revert connection string

test/Integration/IntegrationTestBase.cs
```diff
@@ -107,7 +107,7 @@ private void SetupDatabase()
 
                 // Either use integrated auth or SQL login depending if SA_PASSWORD is set
                 string userId = "SA";
-                string password = Environment.GetEnvironmentVariable("SA_PASSWORD");
+                string password = "Yukon900"; // Environment.GetEnvironmentVariable("SA_PASSWORD");
```
lucyzhang929, line 110: Revert password

Issue #554, opened 2022-12-01
GetProductsColumnTypesSerializationAsyncEnumerableTest failing for OutOfproc


MaddyDev:
I was able to run the function successfully but when running through the test framework, I see the below error:
System.Private.CoreLib: Exception while executing function: Functions.GetProductsColumnTypesSerializationAsyncEnumerable. System.Private.CoreLib: Result: Failure
[19052] Exception: System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)

Issue #550, opened 2022-12-01
Enable TimerTrigger test for OOP


MaddyDev:
Currently the TimerTrigger sample works as expected but the test keeps failing with the below error:

No job functions found. Try making your job classes and methods public. If you're using binding extensions (e.g. Azure Storage, ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. builder.AddAzureStorage(), builder.AddServiceBus(), builder.AddTimers(), etc.).

Charles-Gagnon:
Are you looking into this builder.AddTimers() call? Seems like a good place to start

Issue #549, opened 2022-12-01
Add QueueTrigger sample and test for OOP


MaddyDev:
Seeing the below error when running QueueTrigger for OOP
Error response [c6c156fb-abc8-4936-ad7c-9caaa94445a1] 404 The specified queue does not exist. (00.0s). 

![image](https://user-images.githubusercontent.com/12754347/205155669-2326ada0-0a36-4394-bc19-5d23eb22bf67.png)

Will sync with Fabio to enable this.

MaddyDev:
Fixed by #577

Issue #548, opened 2022-12-01
AddProductColumnTypesTest failing for OutOfproc


MaddyDev:
Test fails with internal server error: System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Core Microsoft SqlClient Data Provider: Conversion failed when converting date and/or time from character string.

PR #555, opened 2022-12-01
Add more tests for OOP

MaddyDev:
I've added all the negative scenario tests that are included for the rest of the languages. Created the below issues to track the few that have been disabled for OOP:

https://github.com/Azure/azure-functions-sql-extension/issues/550
https://github.com/Azure/azure-functions-sql-extension/issues/549
https://github.com/Azure/azure-functions-sql-extension/issues/548
https://github.com/Azure/azure-functions-sql-extension/issues/554


test-outofproc/AddProductColumnTypes.cs
```diff
@@ -0,0 +1,38 @@
...
+        [Function(nameof(AddProductColumnTypes))]
+        [SqlOutput("dbo.ProductsColumnTypes", ConnectionStringSetting = "SqlConnectionString")]
+        public static ProductColumnTypes Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "addproduct-columntypes")] HttpRequestData req)
+        {
+            NameValueCollection queryStrings = HttpUtility.ParseQueryString(req.Url.Query);
+            var product = new ProductColumnTypes()
+            {
+                ProductID = int.Parse(queryStrings["productId"], null),
+                Datetime = DateTime.UtcNow.Date,
```
lucyzhang929, line 29: Does the test pass using DateTime.UtcNow.Date?

MaddyDev: It doesn't on my local machine, it still throws the conversion error but forgot to revert back that change and it passed on the pipeline!

samples/samples-outofproc/GlobalSuppressions.cs
```diff
@@ -18,4 +18,7 @@
...
+[assembly: SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "Unused parameter is required by functions binding", Scope = "member", Target = "~M:Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.OutputBindingSamples.TimerTriggerProducts.Run(Microsoft.Azure.Functions.Worker.TimerInfo,Microsoft.Azure.Functions.Worker.FunctionContext})~System.Collections.Generic.IEnumerable{Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.Common.Product}")]
+//[assembly: SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "Unused parameter is required by functions binding", Scope = "member", Target = "~M:Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.OutputBindingSamples.QueueTriggerProducts.Run(System.String})~System.Collections.Generic.IEnumerable{Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.Common.Product}")]
```
Charles-Gagnon, line 23:
```suggestion
```

test-outofproc/AddProductColumnTypes.cs
Charles-Gagnon: Excluded what? The test seems to be running against OOP here from what I can tell

test-outofproc/host.json
```diff
@@ -0,0 +1,11 @@
...
+    "version": "2.0",
+    "logging": {
+        "applicationInsights": {
+            "samplingSettings": {
+                "isEnabled": true,
+                "excludedTypes": "Request"
+            }
+        }
+    }
+}
```
Charles-Gagnon, line 11:
```suggestion
{
  "version": "2.0",
  "logging": {
    "applicationInsights": {
      "samplingSettings": {
        "isEnabled": true,
        "excludedTypes": "Request"
      }
    },
    "logLevel": {
      "default": "Information"
    }
  }
}

```

test-outofproc/test-outofproc.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net6</TargetFramework>
```
Charles-Gagnon, line 3: This should already be from the root Directory.Build.props

```suggestion
```

test-outofproc/GetProductColumnTypesSerializationAsyncEnumerable.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace DotnetIsolatedTests
+{
+    public static class GetProductsColumnTypesSerializationAsyncEnumerable
+    {
+        /// <summary>
+        /// This function verifies that serializing an item with various data types
+        /// and different languages works when using IAsyncEnumerable.
+        /// </summary>
+        [Function(nameof(GetProductsColumnTypesSerializationAsyncEnumerable))]
+        public static async Task<List<ProductColumnTypes>> Run(
```
Charles-Gagnon, line 21: It looks like this is only fixing the Output binding tests, please remove any functions that aren't currently being used (such as this one) or update the tests to run against these ones too.

Charles-Gagnon: Oh I missed your comment in the description. I guess this is fine - but in the future it'd be best to just avoid checking in code we know isn't working correctly (even if it's just test code) and instead just keeping a feature branch or something that you can use when working on fixing it. So if you want to do that now go ahead, but it's ok if you leave this in here for now then too (assuming you'll be working on fixing those right away)

test-outofproc/AddProductColumnTypes.cs
MaddyDev: Oh I just added it back to check if it's passing on the pipeline after I commented here.

samples/samples-outofproc/GlobalSuppressions.cs
Charles-Gagnon: Ping @MaddyDev

PR #558, opened 2022-12-05
Telemetry cleanup

Charles-Gagnon:
1. Fix small bug where if the client wasn't created successfully we could cause an exception to escape
2. Log event failures using logger instead of just debug so we can see when failures actually happen in the logs
3. Add startup message with the session ID so we can correlate from the logs to the events we get (set as debug since this would only be needed in very specific debugging scenarios)

### DIFF
--- a/src/Telemetry/Telemetry.cs
+++ b/src/Telemetry/Telemetry.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


MaddyDev (approved):

lucyzhang929 (approved):

chlafreniere (approved):

Charles-Gagnon:
Failure is due to timeout - not stuck just taking longer than default 60min to run. I'll be sending out a fix for that.

Issue #411, opened 2022-10-19
Revert support for case insensitive comparison of POCO fields and table column names


lucyzhang929:
Original PR: https://github.com/Azure/azure-functions-sql-extension/pull/191
Follow up PR: https://github.com/Azure/azure-functions-sql-extension/pull/235/files

PR #560, opened 2022-12-05
Revert support for case insensitive comparison of POCO fields and table column names

lucyzhang929:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/411
- most changes are updating casing in doc, samples, and tests
- main change is in `SqlAsyncCollector.cs` where I reverted changes from https://github.com/Azure/azure-functions-sql-extension/pull/191 and https://github.com/Azure/azure-functions-sql-extension/pull/235
- enabled GetProductsArrayTest for Java (previously failing because of mismatched cases)


chlafreniere:
@lucyzhang929 @Charles-Gagnon is this something that we'd like to get in for the next release?

Charles-Gagnon:
Yeah, I think we should, if anything just to have as much time as possible with people using it.

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -370,48 +366,6 @@ public void AddProductWithIdentity_MissingPrimaryColumn(SupportedLanguages lang)
...
-        {
-            // Set table info cache timeout to 0 minutes so that new collation gets picked up
-            var environmentVariables = new Dictionary<string, string>()
-            {
-                { "AZ_FUNC_TABLE_INFO_CACHE_TIMEOUT_MINUTES", "0" }
-            };
-            this.StartFunctionHost(nameof(AddProductParams), lang, false, null, environmentVariables);
-
-            // Change database collation to case sensitive
-            this.ExecuteNonQuery($"ALTER DATABASE {this.DatabaseName} SET Single_User WITH ROLLBACK IMMEDIATE; ALTER DATABASE {this.DatabaseName} COLLATE Latin1_General_CS_AS; ALTER DATABASE {this.DatabaseName} SET Multi_User;");
```
Charles-Gagnon, line 392: We should still have tests that : 

1. Verify a case sensitive database can be used without error (at least one for each binding type)
2. Using incorrect casing for a column name throws an error

lucyzhang929: I will open a separate PR to add the tests in all languages since this PR is quite large already

lucyzhang929:
Discussed offline - for Java Input Bindings if the casing does not match, there is no error thrown but the mismatched field is returned with default values (ex. all Product objects will be returned with {"ProductID": 0} since it doesn't match "ProductId").

docs/SetupGuide_Dotnet.md
```diff
@@ -95,7 +95,7 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a
 - Add 'using System.Collections.Generic;' to the namespaces list at the top of the page.
 - Currently, there is an error for the IEnumerable. We'll fix this by creating an Employee class.
 - Create a new file and call it 'Employee.cs'
-- Paste the below in the file. These are the column names of our SQL table.
+- Paste the below in the file. These are the column names of our SQL table. Note that the casing of the Object field names and the table column names must match.
```
MaddyDev, line 98: Can you also add this note in the other languages docs appropriately.

test/Common/ProductColumnTypes.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
 {
     public class ProductColumnTypes
     {
-        public int ProductID { get; set; }
+        public int ProductId { get; set; }
```
MaddyDev, line 10: I have these duplicated in test-outofproc folder as well, probably a good idea to move them to a common project and reference them to avoid these redundant issues, but for now can you please update them there as well, thanks!

MaddyDev (approved):

PR #561, opened 2022-12-05
Increase build + test timeout

Charles-Gagnon:
With the additional test languages being added the test task is pushing the limits of the default 60min timeout.


chlafreniere (approved):
This is fine for now, but we may want to consider altering our pipelines to parallelize language tests

Charles-Gagnon:
https://github.com/Azure/azure-functions-sql-extension/issues/562

PR #568, opened 2022-12-06
add package info and read me to the sql isolated worker

MaddyDev:

PR #571, opened 2022-12-07
fix README for OutOfProc and clean up Directory.build.props

### LINKED ISSUE #570: Setup for Out of Proc has invalid method definition
https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/SetupGuide_DotnetOutOfProc.md#setup-for-input-bindings

`public static async IEnumerable<Employee> Run(`

this isn't valid - async methods must return a Task.

MaddyDev:
Fixes #570 and addresses the comments from PR #568


docs/SetupGuide_DotnetOutOfProc.md
```diff
@@ -66,7 +66,7 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a
 
     ```csharp
     [Function("GetEmployees")]
-    public static async IEnumerable<Employee> Run(
+    public static async Task<IEnumerable<Employee>> Run(
```
Charles-Gagnon, line 69: I'd just remove the async. This doesn't need to be async the way it's written currently and we want these to be as simple as possible.

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -12,8 +12,6 @@
     <PackageId>Microsoft.Azure.WebJobs.Extensions.Sql</PackageId>
     <PackageTags>Microsoft Azure WebJobs AzureFunctions SQL AzureSQL</PackageTags>
```
Charles-Gagnon, line 13: You should probably add package tags to the worker extension too. See if there's any other tags that other work extensions use (in addition to these ones)

They can stay in the separate csproj's though since those are like IDs in that they are attributed more directly to the specific package being generated.

Directory.Build.props
```diff
@@ -26,5 +25,7 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <PackageProjectUrl>https://github.com/Azure/azure-functions-sql-extension</PackageProjectUrl>
+    <PackageIcon>pkgicon.png</PackageIcon>
```
Charles-Gagnon, line 28: Move the pkgicon.png up to the top level images folder and then update the csproj references to each point to it - we don't need multiple copies of the image in the repo.

Charles-Gagnon (approved):
Pre-emptively signing off, would still be good to get the image moved up before merging

PR #573, opened 2022-12-07
Remove extra events

Charles-Gagnon:
Seeing a huge increase in the number of events so doing a bit more cleanup : 

1. Removing the start events - we only really care about the durations so just the "end" event is enough
2. Only sending GetChanges and RenewLeases events when we actually get/update rows. Since these run constantly in a loop most of the iterations don't actually have any changes processed so we don't really need to send the events for those


src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -108,7 +108,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
 
             this.InitializeTelemetryProps();
             TelemetryInstance.TrackEvent(
-                TelemetryEventName.StartListenerStart,
+                TelemetryEventName.StartListener,
```
lucyzhang929, line 111: Can this be removed since we are sending StartListener in line 177?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -174,7 +174,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
                         [TelemetryMeasureName.TransactionDurationMs] = transactionSw.ElapsedMilliseconds,
                     };
 
-                    TelemetryInstance.TrackEvent(TelemetryEventName.StartListenerEnd, this._telemetryProps, measures);
+                    TelemetryInstance.TrackEvent(TelemetryEventName.StartListener, this._telemetryProps, measures);
```
lucyzhang929, line 177: Can this be removed as well?

PR #577, opened 2022-12-08
Enable queue trigger and timer trigger tests for OOP

### LINKED ISSUE #550: Enable TimerTrigger test for OOP

MaddyDev:
Fixes #550 and #549 

Added QueueTrigger sample and enables tests for both QueueTrigger and TimerTrigger.


test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -162,7 +161,6 @@ public void QueueTriggerProductsTest(SupportedLanguages lang)
 
         [Theory]
         [SqlInlineData()]
-        [UnsupportedLanguages(SupportedLanguages.OutOfProc)]
         public void TimerTriggerProductsTest(SupportedLanguages lang)
```
chlafreniere, line 164: Did we already have TimerTriggerProducts for OOP?

samples/samples-outofproc/OutputBindingSamples/QueueTriggerProducts.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Extension.Sql;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.OutputBindingSamples
+{
+    public static class QueueTriggerProducts
+    {
+        [Function("QueueTriggerProducts")]
+        [SqlOutput("[dbo].[Products]", ConnectionStringSetting = "SqlConnectionString")]
+        public static List<Product> Run([QueueTrigger("testqueue")] string queueMessage)
```
chlafreniere, line 15: out of curiosity what is queueMessage used for? I see that it's an unused parameter, just curious in general for queue triggers

test/Integration/SqlOutputBindingIntegrationTests.cs
MaddyDev: yes we did, the sample and test were working on my local but were failing on the pipeline. I realized that's probably because of the negative scenarios overwriting the samples.

samples/samples-outofproc/OutputBindingSamples/QueueTriggerProducts.cs
MaddyDev: We use the [queue trigger](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue-trigger?tabs=in-process%2Cextensionv5&pivots=programming-language-csharp) to start a function when a new item is received on a queue. The queue message is provided as input to the function. More like event based triggers on sql, here we are checking for a message on queue and then executing the function.

chlafreniere (approved):

PR #579, opened 2022-12-09
convert date to iso 8061 format

### LINKED ISSUE #548: AddProductColumnTypesTest failing for OutOfproc

MaddyDev:
Fixes #548 
The OOP function sends the date in the below format:
2022-12-01T02:39:56.098554Z and Sql fails with the following error: System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Core Microsoft SqlClient Data Provider: Conversion failed when converting date and/or time from character string.

In order for SQL to accept that format, datatype of that field should be DateTime2. Explicitly converting the datetime to ISO 8061 with milliseconds: yyyy-MM-ddTHH:mm:ss.fffZ and sending it so it's acceptable on SQL server.

### DIFF
--- a/test-outofproc/AddProductColumnTypes.cs
+++ b/test-outofproc/AddProductColumnTypes.cs
@@ -26,7 +26,7 @@ namespace DotnetIsolatedTests
             var product = new ProductColumnTypes()
             {
                 ProductID = int.Parse(queryStrings["productId"], null),
-                Datetime = DateTime.Parse(DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ", System.Globalization.CultureInfo.InvariantCulture)),
+                Datetime = DateTime.Parse(DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ", System.Globalization.CultureInfo.InvariantCulture), System.Globalization.CultureInfo.InvariantCulture),
                 Datetime2 = DateTime.UtcNow
             };
 
--- a/test-outofproc/AddProductColumnTypes.cs
+++ b/test-outofproc/AddProductColumnTypes.cs
@@ -8,6 +8,7 @@ using System.Collections.Specialized;
 using Microsoft.Azure.Functions.Worker.Extension.Sql;
 using DotnetIsolatedTests.Common;
 using System;
+using System.Data.SqlTypes;
 
 namespace DotnetIsolatedTests
 {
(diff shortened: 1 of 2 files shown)


test-outofproc/AddProductColumnTypes.cs
```diff
@@ -26,7 +26,7 @@ public static class AddProductColumnTypes
             var product = new ProductColumnTypes()
             {
                 ProductID = int.Parse(queryStrings["productId"], null),
-                Datetime = DateTime.UtcNow.Date,
+                Datetime = DateTime.Parse(DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ", System.Globalization.CultureInfo.InvariantCulture)),
                 Datetime2 = DateTime.UtcNow
```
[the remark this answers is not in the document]
MaddyDev: The error is caused due to the fact that Datetime fields data type is DateTime. DateTime2 fileds data type is DateTime2 and sql accepts the format being sent for that datatype. DateTime data type only accepts upto 3 digits for the milliseconds hence fails with conversion error, so the explicit conversion for that field is required.

VasuBhog (approved):

test-outofproc/AddProductColumnTypes.cs
```diff
@@ -26,7 +26,7 @@ public static class AddProductColumnTypes
             var product = new ProductColumnTypes()
             {
                 ProductID = int.Parse(queryStrings["productId"], null),
-                Datetime = DateTime.UtcNow.Date,
+                Datetime = DateTime.ParseExact(DateTime.UtcNow.ToString(System.Globalization.CultureInfo.InvariantCulture), "yyyy-MM-ddTHH:mm:ss.fffZ", null),
```
chlafreniere, line 29: Out of curiosity have you tried using https://learn.microsoft.com/en-us/dotnet/api/system.data.sqltypes.sqldatetime?view=net-7.0?

Does something like this work?

```
DateTime = new SqlDateTime(DateTime.UtcNow).ToSqlString(),
```

MaddyDev: Didn't know about SqlDateTime, going to give it a try and see if it works. Thanks for the suggestion!

lucyzhang929 (approved):

PR #580, opened 2022-12-09
Alphabetical order columns from tables and JSON data

VasuBhog:
PowerShell Identity Tests Integration have been failing with 
`Core Microsoft SqlClient Data Provider: Conversion failed when converting the nvarchar value 'Glasses' to data type int.` (as an example). This is due to the generated query for the insertion shown below:
` Query=WITH cte AS ( SELECT * FROM OPENJSON(@rowData) WITH (**[name] nvarchar(100),[cost] int)** ) INSERT INTO [dbo].[ProductsWithIdentity] (**[cost],[name]**) SELECT * FROM cte;`

As you can see the first part of the query where we get the column names from table information is in the order that the table is in. However the JSON we get back from the PowerShell worker gives it back in alphabetical order as such will always get the Conversion failed issue. 

I have discussed this with the PowerShell team and we found that the Request query that gets converted from PowerShell (ConvertTo-JSON cmdlet) changes the hashtable formatting and as such it will always give the req_query in the improper format to create the query. 

Mitigation options:
- Change our logic to be alphabetical
- See if there are potential changes that we can do in the PowerShell worker to intercept the order (potential breaking change on their end).
- Discuss with the PowerShell JSON conversion team to understand the expected behavior and if this would ever be changed to aid us.

I chose to change our logic as we can then be unblocked as PS Worker changes most likely will have to come from the PowerShell team first and they do not plan to upgrade to 7.4 until .NET 8.


VasuBhog:
Found another viable solution.

Issue #448, opened 2022-11-03
PowerShell Integration Tests 


VasuBhog:
Known Integration Test Issues from PowerShell:

- [x] AddProductsWithIdentityColumnArray Test, AddProductWithIdentity Test, &
AddProductWithMultiplePrimaryColumnsAndIdentity 
`System.Private.CoreLib: Exception while executing function: Functions.AddProductWithMultiplePrimaryColumnsAndIdentity. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Core Microsoft SqlClient Data Provider: Conversion failed when converting the nvarchar value 'MyProduct' to data type int.`
- This seems (similar issue to the above) to be an issue with how PowerShell gets deserialized as it comes back alphabetically instead of how the input data was upserted.

- [ ] [AddProductWithDefaultPK Test](https://github.com/Azure/azure-functions-sql-extension/blob/main/test/Integration/SqlOutputBindingIntegrationTests.cs#L403) : Not clear issue since the function resolves - but testing locally I get this `System.Private.CoreLib: Exception while executing function: Functions.HttpTrigger1. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Microsoft.Azure.WebJobs.Extensions.Sql: All primary keys for SQL table dbo.ProductsWithDefaultPK need to be found in 'Newtonsoft.Json.Linq.JObject.' Missing primary keys: [productguid].`

- [x] [GetProductsColumnTypesSerialization Test](https://github.com/Azure/azure-functions-sql-extension/blob/main/test/Integration/SqlInputBindingIntegrationTests.cs#L164) `Assert.Equal() Failure
Expected: [{"ProductId":999,"Datetime":"2022-10-20T12:39:13.123Z","Datetime2":"2022-10-20T12:39:13.123Z"}]
Actual:   [{"ProductId":999,"Datetime2":"2022-10-20T12:39:13.123Z","Datetime":"2022-10-20T12:39:13.123Z"}]
`

- [x] AddProductColumnTypesTest - failed on conversion of DateTime from character string.

PR #581, opened 2022-12-10
Use `[ordered]` in PowerShell identity tests

VasuBhog:
Addresses #448.
PowerShell Identity Tests Integration have been failing with
`Core Microsoft SqlClient Data Provider: Conversion failed when converting the nvarchar value 'Glasses' to data type int.`
(as an example). 

This is due to the generated query for the insertion shown below:
`Query=WITH cte AS ( SELECT * FROM OPENJSON(@rowData) WITH (**[name] nvarchar(100),[cost] int)** ) INSERT INTO [dbo].[ProductsWithIdentity] (**[cost],[name]**) SELECT * FROM cte;`

As such I have been investigating multiple solutions with the least amount of disruptive changes we can make, we have found that **hashtables** in PowerShell randomly decided to either stay in the order they are stated in the run.ps1 or becomes alphabetically ordered. We tested this with the latest version of PowerShell 7.3 and 7.2 (which the PS functions worker runtime uses). 

This will fix the inconsistency we find when sending the Request Query for Identity functions that were failing.

Also edited launch.json to include Mac debugging to work and nit readme test changes.


lucyzhang929 (approved):

samples/samples-powershell/AddProductWithIdentityColumn/run.ps1
```diff
@@ -7,7 +7,7 @@ param($Request, $TriggerMetadata)
 Write-Host "PowerShell function with SQL Output Binding processed a request."
 
 # Update req_query with the query of the request
```
Charles-Gagnon, line 9: These comments aren't useful, let's just remove them. Instead it would be great if you could replace it with an explanation of why you're using ordered.

Charles-Gagnon (commented):
Please also add a note to the Python docs about when and why users should use ordered in their own functions.

samples/samples-powershell/AddProductWithIdentityColumn/run.ps1
VasuBhog: > Please also add a note to the Python docs about when and why users should use ordered in their own functions.

Did you mean PowerShell docs?

Charles-Gagnon: Yes 😁

VasuBhog: Sounds good, are there the docs you are referring to [here](https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/SetupGuide_PowerShell.md#output-binding) to confirm? 

I can remove the other comments in another PR and the linting issue.

Charles-Gagnon: We've been putting stuff like that in https://github.com/Azure/azure-functions-sql-extension#known-issues

Although it might be better to have a language-specific known issues section for each language guide (that would then be linked to from the main README). But for now just keep it with the others and we can move them later if we decide that's useful.

PR #594, opened 2022-12-20
Update to .NET 7 SDK

Charles-Gagnon:
Updating to latest SDK to get build improvements (including the check that resulted in the code update in this PR). 

This does not change the target framework for our extension - that will still be net6.

### DIFF
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -174,12 +174,18 @@ steps:
     GdnPublishTsaConfigFile: '$(Build.SourcesDirectory)\builds\TSAConfig.gdntsa'
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'), eq(variables['TSA_UPLOAD'], 'true'))
 
-# 5.0 isn't supported on Mac yet
+# net6+ isn't supported on Mac yet
 - task: UseDotNet@2
-  displayName: 'Install .NET Core 2.1.x SDK for CodeSigning on Mac'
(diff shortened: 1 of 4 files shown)


chlafreniere:
Is there an existing issue that this fixes?

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -174,13 +174,19 @@ steps:
     GdnPublishTsaConfigFile: '$(Build.SourcesDirectory)\builds\TSAConfig.gdntsa'
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'), eq(variables['TSA_UPLOAD'], 'true'))
 
-# 5.0 isn't supported on Mac yet
+# net6+ isn't supported on Mac yet
```
chlafreniere, line 177: Is .net core 2.1 really still needed for mac?

Charles-Gagnon: I'll check once I've gotten the other issues fixed 😄

Charles-Gagnon:
> Is there an existing issue that this fixes?

No, just was reminded of this from the open PR in STS. We should generally be keeping up to date with the latest SDK on a regular basis - there isn't really any reason we wouldn't want to and it makes sure we get the latest improvements they've made to the build tools.

Charles-Gagnon:
Ok, well turns out that we'd need to install net6 in order to run the tests - which while not difficult is just extra stuff I'd rather not add at this point. So closing this and just doing a separate PR to fix the one code suggestion (since that was something I hit when running test functions that directly referenced the project, and is a good suggestion anyways).

PR #598, opened 2022-12-21
Introduce Target Based Scaling

AmeyaRele:
This PR refactors the scaling logic from `SqlTriggerListener` class to `SqlTriggerScaleMonitor` and `SqlTriggerMetricsProvider` class.
It also introduces `SqlTriggerTargetScaler` for target based scaling logic.

Reference PRs: 
https://github.com/Azure/azure-sdk-for-net/pull/31367 
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/806

**ToDo**

- [x] Add unit tests for SqlTriggerTargetScaler
- [x] Test with scale controller 3.x prototype

Local functions host with target based scaling enabled
target-based-scaling-enabled-sql


nuget.config
```diff
@@ -3,5 +3,15 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="azure-appservice" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
```
Charles-Gagnon, line 6: Is this temporary? There's a package on nuget but it isn't up to 36 yet...

AmeyaRele: Yes, using the private nuget package until functions host is released with 36.

AmeyaRele: No, the webjobs sdk with target based scaling support has not been released yet. Will update the package source as soon as it is released.

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+using static Microsoft.Azure.WebJobs.Extensions.Sql.SqlTriggerConstants;
+using static Microsoft.Azure.WebJobs.Extensions.Sql.SqlTriggerUtils;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    internal class SqlTriggerMetricsProvider
```
Charles-Gagnon, line 19: Add doc comment explaining what this class is used for (ideally with links to public docs about scaling)

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.Sql.Telemetry;
+using static Microsoft.Azure.WebJobs.Extensions.Sql.Telemetry.Telemetry;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
+using MoreLinq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    public class SqlTriggerScaleMonitor : IScaleMonitor<SqlTriggerMetrics>
```
Charles-Gagnon, line 16: Add doc comment explaining what this class is used for (ideally with links to public docs about scaling)

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    public class SqlTriggerTargetScaler : ITargetScaler
```
Charles-Gagnon, line 10: Add doc comment explaining what this class is used for (ideally with links to public docs about scaling)

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,55 @@
...
+        public async Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext context)
+        {
+            SqlTriggerMetrics metrics = await this._metricsProvider.GetMetricsAsync();
+
+            return this.GetScaleResultInternal(context, metrics.UnprocessedChangeCount);
+        }
+
+        internal TargetScalerResult GetScaleResultInternal(TargetScalerContext context, long unprocessedChangeCount)
+        {
+            int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
```
Charles-Gagnon, line 38: What does "instance concurrency" mean and how is it related to the max changes per worker setting? Those two seem like slightly different concepts to me.

Charles-Gagnon (commented):
Couple small comments but overall looks good once the package is updated.

src/TriggerBinding/SqlTriggerTargetScaler.cs
AmeyaRele: InstanceConcurrency is a value provided by functions host when [dynamic concurrency](https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency) is enabled, more information is provided in the link. The max changes per worker setting would correspond to a static concurrency setting.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
AmeyaRele: Done. Added general scaling doc for functions. Couldn't find anything on public docs specifically related to incremental scaling.

src/TriggerBinding/SqlTriggerTargetScaler.cs
AmeyaRele: Added doc comment. As target based scaling has not yet rolled out, couldn't get any public docs for it.

src/TriggerBinding/SqlTriggerTargetScaler.cs
AmeyaRele: I did add this in the latest commit, with the other changes, do you want me to provide the link to dynamic concurrency and make the comment more detailed?

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,59 @@
...
+        public async Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext context)
+        {
+            SqlTriggerMetrics metrics = await this._metricsProvider.GetMetricsAsync();
+
+            return this.GetScaleResultInternal(context, metrics.UnprocessedChangeCount);
+        }
+
+        internal TargetScalerResult GetScaleResultInternal(TargetScalerContext context, long unprocessedChangeCount)
+        {
+            // Instance concurrency value is set by the functions host when dynamic concurrency is enabled.
```
Charles-Gagnon, line 41:
```suggestion
            // Instance concurrency value is set by the functions host when dynamic concurrency is enabled. See https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency for more details.
```

Directory.Packages.props
```diff
@@ -4,7 +4,7 @@
     <PackageVersion Include="Microsoft.NET.Sdk.Functions" Version="4.1.3" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions" Version="1.1.0" />
     <PackageVersion Include="Microsoft.ApplicationInsights" Version="2.17.0" />
-    <PackageVersion Include="Microsoft.Azure.WebJobs" Version="3.0.32" />
+    <PackageVersion Include="Microsoft.Azure.WebJobs" Version="3.0.36-11991" />
```
pragnagopa, line 7: 3.0.36 is now available on NUget - https://www.nuget.org/packages/Microsoft.Azure.WebJobs - please update

nuget.config
AmeyaRele: Removed this

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,185 @@
...
+using Microsoft.Extensions.Logging;
+using MoreLinq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    /// <summary>
+    /// Makes the scale decision for incremental scaling(+1, 0, -1) for workers required based on unprocessed changes. 
+    /// Guidance for scaling information can be found here https://learn.microsoft.com/en-us/azure/azure-functions/event-driven-scaling
+    /// </summary>
+    public class SqlTriggerScaleMonitor : IScaleMonitor<SqlTriggerMetrics>
```
chiangvincent, line 20: Does this need to be public? Same with SqlTriggerTargetScaler and the SqlTriggerMetrics class. For the purposes of scale controller v3.x and TBS, we no longer need the extension implementations of TargetScaler and ScaleMonitor to be public (API was redesigned)

src/TriggerBinding/SqlTriggerMetrics.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.WebJobs.Extensions.Sql
 {
-    internal sealed class SqlTriggerMetrics : ScaleMetrics
+    public class SqlTriggerMetrics : ScaleMetrics
```
chiangvincent, line 8: Because SqlTriggerScaleMonitor does not need to be public, change back to internal sealed

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,185 @@
...
+            this._metricsProvider = new SqlTriggerMetricsProvider(connectionString, logger, this._userTable, userFunctionId);
+            this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            this._maxChangesPerWorker = maxChangesPerWorker;
+        }
+
+        public ScaleMonitorDescriptor Descriptor { get; }
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await this.GetMetricsAsync();
+
```
chiangvincent, line 45: Nit: remove extra whitespace

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Sql
+{
+    /// <summary>
+    /// Represents the target based scaler returning a target worker count.
+    /// </summary>
+    public class SqlTriggerTargetScaler : ITargetScaler
```
chiangvincent, line 13: Same as above, change back to internal since public not explicitly needed

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,59 @@
...
+            int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
+
+            if (concurrency < 1)
+            {
+                throw new ArgumentOutOfRangeException($"Unexpected concurrency='{concurrency}' - the value must be > 0.");
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(unprocessedChangeCount / (decimal)concurrency);
+
+            this._logger.LogInformation($"Target worker count for function '{this._userFunctionId}' is '{targetWorkerCount}' UnprocessedChangeCount ='{unprocessedChangeCount}', Concurrency='{concurrency}').");
```
chiangvincent, line 51: It would be helpful to log the user table name here as well (will need to save var in the constructor)

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -175,6 +173,9 @@ public async Task StartAsync(CancellationToken cancellationToken)
                     };
 
                     TelemetryInstance.TrackEvent(TelemetryEventName.StartListenerEnd, this._telemetryProps, measures);
+
+                    this._scaleMonitor = new SqlTriggerScaleMonitor(this._userFunctionId, this._userTable.FullName, this._connectionString, this._maxChangesPerWorker, this._logger);
```
chiangvincent, line 177: I see that the default MaxChangesPerWorker, which is used as the target value, is 1000. This is quite high (defaults for Service Bus and Event Hubs are 16 and 10 respectively). 

What range of values would you expect UnprocessedChangeCount for an average customer to be? If this is well under 1000 I would consider lowering the default value.

chiangvincent (changes requested):
Left minor comments and clarifications. Please also verify these changes with a locally running functions host, and paste screenshots of Target Based Scaling and Incremental Scaling logs from console. See [here](https://github.com/Azure/azure-sdk-for-net/pull/33714) for reference.

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            long unprocessedChangeCount = 0L;
+
+            try
+            {
+                long getUnprocessedChangesDurationMs = 0L;
+
+                using (var connection = new SqlConnection(this._connectionString))
+                {
+                    this._logger.LogDebugWithThreadId("BEGIN OpenGetUnprocessedChangesConnection");
```
alrod, line 54: Where does the log go, kusto FunctionLogs table?  Do we want to log it on every call?

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+                            this._logger.LogDebugWithThreadId($"END GetUnprocessedChangeCount Duration={getUnprocessedChangesDurationMs}ms Count={unprocessedChangeCount}");
+                            transaction.Commit();
+                        }
+                        catch (Exception)
+                        {
+                            try
+                            {
+                                transaction.Rollback();
+                            }
+                            catch (Exception ex2)
```
alrod, line 84: We are trying twice, right?

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: This is something we'll be monitoring with our telemetry and will adjust as needed, for now though 1000 as the default doesn't seem unreasonable as 1000 rows is generally a very manageable amount of data for a single worker to handle.

Do you have information on how Service Bus and Event Hubs arrived at those values for their use cases? Is there some expectation that each change will add enough extra overhead that extra workers need to be spun up?

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,185 @@
...
+                [TelemetryPropertyName.TriggerMetrics] = metrics is null ? "null" : $"[{string.Join(", ", metrics.Select(metric => metric.UnprocessedChangeCount))}]",
+                [TelemetryPropertyName.WorkerCount] = $"{workerCount}",
+            };
+
+            try
+            {
+                status = this.GetScaleStatusCore(workerCount, metrics);
+
+                properties[TelemetryPropertyName.ScaleRecommendation] = $"{status.Vote}";
+                TelemetryInstance.TrackEvent(TelemetryEventName.GetScaleStatus, properties);
```
alrod, line 82: Does this log go to some SQL internal logs?

src/TriggerBinding/SqlTriggerMetricsProvider.cs
Charles-Gagnon: This is logging at the debug level so I would only expect it to be logged if the log level is set to that by the customer. And information like this is useful for helping track down issues (for example if the connection hangs on opening this would help indicate that).

src/TriggerBinding/SqlTriggerScaleMonitor.cs
Charles-Gagnon: Sort of, it goes to our internal telemetry pipeline (I wouldn't call them "logs" though since it's more collection of statistics, for example in this case the number of times that a given type of scale recommendation was given)

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            SqlTriggerMetrics metrics = await this._metricsProvider.GetMetricsAsync();
+
+            return this.GetScaleResultInternal(context, metrics.UnprocessedChangeCount);
+        }
+
+        internal TargetScalerResult GetScaleResultInternal(TargetScalerContext context, long unprocessedChangeCount)
+        {
+            // Instance concurrency value is set by the functions host when dynamic concurrency is enabled. See https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency for more details.
+            int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
```
Charles-Gagnon, line 43: How does InstanceConcurrency map to max changes per worker? The docs don't seem to mention InstanceConcurrency directly and that seems like it would more control the number of workers, not the number of changes per worker.

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            if (concurrency < 1)
+            {
+                throw new ArgumentOutOfRangeException($"Unexpected concurrency='{concurrency}' - the value must be > 0.");
```
Charles-Gagnon, line 47: The first parameter is the name of the arg - although I would also suggest renaming it to something clearer like `maxChangesPerWorker`. 

```suggestion
                throw new ArgumentOutOfRangeException(nameof(concurrency), $"Unexpected concurrency='{concurrency}' - the value must be > 0.");
```

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(unprocessedChangeCount / (decimal)concurrency);
+
+            this._logger.LogInformation($"Target worker count for function '{this._userFunctionId}' is '{targetWorkerCount}' TableName ='{this._userTableName}' UnprocessedChangeCount ='{unprocessedChangeCount}', Concurrency='{concurrency}').");
```
Charles-Gagnon, line 52:
```suggestion
            this._logger.LogDebugWithThreadId($"Target worker count for function '{this._userFunctionId}' is '{targetWorkerCount}' TableName ='{this._userTableName}' UnprocessedChangeCount ='{unprocessedChangeCount}', Concurrency='{concurrency}').");
```

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+                                this._logger.LogDebugWithThreadId($"BEGIN GetUnprocessedChangeCount Query={getUnprocessedChangesCommand.CommandText}");
+                                var commandSw = Stopwatch.StartNew();
+                                unprocessedChangeCount = (long)await getUnprocessedChangesCommand.ExecuteScalarAsync();
+                                getUnprocessedChangesDurationMs = commandSw.ElapsedMilliseconds;
+                            }
+
+                            this._logger.LogDebugWithThreadId($"END GetUnprocessedChangeCount Duration={getUnprocessedChangesDurationMs}ms Count={unprocessedChangeCount}");
+                            transaction.Commit();
+                        }
+                        catch (Exception)
```
Charles-Gagnon, line 78: TrackException here and the rollback

src/TriggerBinding/SqlTriggerTargetScaler.cs
AmeyaRele: You're correct, docs don't seem to mention it directly. I'll follow up on this.
Cosmos DB extension's TBS does map it to [MaxItemsPerInvocation](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/259962519471b7f64b13c5db4d254f4571167175/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosDBTargetScaler.cs#L47)

src/TriggerBinding/SqlTriggerTargetScaler.cs
AmeyaRele: Changed it a bit since I was getting this compiler error.
`Method GetScaleResultInternal passes 'concurrency' as the paramName argument to a ArgumentOutOfRangeException constructor. Replace this argument with one of the method's parameter names. Note that the provided parameter name should have the exact casing as declared on the method. `

Charles-Gagnon:
Latest changes look good, although you have some merge conflicts to fix up here first.

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+                                unprocessedChangeCount = (long)await getUnprocessedChangesCommand.ExecuteScalarAsync();
+                                getUnprocessedChangesDurationMs = commandSw.ElapsedMilliseconds;
+                            }
+
+                            this._logger.LogDebugWithThreadId($"END GetUnprocessedChangeCount Duration={getUnprocessedChangesDurationMs}ms Count={unprocessedChangeCount}");
+                            transaction.Commit();
+                        }
+                        catch (Exception ex)
+                        {
+                            TelemetryInstance.TrackException(TelemetryErrorName.GetUnprocessedChangeCount, ex, null, new Dictionary<TelemetryMeasureName, double>() { { TelemetryMeasureName.GetUnprocessedChangesDurationMs, getUnprocessedChangesDurationMs } });
```
Charles-Gagnon, line 81: We can probably remove this one since you're sending it below too. Make sure you add the duration to that one though.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,184 @@
...
+                if (expectedUnprocessedChangeCount > workerCount * this._maxChangesPerWorker)
+                {
+                    status.Vote = ScaleVote.ScaleOut;
+                    this._logger.LogInformation($"Requesting scale-out: Found the unprocessed changes for table: '{this._userTable.FullName}' to be continuously increasing" +
+                        " and may exceed the maximum limit set for the workers.");
+                    return status;
+                }
+                else
+                {
+                    this._logger.LogDebugWithThreadId($"Avoiding scale-out: Found the unprocessed changes: {expectedUnprocessedChangeCount} for table: '{this._userTable.FullName}' to be increasing" +
```
Charles-Gagnon, line 159: Since all the rest are information it seems like this should match. 

But I'm also thinking we may want to make these all debug to avoid a bunch of noisy messages in the logs. What do other extensions do?

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,184 @@
...
+                // Scale in only if the count of unprocessed changes will not exceed the combined limit post the scale-in operation.
+                if (lastUnprocessedChangeCount <= (workerCount - 1) * this._maxChangesPerWorker)
+                {
+                    status.Vote = ScaleVote.ScaleIn;
+                    this._logger.LogInformation($"Requesting scale-in: Found table: '{this._userTable.FullName}' to be either idle or the unprocessed changes to be continuously decreasing.");
+                    return status;
+                }
+                else
+                {
+                    this._logger.LogDebug($"Avoiding scale-in: Found the unprocessed changes for table: '{this._userTable.FullName}' to be decreasing" +
```
Charles-Gagnon, line 175:
```suggestion
                    this._logger.LogDebugWithThreadId($"Avoiding scale-in: Found the unprocessed changes for table: '{this._userTable.FullName}' to be decreasing" +
```

src/TriggerBinding/SqlTriggerScaleMonitor.cs
```diff
@@ -0,0 +1,184 @@
...
+                        " but they are high enough to require all existing workers for processing.");
+                }
+            }
+
+            this._logger.LogInformation($"Requesting no-scaling: Found the number of unprocessed changes for table: '{this._userTable.FullName}' to not require scaling.");
```
Charles-Gagnon, line 180: I'm wondering if we should just make these all debug. Is this something users will generally find useful? What do other extensions do for logging related to scaling?

Charles-Gagnon (approved):
Please also update https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsOverview.md#scaling-for-trigger-bindings - and work with @dzsquared to make sure the Learn docs are accurate as well.

Charles-Gagnon (changes requested):
Just noticed the test failures, please address those.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
AmeyaRele: I see that Cosmos DB, service bus extensions have logs for scaling at the information level. 
I do think that these logs are helpful for finding out if the scale controller is at fault if the scale decision does not correspond to number of workers allocated for the function.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
AmeyaRele: It might have been Debug level(I just moved this code from Listener class to this while refactoring) since it is not changing the instance counts.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
Charles-Gagnon: Works for me, easy enough to change later on if we think it's too noisy.

src/TriggerBinding/SqlTriggerScaleMonitor.cs
Charles-Gagnon: Yeah, that makes sense I suppose. Alright, I'll let you do what you think makes sense and then we can keep an eye here once we start actually getting people using it to see if it'd be helpful to change these.

docs/BindingsOverview.md
```diff
@@ -167,9 +167,16 @@ This controls the upper limit on the number of pending changes in the user table
...
+
+There are two types of scaling available:
+
+- Incremental scaling - This scales the application serially, increasing or decreasing the workers by 1. There are a couple of checks made to decide on whether the host application needs to be scaled in or out. The rationale behind these checks is to ensure that the count of pending changes per application-worker stays below a certain maximum limit, which is defaulted to 1000, while also ensuring that the number of workers running stays minimal. The scaling decision is made based on the latest count of the pending changes and whether the last 5 times we checked the count, we found it to be continuously increasing or decreasing.
+
+- Target Based Scaling - This type of scaling depends on the pending change count and the value of [dynamic concurrency](https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency#dynamic-concurrency) which if not enabled is defaulted to [Sql_Trigger_MaxChangesPerWorker](#sql_trigger_maxchangesperworker). The target worker count is decided by dividing the pending changes by the concurrency value. The application scales out to the number of instances specified by the target worker count.
```
Charles-Gagnon, line 176: We should also update the Sql_Trigger_MaxChangesPerWorker section to include the current default value if the user doesn't specify it themselves.

```suggestion
- Target Based Scaling - This type of scaling depends on the pending change count and the value of [dynamic concurrency](https://learn.microsoft.com/azure/azure-functions/functions-concurrency#dynamic-concurrency) which if not enabled is defaulted to [Sql_Trigger_MaxChangesPerWorker](#sql_trigger_maxchangesperworker). The target worker count is decided by dividing the pending changes by the concurrency value. The application scales out to the number of instances specified by the target worker count.
```

docs/BindingsOverview.md
```diff
@@ -167,9 +167,16 @@ This controls the upper limit on the number of pending changes in the user table
...
+If your application containing functions with SQL trigger bindings is running as an Azure function app, it will be scaled automatically based on the amount of changes that are pending to be processed in the user table. As of today, we only support scaling of function apps running in Elastic Premium plan. To enable scaling, you will need to go the function app resource's page on Azure Portal, then to Configuration > 'Function runtime settings' and turn on 'Runtime Scale Monitoring'.
+
+There are two types of scaling available:
+
+- Incremental scaling - This scales the application serially, increasing or decreasing the workers by 1. There are a couple of checks made to decide on whether the host application needs to be scaled in or out. The rationale behind these checks is to ensure that the count of pending changes per application-worker stays below a certain maximum limit, which is defaulted to 1000, while also ensuring that the number of workers running stays minimal. The scaling decision is made based on the latest count of the pending changes and whether the last 5 times we checked the count, we found it to be continuously increasing or decreasing.
```
Charles-Gagnon, line 174:
```suggestion
- Incremental scaling - This scales the application serially, increasing or decreasing the workers by 1. There are a couple of checks made to decide on whether the host application needs to be scaled in or out. The rationale behind these checks is to ensure that the count of pending changes per application-worker stays below a certain maximum limit, controlled by [Sql_Trigger_MaxChangesPerWorker](#sql_trigger_maxchangesperworker), while also ensuring that the number of workers running stays minimal. The scaling decision is made based on the latest count of the pending changes and whether the last 5 times we checked the count, we found it to be continuously increasing or decreasing.
```

Charles-Gagnon:
@AmeyaRele - Please take a look at the open comments

Charles-Gagnon (approved):

Charles-Gagnon:
@AmeyaRele Please provide screenshots/logs as requested by @chiangvincent https://github.com/Azure/azure-functions-sql-extension/pull/598#pullrequestreview-1300401704

AmeyaRele:
Having some issues with running it on local functions host, syncing up with Vincent offline to fix them.

Charles-Gagnon:
@AmeyaRele FYI - in preparation for input/output binding GA we've switched trigger stuff back to a separate branch (`release/trigger`). I've switched this to target that new branch instead so nothing should change, just letting you know in case you need to make any more changes.

AmeyaRele:
@chiangvincent Please do take a look at the PR and the attached screenshot. I have addressed all your comments from the last review

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,60 @@
...
+        internal TargetScalerResult GetScaleResultInternal(int concurrency, long unprocessedChangeCount)
+        {
+            if (concurrency < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(concurrency), $"Unexpected concurrency='{concurrency}' - the value must be > 0.");
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(unprocessedChangeCount / (decimal)concurrency);
+
+            this._logger.LogDebugWithThreadId($"Target worker count for function '{this._userFunctionId}' is '{targetWorkerCount}' TableName ='{this._userTableName}' UnprocessedChangeCount ='{unprocessedChangeCount}', Concurrency='{concurrency}').");
```
chiangvincent, line 52: nit: add comma after `'{this._userTableName}'`

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -0,0 +1,60 @@
...
+        internal TargetScalerResult GetScaleResultInternal(int concurrency, long unprocessedChangeCount)
+        {
+            if (concurrency < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(concurrency), $"Unexpected concurrency='{concurrency}' - the value must be > 0.");
+            }
+
+            int targetWorkerCount = (int)Math.Ceiling(unprocessedChangeCount / (decimal)concurrency);
+
+            this._logger.LogDebugWithThreadId($"Target worker count for function '{this._userFunctionId}' is '{targetWorkerCount}' TableName ='{this._userTableName}', UnprocessedChangeCount ='{unprocessedChangeCount}', Concurrency='{concurrency}').");
```
chiangvincent, line 52: I think this should be information level log, as it is the only one emitted for scale decisions

chiangvincent (changes requested):

chiangvincent (approved):
Looks good, thanks for the PR! Approving, and please also get approval from @alrod

src/TriggerBinding/SqlTriggerMetricsProvider.cs
alrod: yes, the AppInsights logs will no be published. I am concerning about  kusto FunctionLogs table as we log the all levels there. I am wondering about the amount of the logs as the checks will be executed periodically. We can log if an operation time was exceeded some expected threshold or an exception was occurred.

There lot of "LogDebugWithThreadId" introduced in the PR. Can you try to deploy a function app + new SQL extension with TBS + RDS enabled and see how much new logs are generated in FunctionLogs kusto table.

alrod (approved):
My only concern is about debug logging.
https://github.com/Azure/azure-functions-sql-extension/pull/598#discussion_r1160086411

TBS related stuff looks good.

Charles-Gagnon:
@AmeyaRele - Chatted with Alexey offline about the logging and we're going to investigate and address that separately. So you're good to merge this in whenever you're ready to and we'll take care of following up about the logging stuff.

Issue #593, opened 2022-12-19
SQLTrigger - Error => Reconnect to SQL Server 


Jetro223:
I am using a SqlTrigger to get changes from a Azure SQL Database table.

```c#
public void Run(
            [SqlTrigger("[raw].[transferMessage]", ConnectionStringSetting = "SqlConnectionStringRaw")]
            IReadOnlyList<SqlChange<TransferMessage>> changes, ILogger log)
{
}
```

When the connection to the Azure SQL DB is lost (scaling of the elastic pool/ manual kill of the connection) the function logs the following every second:

`Failed to check for changes in table 'raw.transferMessage' due to exception: System.InvalidOperationException. Exception message: Invalid operation. The connection is closed.`

It's necessary to restart the function app to resume operation. 

Is there anything like auto-reconnect which I'm missing from the docs or what can I do?

Edit: 
Adding ConnectRetryCount=50;ConnectRetryInterval=10 to the connection string does not work when scaling the database

Charles-Gagnon:
Unfortunately no, currently you will need to restart the app.

This is definitely something we'll want to address though. The most straightforward would be just to check if the connection is closed before each iteration of the loop and attempt to reopen it if isn't. That should cover any scenario where the connection is closed unexpectedly.

PR #599, opened 2022-12-22
Add connection retry logic for trigger binding loops

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/593

Previously any errors that occurred with the connection after the loops had been started were logged and ignored - so if the connection was broken then it would never attempt to reconnect and the function would need to be restarted.

This adds detection for when a fatal SQL Exception occurs (Class >= 20 per https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlexception?view=sqlclient-dotnet-standard-5.0#remarks) and when that happens or when the connection is detected to be in a Closed/Broken state then it will attempt to re-establish the connection before continuing on with the loop.

It will continue trying to re-establish the connection forever until the connection is successfully made, there is currently no configuration planned for letting users control this or for anything else (such as forcing the function to be restarted) to occur. 

Note that originally I had expected to be able to just check the ConnectionState for it being Closed/Broken, but as detailed in https://github.com/dotnet/SqlClient/issues/1874 that isn't always the case and so I am checking for Class 20 or higher as a fallback, since those all indicate critical errors that will usually require reconnection. 

These checks are only necessary for trigger bindings since the other ones establish a new connection every time they're executed, so won't get stuck like the trigger bindings would. A future improvement could be to add automatic retries to those as well, but that's out of the scope of this PR.

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -254,14 +254,35 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     await connection.OpenAsync(token);
                     this._logger.LogDebugWithThreadId("END OpenChangeConsumptionConnection");
 
+                    bool forceReconnect = false;
                     // Check for cancellation request only after a cycle of checking and processing of changes completes.
                     while (!token.IsCancellationRequested)
                     {
+                        bool isConnected = await connection.TryEnsureConnected(forceReconnect, this._logger, "ChangeConsumptionConnection", token);
+                        if (!isConnected)
+                        {
+                            // If we couldn't reconnect then wait our delay and try again
+                            await Task.Delay(TimeSpan.FromMilliseconds(this._pollingIntervalInMs), token);
+                            continue;
+                        }
+                        else
+                        {
+                            forceReconnect = false;
+                        }
                         this._logger.LogDebugWithThreadId($"BEGIN CheckingForChanges State={this._state}");
                         if (this._state == State.CheckingForChanges)
                         {
-                            await this.GetTableChangesAsync(connection, token);
-                            await this.ProcessTableChangesAsync(connection, token);
+                            try
+                            {
+                                await this.GetTableChangesAsync(connection, token);
+                                await this.ProcessTableChangesAsync(connection, token);
+                            }
+                            catch (Exception e) when (e.IsFatalSqlException())
+                            {
+                                this._logger.LogError($"Fatal SQL Client exception processing changes. Will attempt to reestablish connection in {this._pollingIntervalInMs}ms. Exception = {e.Message}");
+                                forceReconnect = true;
+                            }
+
                         }
                         this._logger.LogDebugWithThreadId("END CheckingForChanges");
                         this._logger.LogDebugWithThreadId($"Delaying for {this._pollingIntervalInMs}ms");
@@ -278,6 +299,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     this._logger.LogError($"Exiting change consumption loop due to exception: {e.GetType()}. Exception message: {e.Message}");
                     TelemetryInstance.TrackException(TelemetryErrorName.ConsumeChangesLoop, e, this._telemetryProps);
                 }
+                throw;
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -471,9 +498,29 @@ private async void RunLeaseRenewalLoopAsync()
...
+                            await Task.Delay(TimeSpan.FromSeconds(LeaseRenewalIntervalInSeconds), token);
+                            continue;
+                        }
+                        else
+                        {
+                            forceReconnect = false;
+                        }
+                        try
+                        {
+                            await this.RenewLeasesAsync(connection, token);
```
cheenamalhotra, line 517: Is retrying safe here? 
Do you have checks to ensure it doesn't cause data redundancy if server DOES process query in first attempt even though connection terminates due to network failure?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -387,6 +409,11 @@ private async Task GetTableChangesAsync(SqlConnection connection, CancellationTo
                 this._rows = new List<IReadOnlyDictionary<string, object>>();
                 this._logger.LogError($"Failed to check for changes in table '{this._userTable.FullName}' due to exception: {e.GetType()}. Exception message: {e.Message}");
                 TelemetryInstance.TrackException(TelemetryErrorName.GetChanges, e, this._telemetryProps);
+                if (e.IsFatalSqlException())
```
cheenamalhotra, line 412: Why not throw every exception?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -254,14 +254,35 @@ private async Task RunChangeConsumptionLoopAsync()
...
+                        }
                         this._logger.LogDebugWithThreadId($"BEGIN CheckingForChanges State={this._state}");
                         if (this._state == State.CheckingForChanges)
                         {
-                            await this.GetTableChangesAsync(connection, token);
-                            await this.ProcessTableChangesAsync(connection, token);
+                            try
+                            {
+                                await this.GetTableChangesAsync(connection, token);
+                                await this.ProcessTableChangesAsync(connection, token);
```
cheenamalhotra, line 278: If connection drops before executing 'ProcessTableChangesAsync()` - you'll be in ProcessingChanges state, so this will not repeat?

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: It should be, yes. All queries are done in transactions so any errors will result in a rollback and retry next time around.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: That's the design for now, we don't want to exit the loop completely and instead retry if an error occurs.

A follow up task is to investigate all possible retry logic for the function - which may change that behavior at that time (most likely implementing some sort of "dead letter queue" like [ServiceBus](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues#the-dead-letter-queue) does so that we can extract problematic data and the user can investigate and re-enter it if needed)

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Good catch! That's actually a bug with the current implementation. I'll file an issue and fix it as a follow up PR.

Charles-Gagnon: https://github.com/Azure/azure-functions-sql-extension/issues/600

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -254,14 +254,35 @@ private async Task RunChangeConsumptionLoopAsync()
...
 
+                    bool forceReconnect = false;
                     // Check for cancellation request only after a cycle of checking and processing of changes completes.
                     while (!token.IsCancellationRequested)
                     {
+                        bool isConnected = await connection.TryEnsureConnected(forceReconnect, this._logger, "ChangeConsumptionConnection", token);
+                        if (!isConnected)
+                        {
+                            // If we couldn't reconnect then wait our delay and try again
+                            await Task.Delay(TimeSpan.FromMilliseconds(this._pollingIntervalInMs), token);
```
JatinSanghvi, line 265: If the SQL service is unresponsive due to heavy load, will trying to reconnect constantly be acceptable? Should we limit the number of retries or have an exponential backoff?

JatinSanghvi (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Yeah, user configuration is likely something we'll want to add. I'll make an issue and we can add that as a follow up item.

PR #606, opened 2023-01-03
Add tests for case sensitive database

lucyzhang929:
Adding three test cases:
1. query from case sensitive database
2. upsert to case sensitive database
3. upserting POCO with mismatched casing fails


test/Integration/test-js/AddProductIncorrectCasing/index.js
```diff
@@ -0,0 +1,18 @@
+/**
```
Charles-Gagnon, line 1: copyright (could you follow up with adding linting for the test files too? Looks like it's just doing the samples folder currently)

test/Integration/test-powershell/AddProductIncorrectCasing/run.ps1
```diff
@@ -0,0 +1,18 @@
+using namespace System.Net
```
Charles-Gagnon, line 1: I suppose these should also have a copyright header. @VasuBhog Can you make an issue to follow up adding that for the powershell files? (ideally with some sort of linting to ensure it's not missed in the future...not sure what exists out there for that)

Charles-Gagnon (approved):

test/Integration/test-js/AddProductIncorrectCasing/index.js
lucyzhang929: Will send a follow up PR for linting test-js folder

test/Integration/test-powershell/AddProductIncorrectCasing/run.ps1
lucyzhang929: Created a follow up issue here: https://github.com/Azure/azure-functions-sql-extension/issues/610

MaddyDev (approved):

Issue #600, opened 2022-12-22
Exceptions while processing changes will result in no changes ever being processed


Charles-Gagnon:
Currently in our change consumption loop we call GetTableChangesAsync and ProcessTableChangesAsync only if the current state is CheckingForChanges. 

https://github.com/Azure/azure-functions-sql-extension/blob/639a7e84ce6c7c52ea009f13482cd46298107ea5/src/TriggerBinding/SqlTableChangeMonitor.cs#L278

But in ProcessTableChangesAsync  we set the state to [ProcessingChanges](https://github.com/Azure/azure-functions-sql-extension/blob/639a7e84ce6c7c52ea009f13482cd46298107ea5/src/TriggerBinding/SqlTableChangeMonitor.cs#L426) before starting, and only clear it in some of the code paths. An example of where we don't clear it is if we successfully process the changes but then an exception is thrown while opening the transaction for releasing the leases ([link](https://github.com/Azure/azure-functions-sql-extension/blob/639a7e84ce6c7c52ea009f13482cd46298107ea5/src/TriggerBinding/SqlTableChangeMonitor.cs#L656)) then it won't correctly reset the state and when the outer loop executes again it will see the state is still set as ProcessingChanges and won't execute either of the above methods - and nothing else will correctly reset those.

PR #615, opened 2023-01-04
Handle errors that occur while processing changes

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/600

Two main changes here : 

- Always run ProcessTableChangesAsync every iteration of the loop - we always want to try and process any changes we have (it'll be a no-op if there aren't any changes). We just want to skip checking for changes if we're still in the ProcessingChanges state since that means an unexpected exception occurred and we weren't able to fully process the current changes
- Add logic to avoid triggering the user function more than once for a given set of changes

With this if any error occurs during the process at all we will still always attempt to process existing changes and clean up afterwards to ensure that we're in a good state before continuing. 

If we get in a bad state - for example if the connection to the server is closed - then it will keep retrying the cleanup until either the function is exited or the connection is re-established.


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -85,6 +85,12 @@ internal sealed class SqlTableChangeMonitor<T> : IDisposable
         private int _leaseRenewalCount = 0;
         private State _state = State.CheckingForChanges;
 
+        /// <summary>
+        /// The result from the last time we triggered the user function. This will normally be cleared immediately after executing
+        /// the function but if an error occurs during cleanup we want to keep retrying the cleanup until it succeeds.
+        /// </summary>
+        private FunctionResult _triggeredFunctionResult = null;
```
chlafreniere, line 92: I'm trying to rationalize this with the State enum. Have you considered adding an explicit new item in that enum instead of saving the FunctionResult and checking against null?

Charles-Gagnon: Alright, tried cleaning it up a bit. Now there's 3 states and I removed the need for tracking the success of the triggered function. Instead, we now have a separate list for tracking what rows are marked to have their leases released. This makes it clearer what the purpose of the variable is and makes it easier to ensure that we only release leases for rows that are done being processed.

MaddyDev (approved):

lucyzhang929 (approved):

PR #621, opened 2023-01-10
use AssemblyMetadata property in csproj file

### LINKED ISSUE #404: Build task to set latest supported version for out of proc sql worker
- The ExtensionInformation metadata should point to the dev version (99.99.99) for local/private builds
- For pipeline builds it should use the major version + preview flag if it exists (e.g. `0.*` or `0.*-preview`)

The easiest way would be to use an [AssemblyMetadata](https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2019#assemblymetadata) property in the csproj since that can easily use build-time values, but this apparently only works for .NET5+ projects. 

If we can't use that then we should have the version set to `99.99.99` in the file and then in the build pipeline replace/rewrite that to use the version generated during the build.

MaddyDev:
Fixes #404

### DIFF
--- a/Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
+++ b/Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
@@ -22,12 +22,10 @@
 		<None Include=".\README.md" Pack="true" PackagePath="" />
 	</ItemGroup>
 
-	<Target Name="AddGitMetadaAssemblyAttributes" BeforeTargets="CoreGenerateAssemblyInfo">
-		<ItemGroup>
-			<AssemblyAttribute Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions.ExtensionInformationAttribute ">
-				<_Parameter1>Microsoft.Azure.WebJobs.Extensions.Sql</_Parameter1>
-				<_Parameter2>0.*-*</_Parameter2>
-			</AssemblyAttribute>
-		</ItemGroup>
-	</Target>
+	<ItemGroup>
+		<AssemblyAttribute Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions.ExtensionInformationAttribute">
+			<_Parameter1>Microsoft.Azure.WebJobs.Extensions.Sql</_Parameter1>
+			<_Parameter2>0.*-*</_Parameter2>
+		</AssemblyAttribute>
+	</ItemGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
```diff
@@ -22,4 +22,10 @@
 		<None Include=".\README.md" Pack="true" PackagePath="" />
 	</ItemGroup>
 
-</Project>
+	<ItemGroup>
+		<AssemblyAttribute Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions.ExtensionInformationAttribute">
+			<_Parameter1>Microsoft.Azure.WebJobs.Extensions.Sql</_Parameter1>
+			<_Parameter2>0.*-*</_Parameter2>
```
Charles-Gagnon, line 28: This doesn't resolve all the tasks listed in the issue : 

1. The published extension version is now 1.*, so this should be updated correct? (which demonstrates the usefulness of it being here in the first place...)
2. You should also be creating a pipeline var and then using that instead of hardcoding it https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/build-release.yml#L13 (defaulting to 99.99.99 so it continues to work for local builds)

Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
```diff
@@ -7,11 +7,12 @@
 		<Product>SQL Binding Worker</Product>
 		<!--Version information-->
 		<VersionPrefix>99.99.99</VersionPrefix>
+		<SupportedVersion>1.*-*</SupportedVersion>
```
Charles-Gagnon, line 10: This should be 99.99.99 to match the local dev package version

(and it would be even better to make a top level prop that is used by both this and the extension project)

Charles-Gagnon: Well, this goes into the whole "testing locally should use the local version of packages" concept. So it shouldn't be trying to pull down from nuget.org - that's the whole point. And the fact that it does that currently shows that we're not using the correct version of the extension when running our tests which is problematic. 

So what we should be aiming for here is to have : 

1. The supported version be 99.99.99 when built locally (overridden in pipelines with whatever the new version is)
2. When you run using the local worker it is able to pick up the local nuget package. This will mean the tests need to know how to load the local version of the package (which likely means a nuget.config pointing to the bin folder as a source)

MaddyDev: Looking into adding a nuget config and make the tests use the local version of the package. Can I get this in for the pipeline variable changes? Or close this out and address all of them together in one PR?

Charles-Gagnon: Let's do it as a follow up item - I suspect it'll be a bit more annoying and no reason to delay this since we're not regressing anything. Could you make an issue to track getting that done?

Charles-Gagnon (approved):
Make sure you get a passing pipeline build, failure looks transient but you should re-run to make sure.

MaddyDev:
Created the following to track https://github.com/Azure/azure-functions-sql-extension/pull/621

PR #627, opened 2023-01-13
Update README.md

lucyzhang929:
Add note about the current datetime format behavior difference between C# in-process and other languages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which
 
 ## Known Issues
 - The table used by a SQL binding or SQL trigger cannot contain two columns that only differ by casing (Ex. 'Name' and 'name'). 
+- For C# in process functions using input or output bindings, any columns of the types 'DATE', 'DATETIME', 'DATETIME2', 'DATETIMEOFFSET', or 'SMALLDATETIME' will get converted to ISO8061 format ("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ"). This does not happen for functions written in C# out-of-proc or other languages.
 
 ### Output Bindings
 - Output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` are not supported and data upserts will fail. These types [will be removed](https://docs.microsoft.com/sql/t-sql/data-types/ntext-text-and-image-transact-sql) in a future version of SQL Server and are not compatible with the `OPENJSON` function used by this Azure Functions binding.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +55,6 @@ Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which
 - PowerShell Functions using Output bindings cannot pass in null or empty values via the query string. The workaround is to use the `$TriggerMetadata[$keyName]` to retrieve the query property, an example can be found [here](https://github.com/Azure/azure-functions-sql-extension/blob/main/samples/samples-powershell/AddProductParams/run.ps1). Issue is tracked [here](https://github.com/Azure/azure-functions-powershell-worker/issues/895).
 
 ### Input Bindings
-- Input bindings against tables with columns of data types 'DATETIME', 'DATETIME2', or 'SMALLDATETIME' will assume that the values are in UTC format.
```
lucyzhang929, line 57: Removing this since Datetime types dont have time zone information. SQL DatetimeOffset, which includes the time zone, gets converted to C# DatetimeOffset type correctly.

VasuBhog (approved):

README.md
```diff
@@ -44,6 +44,7 @@ Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which
 
 ### Output Bindings
 - Output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` are not supported and data upserts will fail. These types [will be removed](https://docs.microsoft.com/sql/t-sql/data-types/ntext-text-and-image-transact-sql) in a future version of SQL Server and are not compatible with the `OPENJSON` function used by this Azure Functions binding.
+- Output bindings against tables with columns of data types `DATE`, `DATETIME`, `DATETIME2`, `DATETIMEOFFSET`, or `SMALLDATETIME` will convert those values to ISO8061 format ("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ"). This does not happen for functions written in C# out-of-proc or other languages.
```
Charles-Gagnon, line 47:
```suggestion
- .NET In-Proc output bindings against tables with columns of data types `DATE`, `DATETIME`, `DATETIME2`, `DATETIMEOFFSET`, or `SMALLDATETIME` will convert values for those columns to ISO8061 format ("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ") before upsertion. This does not happen for functions written in C# out-of-proc or other languages.
```

Is this correct though about only being for .NET In-Proc? I thought upserting always went through the same code path...

README.md
Charles-Gagnon: But when we convert the types don't we end up with time values that are in UTC (because they don't have a timezone offset to go off of)? This is just telling the user that if they have those types we can't tell what the timezone is so it assumes UTC. 

Maybe we should just add another line explaining that if the timestamp being stored isn't UTC they should use datetimeoffset instead?

README.md
lucyzhang929: It is going through the same path but I believe the DateFormatString setting in JsonSerializerSettings only applies to Datetime objects and not the rows coming in as Json from out of proc language workers.
https://github.com/Azure/azure-functions-sql-extension/blob/349eaa77b9059d2809dc6317895993679fe048df/src/SqlAsyncCollector.cs#L420-L424
https://github.com/Azure/azure-functions-sql-extension/blob/349eaa77b9059d2809dc6317895993679fe048df/src/SqlAsyncCollector.cs#L373

README.md
lucyzhang929: Hm I might be misunderstanding something here but since the CSharp Datetime type doesn't have a time zone field, when we store a DATETIME value from the SQL table in a CSharp Datetime object, it technically isn't in either UTC or local format. We are just storing the datetime in the object as is and not doing any time zone conversions (ex. we're not taking a local datetime and converting it to UTC). 

I think using DATETIME vs DATETIMEOFFSET is up to the user when creating their database and not necessarily an issue we should note with sql bindings.

Charles-Gagnon: Ok, I think that makes sense. One of these days I need to actually spend time playing around with this myself though...

Charles-Gagnon (approved):

Issue #624, opened 2023-01-12
Sending exception event: Invalid column name '<COLUMN NAME>'


NathZ1:
Hi team, I have encountered a strange error which i believe is something to do with the way table information is cached on first load. 

I can replicate it as follows: 

I have a very simple table:

```
CREATE TABLE [dbo].[last_loaded](
	[ID] [int] NOT NULL,
	[IEMR] [datetime2](3) NULL,
	[PFM] [datetime2](3) NULL,
	[ELMS] [datetime2](3) NULL,
	[PPT] [datetime2](3) NULL,
	[CONFIG] [datetime2](3) NULL,
	[IEMR_5MIN] [datetime2](3) NULL,
	[IEMR_10MIN] [datetime2](3) NULL,
	[IEMR_15MIN] [datetime2](3) NULL,
 CONSTRAINT [PK_last_loaded] PRIMARY KEY CLUSTERED 
(
	[ID] ASC
)WITH (STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
) ON [PRIMARY]
```

This table always has a single row of data:

![image](https://user-images.githubusercontent.com/3673816/211958259-a32730a6-5c17-4618-9e25-f13c86a27495.png)

My front end app sends a request of to the Function app with a body to update either the CONFIG or PPT datetime:

![image](https://user-images.githubusercontent.com/3673816/211958758-b760f940-9353-4f8c-80b4-5cb401d15a58.png)

![image](https://user-images.githubusercontent.com/3673816/211958705-805631c8-6af5-4ca2-adc9-04660f35d907.png)

The function is called addLastLoaded and is coded:

```
{
  "bindings": [
    {
      "name": "req",
      "authLevel": "function",
      "type": "httpTrigger",
      "direction": "in",
      "methods": [
        "put"
      ]
    },
    {
      "name": "res",
      "type": "http",
      "direction": "out"
    },
    {
      "name": "lastLoaded",
      "type": "sql",
      "direction": "out",
      "commandText": "dbo.last_loaded",
      "commandType": "Text",
      "connectionStringSetting": "SqlConnectionString"
    }
  ],
  "scriptFile": "../dist/addLastLoaded/index.js"
}
```

```
import { AzureFunction, Context, HttpRequest } from "@azure/functions"
import { isUserAuth } from "../utilities"

[...]

PR #628, opened 2023-01-13
Don't cache merge or insert query

lucyzhang929:
Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/624

Previously we were caching the merge/insert query for 10 minutes. This causes an error when the objects being upserted do not have the same properties.

For example: 
Running AddProduct in JavaScript with
{“ProductId”: 0, “Name”: “first”, “Cost”: 100}
And subsequently with
{“ProductId”: 0, “Name”: “second”}

The update part of the merge statement was cached and still includes "Cost", causing an error during the second function call:
"WITH cte AS ( SELECT * FROM OPENJSON(@rowData) WITH ([ProductId] int,[Name] nvarchar(100)) ) \n                    MERGE INTO [dbo].[Products] WITH (HOLDLOCK)\n                        AS ExistingData\n                    USING cte\n                        AS NewData\n                    ON\n                        ExistingData.[ProductId] = NewData.[ProductId]\n                    WHEN MATCHED THEN\n                        **UPDATE SET  ExistingData.[ProductId] = NewData.[ProductId], ExistingData.[Name] = NewData.[Name], ExistingData.[Cost] = NewData.[Cost]\n                    WHEN NOT MATCHED THEN\n                        INSERT ([ProductId],[Name],[Cost]) VALUES ([ProductId],[Name],[Cost])**;"

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     .Where(prop => !tableInfo.PrimaryKeys.Any(k => k.IsIdentity && string.Equals(k.Name, prop, StringComparison.Ordinal))) // Skip any identity columns, those should never be updated
                     .Select(prop => prop.AsBracketQuotedString());
                 var table = new SqlObject(fullTableName);
-                string tableInfoQuery = tableInfo.QueryType == QueryType.Insert ? TableInformation.GetInsertQuery(table, bracketedColumnNamesFromItem) :
+                string mergeOrInsertQuery = tableInfo.QueryType == QueryType.Insert ? TableInformation.GetInsertQuery(table, bracketedColumnNamesFromItem) :
                     TableInformation.GetMergeQuery(tableInfo.PrimaryKeys, table, bracketedColumnNamesFromItem);
 
                 TelemetryInstance.TrackEvent(TelemetryEventName.UpsertStart, props);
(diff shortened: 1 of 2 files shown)


src/SqlAsyncCollector.cs
```diff
@@ -379,7 +392,9 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
 
         public class TableInformation
         {
-            public IEnumerable<PropertyInfo> PrimaryKeys { get; }
+            public List<PrimaryKey> PrimaryKeys { get; }
```
Charles-Gagnon, line 395: It would be nice to follow this up with merging these into a single Tuple list - avoids confusion about why we'd need to have two separate lists that both contain a value for each primary key.

Charles-Gagnon (approved):

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -447,5 +447,34 @@ public void AddProductIncorrectCasing(SupportedLanguages lang)
             Assert.Throws<AggregateException>(() => this.SendOutputGetRequest("addproduct-incorrectcasing").Wait());
             Assert.Equal(0, this.ExecuteScalar("SELECT COUNT(*) FROM Products"));
         }
+
+        /// <summary>
+        /// Tests that subsequent upserts work correctly when the object properties are different from the first upsert.
+        ///
+        [Theory]
+        [SqlInlineData()]
+        public void AddProductWithDifferentPropertiesTest(SupportedLanguages lang)
```
MaddyDev, line 456: AddProductWithDifferentPropertiesTest is confusing, AddConsecutiveProductsTest ?

Charles-Gagnon: It's not just about adding consecutive products though (we do that in other tests). It's specifically testing that calling the function with different sets of properties still results in the row being updated. So I think the name is fine (although if you can think of a better one Lucy go ahead)

MaddyDev (approved):

chlafreniere (approved):

test/Integration/SqlOutputBindingIntegrationTests.cs
MaddyDev: At first look, I was looking for different properties and was confused and then read the test comments and understood. It's just a nit, we can move on if you can't come up with a better name :)

PR #629, opened 2023-01-13
Update nuget config with nuget.local

### LINKED ISSUE #623: Use local extension package for OOP tests
Currently the tests pull the sql extension from nuget to test, update this to pick up the local package. This will mean the tests need to know how to load the local version of the package (which likely means a nuget.config pointing to the bin folder as a source)

MaddyDev:
Add new nuget source for referencing the locally generated sql package with version 99.99.99 for tests.
Fixes #623

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -39,5 +39,12 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-
+  <Target Name="CopyNupkg" AfterTargets="Build">
+    <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -39,5 +39,12 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-
+  <Target Name="CopyNupkg" AfterTargets="Build">
+    <ItemGroup>
+      <_Packages Include=".\bin\$(Configuration)\*.*" />
+      <_Packages Include=".\obj\$(Configuration)\*.*" />
```
Charles-Gagnon, line 45: You should only copy the nuget packages, not all files (and especially not ones from the obj folder, that doesn't have anything we need)

```suggestion
      <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
```

Charles-Gagnon (approved):

chlafreniere (approved):

Issue #473, opened 2022-11-10
Finish doc TODOs - Java


Charles-Gagnon:
Finish TODO sections in https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsGuide_Java.md

PR #633, opened 2023-01-19
Update SetupGuide_Java.md

lucyzhang929:
Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/473

### DIFF
--- a/docs/SetupGuide_Java.md
+++ b/docs/SetupGuide_Java.md
@@ -47,7 +47,7 @@ These instructions will guide you through creating your Function App and adding 
     <dependency>
         <groupId>com.microsoft.azure.functions</groupId>
         <artifactId>azure-functions-java-library-sql</artifactId>
-        <version>0.1.1</version>
+        <version>[0.1.1,)</version>
     </dependency>
     ```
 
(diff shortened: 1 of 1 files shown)


docs/SetupGuide_Java.md
```diff
@@ -42,6 +42,14 @@ These instructions will guide you through creating your Function App and adding
         "version": "[4.*, 5.0.0)"
     }
     ```
+    Add the Java library for SQL bindings to the pom.xml file.
+    ```xml
+    <dependency>
+        <groupId>com.microsoft.azure.functions</groupId>
+        <artifactId>azure-functions-java-library-sql</artifactId>
+        <version>0.1.1</version>
```
chlafreniere, line 50: How will we keep this version up to date? Is this documented in release instructions?

Alternatively, can we use wildcards to avoid putting an explicit version here?

lucyzhang929: Good point - I can add this to release instructions (we already have a note to update Microsoft docs). The library itself should not be updated very often.

Charles-Gagnon (approved):

docs/SetupGuide_Java.md
lucyzhang929: Updated to `[0.1.1,)` so that it can pick up new versions.

VasuBhog (approved):

Issue #634, opened 2023-01-19
Add all SQL data types to ProductsColumnTypes tests


lucyzhang929:
- [x] binary and varbinary
- [ ] Spatial types (geometry and geography)

lucyzhang929:
Created separate issue to track spatial types: https://github.com/Azure/azure-functions-sql-extension/issues/654

PR #635, opened 2023-01-19
Add more datatypes to ProductsColumnTypes tests

lucyzhang929:
Tracking missing types here: https://github.com/Azure/azure-functions-sql-extension/issues/634

AddProductColumnsTest is being skipped for Java because of datetime format issue on Linux (test is passing on Windows):
https://github.com/Azure/azure-functions-sql-extension/issues/521. Will investigate and document any workarounds for this.


samples/samples-python/Common/productcolumntypes.py
```diff
@@ -1,11 +0,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
lucyzhang929, line 1: ProductColumnTypes is only used in test-python folder.

Charles-Gagnon (approved):

Issue #423, opened 2022-10-25
Java input and output bindings do not work with binary and varbinary type


lucyzhang929:
The following errors are thrown when trying to map Binary and Varbinary columns to Byte[] in Java.
Input binding:
```
Exception: IllegalStateException: Expected BEGIN_ARRAY but was STRING at line 1 column 479 path $[0].Binary
Stack: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was STRING at line 1 column 479 path $[0].Binary
```
Output binding:
```
System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Newtonsoft.Json: Error reading string. Unexpected token: StartArray. Path 'Binary', line 24, position 13.
```

The errors occur because Gson uses a json array while Newtonsoft uses a json string https://stackoverflow.com/questions/12109536/cant-deserialize-byte-using-gson.

PR #641, opened 2023-01-24
Add binary types to tests

lucyzhang929:
Normally the sql binary type maps to a byte array. But Newtonsoft serializes the byte array as a base64 encoded string when sending it to the language workers. So non-CSharp languages need to map binary types to a string (added a note to the known issues section of the readme).

If we try to use a byte array in a non-CSharp output binding, we get the following error because Newtonsoft expects a string:
`System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Newtonsoft.Json: Error reading string. Unexpected token: StartArray. Path 'Binary', line 23, position 13.`

Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/423

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which
 
 ## Known Issues
 - The table used by a SQL binding or SQL trigger cannot contain two columns that only differ by casing (Ex. 'Name' and 'name'). 
+- Non-CSharp functions using SQL bindings against tables with columns of data types `BINARY` or `VARBINARY` need to map those columns to a string type. Input bindings will return the binary value as a base64 encoded string. Output bindings require the value upserted to binary columns to be a base64 encoded string.
 
 ### Output Bindings
 - Output bindings against tables with columns of data types `NTEXT`, `TEXT`, or `IMAGE` are not supported and data upserts will fail. These types [will be removed](https://docs.microsoft.com/sql/t-sql/data-types/ntext-text-and-image-transact-sql) in a future version of SQL Server and are not compatible with the `OPENJSON` function used by this Azure Functions binding.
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -41,6 +41,7 @@ Databases on SQL Server, Azure SQL Database, or Azure SQL Managed Instance which
 
 ## Known Issues
 - The table used by a SQL binding or SQL trigger cannot contain two columns that only differ by casing (Ex. 'Name' and 'name'). 
+- Non-CSharp functions using SQL bindings against tables with columns of data types `BINARY` or `VARBINARY` need to map those columns to a string type. Input bindings will return the binary value as a base64 encoded string. Output bindings require the value upserted to binary columns to be a base64 encoded string.
```
Charles-Gagnon, line 44: Shouldn't this be the below? If they turn it into a string themselves then it'll stay as a string, right?

```suggestion
- Non-CSharp functions using SQL bindings against tables with columns of data types `BINARY` or `VARBINARY` need to map those columns to a string type. If not then input bindings will return the binary value as a base64 encoded string and output bindings require the value upserted to binary columns to be a base64 encoded string.
```

Charles-Gagnon (approved):

README.md
lucyzhang929: For input bindings, the binary values are sent to the function as a base64 string. For output bindings, the value being upserted needs to be converted to a base64 string.

If we try to use a byte array in a non-CSharp output binding, we get the following error because Newtonsoft expects a string:
System.Private.CoreLib: Exception while executing function: Functions.AddProductColumnTypes. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Newtonsoft.Json: Error reading string. Unexpected token: StartArray. Path 'Binary', line 23, position 13.

chlafreniere (approved):

PR #642, opened 2023-01-24
[LogicApps] Revert to MDS 3.1.1

Charles-Gagnon:
The latest merge from main brought in the update to SqlClient 5.x, but unfortunately this required updating the host runtime packages to v4 (see https://github.com/Azure/azure-functions-sql-extension/pull/424 for details). 

v4 then requires Newtonsoft >=13.x, which we can't have yet here. 

So this PR is reverting MDS back to 3.1.1 (which is a slight bump over the previous version of 3.0.1 - but wanting to take the latest MDS 3.x version to ensure we have all the latest fixes and improvements)


Charles-Gagnon:
Optimistically merging this in again...will continue to follow up if there are still test issues

PR #653, opened 2023-01-30
Sql Server Edition telemetry

### LINKED ISSUE #326: Add connection type information for telemetry data
Need to identify how users are connecting to sql server while using sql bindings.

MaddyDev:
Fixes #326

Query the server to get the engine edition and log that to telemetry along with connection props.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
         private readonly List<T> _rows = new List<T>();
         private readonly SemaphoreSlim _rowLock = new SemaphoreSlim(1, 1);
-        private static ServerProperties _engineEdition;
+        private static ServerProperties _serverProperties;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqlAsyncCollector<typeparamref name="T"/>"/> class.
@@ -179,8 +179,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 this._logger.LogDebugWithThreadId("BEGIN OpenUpsertRowsAsyncConnection");
                 await connection.OpenAsync();
                 this._logger.LogDebugWithThreadId("END OpenUpsertRowsAsyncConnection");
-                _engineEdition = await GetSqlServerEditions(connection, this._logger, CancellationToken.None);
-                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps(_engineEdition);
+                _serverProperties = await GetSqlServerEditions(connection, this._logger, CancellationToken.None);
+                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps(_serverProperties);
 
                 string fullTableName = attribute.CommandText;
 
@@ -544,7 +544,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
             {
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(_engineEdition);
+                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(_serverProperties);
                 logger.LogDebugWithThreadId("BEGIN RetrieveTableInformationAsync");
                 var table = new SqlObject(fullName);
 
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
         private readonly List<T> _rows = new List<T>();
         private readonly SemaphoreSlim _rowLock = new SemaphoreSlim(1, 1);
-        private static ServerProperties _serverProperties;
+        private ServerProperties _serverProperties;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SqlAsyncCollector<typeparamref name="T"/>"/> class.
@@ -179,8 +179,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 this._logger.LogDebugWithThreadId("BEGIN OpenUpsertRowsAsyncConnection");
                 await connection.OpenAsync();
                 this._logger.LogDebugWithThreadId("END OpenUpsertRowsAsyncConnection");
-                _serverProperties = await GetSqlServerEditions(connection, this._logger, CancellationToken.None);
-                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps(_serverProperties);
+                this._serverProperties = await GetServerTelemetryProperties(connection, this._logger, CancellationToken.None);
+                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps(this._serverProperties);
 
                 string fullTableName = attribute.CommandText;
 
@@ -544,7 +544,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
             {
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(_serverProperties);
+                // pass null for ServerProperties since it's already logged before this call.
+                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(null);
                 logger.LogDebugWithThreadId("BEGIN RetrieveTableInformationAsync");
                 var table = new SqlObject(fullName);
 
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -178,7 +178,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 this._logger.LogDebugWithThreadId("BEGIN OpenUpsertRowsAsyncConnection");
                 await connection.OpenAsync();
                 this._logger.LogDebugWithThreadId("END OpenUpsertRowsAsyncConnection");
-                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps();
+                string engineEdition = await GetSqlServerEdition(connection, this._logger, CancellationToken.None);
+                Dictionary<TelemetryPropertyName, string> props = connection.AsConnectionProps(engineEdition);
 
                 string fullTableName = attribute.CommandText;
 
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -208,7 +208,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 {
                     TelemetryInstance.TrackEvent(TelemetryEventName.TableInfoCacheMiss, props);
                     // set the columnNames for supporting T as JObject since it doesn't have columns in the member info.
-                    tableInfo = await TableInformation.RetrieveTableInformationAsync(connection, fullTableName, this._logger, GetColumnNamesFromItem(rows.First()));
+                    tableInfo = await TableInformation.RetrieveTableInformationAsync(connection, fullTableName, this._logger, GetColumnNamesFromItem(rows.First()), this._serverProperties);
                     var policy = new CacheItemPolicy
                     {
                         // Re-look up the primary key(s) after timeout (default timeout is 10 minutes)
@@ -542,10 +542,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <param name="logger">ILogger used to log any errors or warnings.</param>
             /// <param name="objectColumnNames">Column names from the object</param>
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
-            public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
+            public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames, ServerProperties serverProperties)
             {
                 // pass null for ServerProperties since it's already logged before this call.
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(null);
+                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(serverProperties);
                 logger.LogDebugWithThreadId("BEGIN RetrieveTableInformationAsync");
                 var table = new SqlObject(fullName);
 
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -543,7 +543,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
             {
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps();
+                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(EngineEdition.NotQueried.ToString());
                 logger.LogDebugWithThreadId("BEGIN RetrieveTableInformationAsync");
                 var table = new SqlObject(fullName);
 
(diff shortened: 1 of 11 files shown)


src/SqlBindingUtilities.cs
```diff
@@ -290,5 +290,51 @@ internal static bool IsFatalSqlException(this Exception e)
             }
             return true;
         }
+
+        /// <summary>
+        /// Query the ServerProperty('EngineEdition') for logging the Server Type for Telemetry
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Throw if an error occurs while querying the engine edition</exception>
+        public static async Task<string> GetSqlServerEdition(SqlConnection connection, ILogger logger, CancellationToken cancellationToken)
```
chlafreniere, line 298: If this is only used for telemetry, then we probably should only be sending these queries if a user hasn't opted out of telemetry. Thoughts?

MaddyDev: I didn't know users can opt out of telemetry, since this is an overhead call on top of the function  calls, we should skip if users opted out of telemetry.

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +290,51 @@ internal static bool IsFatalSqlException(this Exception e)
...
+                        return SqlBindingConstants.EngineEdition.SQLDatabase.ToString();
+                    case 6:
+                        return SqlBindingConstants.EngineEdition.AzureSynapseAnalytics.ToString();
+                    case 8:
+                        return SqlBindingConstants.EngineEdition.AzureSQLManagedInstance.ToString();
+                    case 9:
+                        return SqlBindingConstants.EngineEdition.AzureSQLEdge.ToString();
+                    case 11:
+                        return SqlBindingConstants.EngineEdition.AzureSynapseserverlessSQLpool.ToString();
+                    default: return "Unknown";
```
Charles-Gagnon, line 334: Should return the stringified number here instead in this case

src/Telemetry/TelemetryUtils.cs
```diff
@@ -13,20 +13,21 @@ public static class TelemetryUtils
...
+            props.Add(TelemetryPropertyName.EngineEdition, string.IsNullOrEmpty(engineEdition) ? "Unknown" : engineEdition);
         }
 
         /// <summary>
         /// Returns a dictionary with common connection properties for the specified connection.
         /// </summary>
         /// <param name="conn">The connection to get properties of</param>
         /// <returns>The property dictionary</returns>
-        public static Dictionary<TelemetryPropertyName, string> AsConnectionProps(this SqlConnection conn)
+        public static Dictionary<TelemetryPropertyName, string> AsConnectionProps(this SqlConnection conn, string engineEdition = null)
```
Charles-Gagnon, line 27: Why are you giving this a default value? Just make it required

Charles-Gagnon, line 27: Update XML doc

src/Telemetry/TelemetryUtils.cs
```diff
@@ -13,20 +13,21 @@ public static class TelemetryUtils
         /// </summary>
         /// <param name="props">The property bag to add our connection properties to</param>
         /// <param name="conn">The connection to add properties of</param>
-        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn)
+        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn, string engineEdition)
```
Charles-Gagnon, line 16: Update XML doc

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +290,51 @@ internal static bool IsFatalSqlException(this Exception e)
...
             return true;
         }
+
+        /// <summary>
+        /// Query the ServerProperty('EngineEdition') for logging the Server Type for Telemetry
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Throw if an error occurs while querying the engine edition</exception>
+        public static async Task<string> GetSqlServerEdition(SqlConnection connection, ILogger logger, CancellationToken cancellationToken)
+        {
+            string selectServerEdition = $"SELECT ServerProperty('EngineEdition')";
```
Charles-Gagnon, line 300: We should also get the `SERVERPROPERTY('Edition')` to get the server edition as well (Azure vs On-Prem mostly)

Charles-Gagnon, line 300: nit: Built in functions are all upper cased by convention
```suggestion
            string selectServerEdition = $"SELECT SERVERPROPERTY('EngineEdition')";
```

src/SqlAsyncCollector.cs
```diff
@@ -542,7 +543,7 @@ AS NewData
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
             {
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps();
+                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps(EngineEdition.NotQueried.ToString());
```
Charles-Gagnon, line 546: It would be preferable to have this (and any other places that use this "NotQueried" value) to instead have the engine edition and other fields populated once on startup with the initial connection and then added to all events after that.

We may want to look into when/where we send these connection properties at a later point - since right now it seems a bit random (maybe they should even be common properties attached to all events). But for now I'd just say in each of these classes populate it once on the "initial" connection and then use that value as needed in the below places. 

We shouldn't ever send a "NotQueried" value for the engine edition - just don't send it if we don't need/want to add it.

src/SqlBindingUtilities.cs
MaddyDev: Edition returns the below list, not sure if that'll be any more insightful than EngineEdition:
image

Charles-Gagnon: Edition != Engine Edition. They are separate things, engine edition is per database and edition is per server. Both are useful to have.

MaddyDev: So Edition = Server edition and EngineEdition = Database Engine edition.
@chlafreniere  @dzsquared Want me to add both now?

src/SqlAsyncCollector.cs
```diff
@@ -102,6 +103,7 @@ public SqlAsyncCollector(IConfiguration configuration, SqlAttribute attribute, I
                 connection.OpenAsyncWithSqlErrorHandling(CancellationToken.None).Wait();
                 this._logger.LogDebugWithThreadId("END OpenSqlAsyncCollectorVerifyDatabaseSupportedConnection");
                 VerifyDatabaseSupported(connection, logger, CancellationToken.None).Wait();
+
```
VasuBhog, line 106: nit extra space

src/Telemetry/TelemetryUtils.cs
```diff
@@ -13,20 +13,24 @@ public static class TelemetryUtils
         /// </summary>
         /// <param name="props">The property bag to add our connection properties to</param>
         /// <param name="conn">The connection to add properties of</param>
-        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn)
+        /// <param name="engineEdition">The Engine Edition of the target Sql Server</param>
+        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn, ServerProperties engineEditions)
         {
             props.Add(TelemetryPropertyName.ServerVersion, conn.ServerVersion);
+            props.Add(TelemetryPropertyName.EngineEdition, engineEditions?.EngineEdition ?? "Unknown");
```
Charles-Gagnon, line 20: Just skip adding these if you don't have a server properties object

src/Telemetry/TelemetryUtils.cs
```diff
@@ -13,20 +13,24 @@ public static class TelemetryUtils
...
+            props.Add(TelemetryPropertyName.Edition, engineEditions?.Edition ?? "Unknown");
         }
 
         /// <summary>
         /// Returns a dictionary with common connection properties for the specified connection.
         /// </summary>
         /// <param name="conn">The connection to get properties of</param>
+        /// <param name="engineEdition">The Engine Edition of the target Sql Server</param>
```
Charles-Gagnon, line 28: Update this to be serverProperties everywhere (and comments too)

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +291,67 @@ internal static bool IsFatalSqlException(this Exception e)
             }
             return true;
         }
+
+        /// <summary>
+        /// Query the ServerProperty('EngineEdition') for logging the Server Type for Telemetry
```
Charles-Gagnon, line 296:
```suggestion
        /// Get the Server Properties for the given connection.
```

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +291,67 @@ internal static bool IsFatalSqlException(this Exception e)
...
+            {
+                string selectServerEdition = $"SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('EngineEdition')";
+
+                logger.LogDebugWithThreadId($"BEGIN GetSqlServerEdition Query={selectServerEdition}");
+                using (var selectServerEditionCommand = new SqlCommand(selectServerEdition, connection))
+                using (SqlDataReader reader = await selectServerEditionCommand.ExecuteReaderAsync(cancellationToken))
+                {
+                    if (!await reader.ReadAsync(cancellationToken))
+                    {
+                        throw new InvalidOperationException($"Received empty response when querying the engine edition.");
```
Charles-Gagnon, line 311: I don't think we should throw in this case - since it's only for telemetry we want to continue on even if an error occurs (it just won't include these values in the telemetry which is fine enough)

src/SqlBindingConstants.cs
```diff
@@ -6,5 +6,28 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
...
+            Express,
+            SQLDatabase,
+            AzureSynapseAnalytics,
+            AzureSQLManagedInstance,
+            AzureSQLEdge,
+            AzureSynapseserverlessSQLpool,
+        }
+    }
+
+    public class ServerProperties
```
Charles-Gagnon, line 28: Put this in the Telemetry file for now, it isn't really a "constant"

src/SqlBindingUtilities.cs
MaddyDev: Added it, we can clean up if it's not required.

src/SqlAsyncCollector.cs
```diff
@@ -73,6 +73,7 @@ internal class SqlAsyncCollector<T> : IAsyncCollector<T>, IDisposable
 
         private readonly List<T> _rows = new List<T>();
         private readonly SemaphoreSlim _rowLock = new SemaphoreSlim(1, 1);
+        private static ServerProperties _serverProperties;
```
Charles-Gagnon, line 76: This shouldn't need to be static - and would cause problems if there's multiple functions running in the same host but connecting to different servers.

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +292,65 @@ internal static bool IsFatalSqlException(this Exception e)
             }
             return true;
         }
+
+        /// <summary>
+        /// Get the Server Properties for the given connection.
+        /// </summary>
+        /// <returns>ServerProperties (EngineEdition and Edition) of the target Sql Server.</returns>
+        public static async Task<ServerProperties> GetSqlServerEditions(SqlConnection connection, ILogger logger, CancellationToken cancellationToken)
```
Charles-Gagnon, line 300: `GetServerTelemetryProperties`

(update the logging statements and other variable names too)

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +292,65 @@ internal static bool IsFatalSqlException(this Exception e)
...
+        {
+            if (TelemetryInstance.Enabled)
+            {
+                string selectServerEdition = $"SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('EngineEdition')";
+
+                logger.LogDebugWithThreadId($"BEGIN GetSqlServerEdition Query={selectServerEdition}");
+                using (var selectServerEditionCommand = new SqlCommand(selectServerEdition, connection))
```
Charles-Gagnon, line 307:
```suggestion
                string serverPropertiesQuery = $"SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('EngineEdition')";

                logger.LogDebugWithThreadId($"BEGIN GetSqlServerEdition Query={serverPropertiesQuery}");
                using (var selectServerEditionCommand = new SqlCommand(serverPropertiesQuery, connection))
```

src/Telemetry/Telemetry.cs
```diff
@@ -426,5 +428,11 @@ public enum TelemetryErrorName
         Upsert,
         UpsertRollback,
     }
+
+    public class ServerProperties
```
Charles-Gagnon, line 432: Can this be internal too?

src/Telemetry/TelemetryUtils.cs
```diff
@@ -13,20 +13,30 @@ public static class TelemetryUtils
...
         /// <param name="props">The property bag to add our connection properties to</param>
         /// <param name="conn">The connection to add properties of</param>
-        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn)
+        /// <param name="serverProperties">Server Properties (EngineEdition, Edition) of the target Sql Server queried for a given connection.</param>
+        public static void AddConnectionProps(this IDictionary<TelemetryPropertyName, string> props, SqlConnection conn, ServerProperties serverProperties)
         {
             props.Add(TelemetryPropertyName.ServerVersion, conn.ServerVersion);
+            if (!string.IsNullOrEmpty(serverProperties?.EngineEdition))
+            {
+                props.Add(TelemetryPropertyName.EngineEdition, serverProperties.Edition);
```
Charles-Gagnon, line 22:
```suggestion
                props.Add(TelemetryPropertyName.EngineEdition, serverProperties.EngineEdition);
```

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +292,65 @@ internal static bool IsFatalSqlException(this Exception e)
...
+
+        /// <summary>
+        /// Get the Server Properties for the given connection.
+        /// </summary>
+        /// <returns>ServerProperties (EngineEdition and Edition) of the target Sql Server.</returns>
+        public static async Task<ServerProperties> GetSqlServerEditions(SqlConnection connection, ILogger logger, CancellationToken cancellationToken)
+        {
+            if (TelemetryInstance.Enabled)
+            {
+                string selectServerEdition = $"SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('EngineEdition')";
```
Charles-Gagnon, line 304: Put all this in a try/catch - if an exception happens log it then just return null

src/Telemetry/Telemetry.cs
MaddyDev: I started with internal, there was an error in SqlBindingingUtilities class that cannot access the object so made it public

Charles-Gagnon: Was the error along the lines of "type has more visible modifier"? Because you'd need to make the method internal as well since it returns an internal type.

src/SqlAsyncCollector.cs
```diff
@@ -542,7 +544,8 @@ AS NewData
             /// <returns>TableInformation object containing primary keys, column types, etc.</returns>
             public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConnection sqlConnection, string fullName, ILogger logger, IEnumerable<string> objectColumnNames)
             {
-                Dictionary<TelemetryPropertyName, string> sqlConnProps = sqlConnection.AsConnectionProps();
+                // pass null for ServerProperties since it's already logged before this call.
```
Charles-Gagnon, line 547: I would still pass the connection params so that anywhere we add any connection params we add as many as we can - no reason to leave some out some of the time.

MaddyDev: connection params? That's read from the connection object itself, I don't need to pass it explicitly, am I missing something here?

Charles-Gagnon: Whoops, meant server properties 😄

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +292,73 @@ internal static bool IsFatalSqlException(this Exception e)
...
+        {
+            if (TelemetryInstance.Enabled)
+            {
+                string serverPropertiesQuery = $"SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('EngineEdition')";
+
+                logger.LogDebugWithThreadId($"BEGIN GetServerTelemetryProperties Query={serverPropertiesQuery}");
+                using (var selectServerEditionCommand = new SqlCommand(serverPropertiesQuery, connection))
+                using (SqlDataReader reader = await selectServerEditionCommand.ExecuteReaderAsync(cancellationToken))
+                {
+                    try
```
Charles-Gagnon, line 310: Try try/catch should go around everything starting right after the if - the main thing being to catch possible exceptions from ExecuteReaderAsync

src/SqlBindingUtilities.cs
```diff
@@ -290,5 +292,73 @@ internal static bool IsFatalSqlException(this Exception e)
...
+                                    return serverProperties;
+                                default:
+                                    serverProperties.EngineEdition = engineEdition.ToString(System.Globalization.CultureInfo.InvariantCulture);
+                                    return serverProperties;
+                            }
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        logger.LogError($"Exception in GetServerTelemetryProperties. Exception = {ex.Message}");
```
Charles-Gagnon, line 355: Send a telemetry exception event for this too

src/SqlBindingUtilities.cs
MaddyDev: I thought I added it after if, my bad. Thanks for catch :)

Charles-Gagnon (approved):

VasuBhog (approved):

PR #661, opened 2023-02-03
Worker unit tests

MaddyDev:
I was torn between adding a new worker unit tests project and adding them to the existing test-outofproc project. Decided to reuse the existing one since the tests are minimal and didn't want to create an extra project for them.

### DIFF
--- a/test/Common/TestUtils.cs
+++ b/test/Common/TestUtils.cs
@@ -196,7 +196,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
         /// <summary>
         /// Returns a task that will complete when either the original task completes or the specified timeout is reached.
         /// </summary>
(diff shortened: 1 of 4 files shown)


test/Common/TestUtils.cs
```diff
@@ -171,7 +171,6 @@ public static string CleanJsonString(string jsonStr)
         /// <summary>
         /// Returns a task that will complete when either the original task completes or the specified timeout is reached.
         /// </summary>
-        /// <typeparam name="TResult"></typeparam>
```
Charles-Gagnon, line 174: Nice, I had this change staged locally and was going to push it later today 😄

Charles-Gagnon (approved):
Good as a first step, but please follow up with adding other cases similar to the existing ones we have for input/output bindings.

lucyzhang929 (approved):

PR #665, opened 2023-02-06
Use local packages for OOP tests in pipeline

Charles-Gagnon:
Partial fix for https://github.com/Azure/azure-functions-sql-extension/issues/623

This updates the pipelines to use the locally built extension DLL for tests, instead of the latest Nuget.org version. 

The steps now look like this : 

1. Set up dotnet nuget source for local-packages folder
2. Build "test" version of DLLs. These are the same as the release version except that the Worker extension will have the exact binary version (.e.g `1.0.124` from the pipeline run as its "supported version" instead of something like `1.*`) so that when the OOP test project is built it pulls in that nupkg from the local-pacakages folder (since that version shouldn't exist on Nuget.org yet)
3. Run tests
4. Rebuild project, this time with `1.*` supported version

This should solve the issues with our tests not running against the local version of the package for OOP. 

Additional things : 

1. This does not solve the local build issue noted in the comments for the original issue
2. This does not solve the issue with our tests not catching breaking issues in the bundle. The OOP stuff doesn't even use the bundle so it's a separate issue there. 

nupkg dependencies before change:
image

nupkg dependencies after change:
image

Assembly references before change:
image

Assembly references after change:
image

### DIFF
--- a/Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
+++ b/Worker.Extensions.Sql/src/Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj
@@ -23,6 +23,10 @@
 		<None Include=".\README.md" Pack="true" PackagePath="" />
 	</ItemGroup>
 
+    <ItemGroup>
+        <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Extensions.Sql.csproj" />
+    </ItemGroup>
+
 	<ItemGroup>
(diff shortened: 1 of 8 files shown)


Charles-Gagnon:
Converting to draft just so I can do some more verification to make sure the versions and dependencies are all set up correctly. Feel free to take a look if you want, otherwise I'll ping when it's ready for full review.

Charles-Gagnon:
Ok, did a test build and all the dependencies look good now so this is ready for review

@MaddyDev @lucyzhang929 @chlafreniere @VasuBhog

MaddyDev (approved):

lucyzhang929 (approved):

Issue #270, opened 2022-06-15
Make error message clearer when using unsupported column types


Charles-Gagnon:
We don't support NTEXT, TEXT, or IMAGE columns in Output bindings (they aren't supported by OPENJSON and are going to be deprecated). 

https://github.com/Azure/azure-functions-sql-extension#known-issues

Currently if a binding with one of those is used it just gives a generic error similar to `Incorrect syntax near '*'` which isn't very helpful. We should detect when columns are of any of the above types and throw a specific exception explaining to the user what the issue is.

PR #666, opened 2023-02-07
Throw a more helpful error message for unsupported types

lucyzhang929:
Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/270


src/SqlAsyncCollector.cs
```diff
@@ -231,7 +232,15 @@ private async Task UpsertRowsAsync(IList<T> rows, SqlAttribute attribute, IConfi
                     throw ex;
                 }
 
-                IEnumerable<string> bracketedColumnNamesFromItem = GetColumnNamesFromItem(rows.First())
+                IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
+                IEnumerable<string> unsupportedColumns = columnNamesFromItem.Where(prop => IsUnsupportedType(prop, tableInfo.Columns));
+                if (unsupportedColumns.Any())
+                {
+                    string message = $"The type(s) of the following column(s) are not supported: {string.Join(", ", unsupportedColumns.ToArray())}.";
```
Charles-Gagnon, line 239:
```suggestion
                    string message = $"The type(s) of the following column(s) are not supported: {string.Join(", ", unsupportedColumns.ToArray())}. See https://github.com/Azure/azure-functions-sql-extension#output-bindings for more details.";
```

src/SqlAsyncCollector.cs
```diff
@@ -292,6 +301,18 @@ private async Task UpsertRowsAsync(IList<T> rows, SqlAttribute attribute, IConfi
             }
         }
 
+        private static bool IsUnsupportedType(string property, IDictionary<string, string> columns)
+        {
+            foreach (string unsupportedType in UnsupportedTypes)
+            {
+                if (columns[property].StartsWith(unsupportedType, StringComparison.OrdinalIgnoreCase))
```
Charles-Gagnon, line 308: Is there a reason to use StartsWith instead of comparing the whole string?

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -472,5 +472,26 @@ public void AddProductWithDifferentPropertiesTest(SupportedLanguages lang)
...
+        /// </summary>
+        [Theory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.OutOfProc)]
+        public async Task AddProductUnsupportedTypesTest(SupportedLanguages lang)
+        {
+            var foundExpectedMessageSource = new TaskCompletionSource<bool>();
+            this.StartFunctionHost(nameof(AddProductUnsupportedTypes), lang, true, (object sender, DataReceivedEventArgs e) =>
+            {
+                if (e.Data.Contains("The type(s) of the following column(s) are not supported: Text, Ntext, Image."))
```
Charles-Gagnon, line 487: nit: I was initially confused since it seemed to say the type instead of the column names. So maybe name the columns something like `TextCol` just to make it clear

Charles-Gagnon (approved):

src/SqlAsyncCollector.cs
lucyzhang929: The column type also includes the size, for example text(10).

Charles-Gagnon: Can you have sizes for these types? I tried and it gave me an error

image

lucyzhang929: Hm, the value I got while testing was text(*) so I thought there can be sizes. Maybe the column type returned will always be text(*), ntext(*), and image(*). I'll double check.

lucyzhang929: The `(*)` gets added on because of this line in the query https://github.com/Azure/azure-functions-sql-extension/blob/61ca7215d75e8c6e092aabcea3f75a17f7d7f059/src/SqlAsyncCollector.cs#L511 I updated the UnsupportedTypes to include the `(*)` and simplified the check.

MaddyDev (approved):

PR #668, opened 2023-02-07
BatchSize -> MaxBatchSize

Charles-Gagnon:
Got some feedback recently about BatchSize being confusing since it's really the max batch size - the trigger will send as many changes as it has each time up to the max batch size specified. So updating the config override value to match. 

Old value will still be read if the new value isn't specified for backwards compat support. I'll open a tracking issue for removing this for GA since removing it at that point should be ok.


src/Telemetry/Telemetry.cs
```diff
@@ -383,6 +383,7 @@ public enum TelemetryMeasureName
         GetPrimaryKeysDurationMs,
         GetUnprocessedChangesDurationMs,
         InsertGlobalStateTableRowDurationMs,
+        MaxBatchSize,
```
lucyzhang929, line 386: Can BatchSize in line 375 be removed?

lucyzhang929 (approved):

src/Telemetry/Telemetry.cs
Charles-Gagnon: That's being used for the AsyncCollector which also has a batch size (but that's hardcoded to 1000 and is a limitation of how many items can be inserted at once)

Charles-Gagnon:
Ok finally got the failing test - and fixed the issue causing it to hang instead of actually failing too!

MaddyDev (approved):

PR #671, opened 2023-02-08
Add local-packages as nuget source

### LINKED ISSUE #623: Use local extension package for OOP tests

MaddyDev:
The below error is seen when trying to build the samples (samples-outofproc specifically)
WorkerExtensions.csproj : error NU1102: Unable to find package Microsoft.Azure.WebJobs.Extensions.Sql with version (>= 99.99.99)

Even though nuget.local has been added as a source pointing to the local Sql extension with version 99.99.99, out of proc functions spins up a new worker project in temp folder which is outside the repo and nuget.local isn't visible and is failing with the above error.

Fix: 
1. Add the local-packages folder as a nuget source from home directory (C drive on windows or /Users/{userprofile} on mac/linux)
2. Remove the local-packages folder every time we build to avoid using the cached package.

Fixes  #623

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,9 +30,11 @@ code .
 
 4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/GeneralSetup.md#configure-function-app)
 
-5. Configure a local nuget source to run the out-of-proc functions with locally built extension.
-    Add local nuget source from your **home directory** (or a directory above the project repo).
+5. Configure a local nuget source that will be used to build the .NET Out-of-Proc sample with the latest locally built extension package.
 
+Building the sample will cause a project to be created and built in the temp directory, which means it does not pick up on the nuget.config for this project. Without adding the source at the global level that project will fail to build since it will be looking for the dev-local version of the package (99.99.99).
+
+Note: This command must be ran from your **home directory** (or a directory outside the project repo).
     ```powershell
     dotnet nuget add source <PATH_TO_REPO_AZURE-FUNCTIONS-SQL-EXTENSION/local-packages>
     ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ code .
 4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/GeneralSetup.md#configure-function-app)
 
 5. Configure a local nuget source to run the out-of-proc functions with locally built extension.
-    Add local nuget source from your home directory.
+    Add local nuget source from your **home directory** (or a directory above the project repo).
 
     ```powershell
     dotnet nuget add source <PATH_TO_REPO_AZURE-FUNCTIONS-SQL-EXTENSION/local-packages>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,4 +30,11 @@ code .
 
 4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/GeneralSetup.md#configure-function-app)
 
-5. Press F5 to run SQL bindings samples that are included in this repo. The output window should display startup information as well as the function endpoints that were started.
+5. Configure a local nuget source to run the out-of-proc functions with locally built extension.
+    Add local nuget source from your home directory.
+
+    ```powershell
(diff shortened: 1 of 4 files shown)


nuget.config
```diff
@@ -1,7 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <clear />
```
Charles-Gagnon, line 4: We can't remove this - this is a security thing (so you're always sure where you get packages from when building the project)

I'm not sure why this would be necessary to remove in the first place though - the only two places we should be pulling packages from are the two right below so that should be all that's needed for building the projects in this repo.

MaddyDev: WorkerExtensions.csproj is not part of the repo and hence the issue. If I remove that, it will not be able to see the new source added and will fail with the same error.
@lucyzhang929 Can you test and confirm this since am still not able to repro the issue and figuring it out?

MaddyDev: Resolved this offline through teams. Thanks again @Charles-Gagnon for explaining how nuget commands work.

lucyzhang929 (approved):

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -46,4 +46,8 @@
     <Copy SourceFiles="@(_Packages)" DestinationFolder="../local-packages" />
     <Message Text="Copied sql .nupkg to local-packages" Importance="high" />
   </Target>
+  <Target Name="RemoveLocalPackages" BeforeTargets="Build" Condition="'$(Configuration)' == 'Debug'">
```
Charles-Gagnon, line 49: Why are you removing this folder anyways? The folder isn't the problem - the package there will get overwritten during the build. It's the nuget cache (%USERPROFILE%/.nuget for Windows) that's the thing that needs to be cleared out.

docs/SetupGuide_DotnetOutOfProc.md
```diff
@@ -35,6 +36,28 @@ Please refer to the functions documentation [here](https://learn.microsoft.com/a
 - There's also no direct support for types inherited from underlying service SDKs, such as SqlCommand. Instead, bindings rely on strings, arrays, and serializable types, such as plain old class objects (POCOs).
 - For HTTP triggers, you must use HttpRequestData and HttpResponseData to access the request and response data. This is because you don't have access to the original HTTP request and response objects when running out-of-process.
 
+## Setup Function App
```
Charles-Gagnon, line 39: This seems unrelated to this PR - could you just do it as a separate PR? It should be quick to get in since it's just a doc update.

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
MaddyDev: $(NugetPackageRoot)/$(PackageId.ToLower())/99.99.99 should do?

Charles-Gagnon, line 49: Making the name match what it's actually doing and removing the debug condition (it should run regardless, there's no reason to only do it for debug builds)

```suggestion
  <Target Name="RemoveCachedExtensionPackage" BeforeTargets="Build">
```

CONTRIBUTING.md
```diff
@@ -30,4 +30,11 @@ code .
 
 4. Configure the Function App located in the [samples](./samples) folder by following the instructions [here](./docs/GeneralSetup.md#configure-function-app)
 
-5. Press F5 to run SQL bindings samples that are included in this repo. The output window should display startup information as well as the function endpoints that were started.
\ No newline at end of file
+5. Configure a local nuget source to run the out-of-proc functions with locally built extension.
+    Add local nuget source from your **home directory** (or a directory above the project repo).
```
Charles-Gagnon, line 34:
```suggestion
5. Configure a local nuget source that will be used to build the .NET Out-of-Proc sample with the latest locally built extension package. 

Building the sample will cause a project to be created and built in the temp directory, which means it does not pick up on the nuget.config for this project. Without adding the source at the global level that project will fail to build since it will be looking for the dev-local version of the package (99.99.99).

Note: This command must be ran from your **home directory** (or a directory outside the project repo).
```

Issue #495, opened 2022-11-14
Update Python setup instructions


lucyzhang929:
Update Python setup instructions and Python issue (https://github.com/Azure/azure-functions-sql-extension/issues/250) after testing/verifying python functions with SQL Bindings are working with latest Python worker. Also remove the     "PYTHON_ISOLATE_WORKER_DEPENDENCIES" flag from out python samples.

PR #678, opened 2023-02-10
Remove pinned python library version in python sample

lucyzhang929:
The pinned version is no longer needed since the latest azure-functions library contains sql. 

Once a new core tools version is out we will be able to remove the "PYTHON_ISOLATE_WORKER_DEPENDENCIES" flag from local.settings.json. Updated this [issue](https://github.com/Azure/azure-functions-sql-extension/issues/495) to track this.

### DIFF
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -128,6 +128,11 @@ steps:
   workingDirectory: $(Build.SourcesDirectory)/samples/samples-python
   displayName: Lint samples-python
 
+- script: |
+    pip3 install -r requirements.txt
+  workingDirectory: $(Build.SourcesDirectory)/samples/samples-python
+  displayName: Install samples-python dependencies
+
 - task: Maven@3
   displayName: Build Java Samples
   inputs:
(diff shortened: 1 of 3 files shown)


samples/samples-python/requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Do not include azure-functions-worker as it may conflict with the Azure Functions platform
 
-azure-functions==1.11.3b1
\ No newline at end of file
+azure-functions
```
Charles-Gagnon, line 3: There's a couple other places we reference this - should remove those too right? And then do we have this in the Learn docs anywhere?

lucyzhang929: Sounds good, thanks Charles!

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -128,11 +128,6 @@ steps:
   workingDirectory: $(Build.SourcesDirectory)/samples/samples-python
   displayName: Lint samples-python
 
-- script: |
-    pip3 install -r requirements.txt
```
Charles-Gagnon, line 132: Don't we still need to install the azure-functions library? It just won't be pinned anymore (so you'd just update this to have a name like `Install samples-python dependencies`)

chlafreniere (approved):

Charles-Gagnon (approved):

VasuBhog (approved):

PR #682, opened 2023-02-14
Make attribute properties readonly + cleanup

Charles-Gagnon:
Continuing cleanup of some stuff I noticed while updating the attributes before. Will help to make it easier to discover the properties we make available and avoid potential weird issues down the line. 

1. Made attribute properties readonly - there's no reason they should be updated outside of the constructor.
1. Move CommandType and Parameters into constructor as needed
1. Update/clarify docs

### DIFF
--- a/Worker.Extensions.Sql/src/SqlInputAttribute.cs
+++ b/Worker.Extensions.Sql/src/SqlInputAttribute.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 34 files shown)


lucyzhang929 (approved):

MaddyDev (approved):

Charles-Gagnon:
Pipeline failure was due to a separate problem I opened an issue for : https://github.com/Azure/azure-functions-sql-extension/issues/687 - the rest of the jobs passed so it should be fine.

Issue #690, opened 2023-02-21
Issue with column order in the sample project


JetterMcTedder:
When deploying the sample project, and entering a value into the comments field and submitting it; seeing this error:

**[2023-02-20T23:48:36.702Z] Executed 'Functions.AddEntry' (Failed, Id=89433610-213f-4d90-a3cb-1a7976579014, Duration=535ms)
[2023-02-20T23:48:36.705Z] System.Private.CoreLib: Exception while executing function: Functions.AddEntry. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Core Microsoft SqlClient Data Provider: Conversion failed when converting the nvarchar value 'my comment' to data type bit.**

Looking at the column order in this file

https://github.com/Azure-Samples/sql-bindings-guestbook/blob/main/apis/js-api/AddEntry/index.js

        newRow = {
            TextEntry: req.body.newEntry,
            DisableView: false,
            DateCreated: date.toISOString(),
            DateModerated: null
        };
 
 
I switched the TextEntry and the DisableView and the above error went away.
 
        newRow = {
            **DisableView: false,
            TextEntry: req.body.newEntry,**
            DateCreated: date.toISOString(),
            DateModerated: null
        };

I can only guess that the binding is column sensitive?

Charles-Gagnon:
Am I correct in assuming that the table schema has the columns in the order of the second data blob? (DisableView followed by TextEntry)

JetterMcTedder:
Correct:


```
CREATE TABLE [app].[Entry]
(
  [Id] INT NOT NULL IDENTITY(1,1) PRIMARY KEY,
  [DisableView] BIT NOT NULL DEFAULT(0),
  [TextEntry] NVARCHAR(MAX) NOT NULL,
  [DateCreated] DATETIME NOT NULL DEFAULT(GETDATE()),
  [DateModerated] DATETIME NULL
);
GO
```

lucyzhang929:
@Charles-Gagnon I can take a look if you haven't already.

Charles-Gagnon:
That would be great! Let me know if you need anything

lucyzhang929:
@JetterMcTedder A new extension bundle version (v4.6.0) has been released with this fix. Your function should automatically pick up the latest version if you have the following in your host.json:
```
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
[...]

PR #693, opened 2023-02-22
Fix column name order issue in insert query

lucyzhang929:
This PR fixes: https://github.com/Azure/azure-functions-sql-extension/issues/690.

The issue was introduced in this PR: https://github.com/Azure/azure-functions-sql-extension/pull/388/files#diff-2ccba51b9e200b38816cfd9cf11f06fea1f771e3f2105321fbe6acb9c305d30b

Previously the order of the column names in the [newDataQuery](https://github.com/Azure/azure-functions-sql-extension/blob/137a14c3f5446362c990d3e04da27966aef5f759/src/SqlAsyncCollector.cs#L411) did not match the order of the column names in the [insert query.](https://github.com/Azure/azure-functions-sql-extension/blob/137a14c3f5446362c990d3e04da27966aef5f759/src/SqlAsyncCollector.cs#L517)


src/SqlAsyncCollector.cs
```diff
@@ -403,8 +403,11 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
 
             rowData = JsonConvert.SerializeObject(rowsToUpsert, table.JsonSerializerSettings);
             IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
-            IEnumerable<string> bracketColumnDefinitionsFromItem = table.Columns.Where(c => columnNamesFromItem.Contains(c.Key))
-                .Select(c => $"{c.Key.AsBracketQuotedString()} {c.Value}");
+            var bracketColumnDefinitionsFromItem = new List<string>();
```
lucyzhang929, line 406: Previously `bracketColumnDefinitionsFromItem`'s order was based on the SQL table (`table.Columns`). Instead it should be based on the order of the POCO or JSON item (`columnNamesFromItem`) to match the insert part of the query.

src/SqlAsyncCollector.cs
```diff
@@ -403,8 +403,11 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
...
+            foreach (string columnName in columnNamesFromItem)
+            {
+                bracketColumnDefinitionsFromItem.Add($"{columnName.AsBracketQuotedString()} {table.Columns[columnName]}");
+            }
```
Charles-Gagnon, line 410: (might want to do this one yourself, easy to make typos in Github...)

```suggestion
            IEnumerable<string> bracketColumnDefinitionsFromItem = columnNamesFromItem.Select(c => $"{c.AsBracketQuotedString()} {table.Columns[columnName]}");
            }
```

Charles-Gagnon (approved):

MaddyDev (approved):

PR #699, opened 2023-02-28
Add csharp script samples

### LINKED ISSUE #218: Support Azure Function C# Script(CSX)  binding
Hi, there:
      We tried to use Azure function SQL extension for Azure Function C# script binding, but it doesn't work after many attempts. Below is what we have done:
    Please add sql extension support for Azure Function C# Script(CSX).
    Thanks,

Li

1. Add Microsoft.Azure.WebJobs.Extensions.Sql into function.proj and got the error - see below:

image

2. Try to add SQL binding into function.json - there is no document of how to add them, we added them by referencing to event hub binding - azure function gives no logs and nothing works, our understanding is if the function.json has some error, function host will not work:
   {
      "name": "outputData",
      "direction": "out",
      "type": "SQL",
      "connectionStringSetting": "SQLConnectionString",
      "sqlServerName": "odpsqlserver",
      "databaseName": "ODPUserDB"
    }

MaddyDev:
Added samples for csharp scritping language.

Fixes #218


docs/SetupGuide_Dotnet.md
```diff
@@ -399,6 +401,33 @@ public static IActionResult Run(
     return new CreatedResult($"/api/addproduct", product);
 }
 ```
+## CSharp Scripting
```
Charles-Gagnon, line 404: Let's move this under `Setup Function Project` and have two subsections there - one for traditional .NET and the other for CSX

samples/samples-csharpscript/host.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": "2.0",
```
Charles-Gagnon, line 2: Do we need these duplicate top level files? What happens if you delete them?

samples/samples-csharpscript/product.csx
```diff
@@ -0,0 +1,63 @@
+public class Product
```
Charles-Gagnon, line 1: Does this actually need to be a .csx or can it just be a .cs? It doesn't seem to use any scripting functionality. 

I'd also suggest putting it in a "Common" folder

Charles-Gagnon, line 1: Add copyright header to all these

docs/SetupGuide_Dotnet.md
```diff
@@ -21,6 +21,8 @@
       - [ICollector\<T\>/IAsyncCollector\<T\>](#icollectortiasynccollectort)
       - [Array](#array)
       - [Single Row](#single-row)
+  - [CSharp Scripting](#csharp-scripting)
```
Charles-Gagnon, line 24: Also add links to the specific samples under each of the samples sections in this file

Charles-Gagnon (commented):
Was the plan to not add tests at this time?

MaddyDev:
> Was the plan to not add tests at this time?

Yes, I've created #700 to track it.

samples/samples-csharpscript/product.csx
MaddyDev: It's an example of how to use scripting files and include them, if you do .cs the way we include it in the run.csx file is using statement.

docs/SetupGuide_Dotnet.md
MaddyDev: subsections for just setup instructions or for everything we follow that pattern?

Charles-Gagnon: Everything where there's differences between the two

Charles-Gagnon: Right now the way it's structured will be confusing I think because users will come in and start at the top which tells them how to create a normal .NET function project - and only when they finally get down here will they see what to do for C# scripting.

Instead I think it would be clearer to just have each section with differences for the two types to have a subsection for each instead (in the case of the samples probably fine just to have two links in each since it's just one line)

MaddyDev: okay after contemplating on how to structure this, I felt it would be way cleaner to have a separate file for csx and added it.

Charles-Gagnon (approved):

lucyzhang929 (approved):

Issue #721, opened 2023-03-10
Add Trigger tests for all languages


lucyzhang929:
- [x] JavaScript
- [x] PowerShell
- [x] Python
- [x] Java
- [x] OOP
- [x] CSX

PR #722, opened 2023-03-10
Add JObject support for SQL trigger

lucyzhang929:
Closes: https://github.com/Azure/azure-functions-sql-extension/issues/352, https://github.com/Azure/azure-functions-sql-extension/issues/353, https://github.com/Azure/azure-functions-sql-extension/issues/354
Added ProductsTrigger sample for JavaScript, PowerShell, and Python.

Created a separate [issue](https://github.com/Azure/azure-functions-sql-extension/issues/721) to track adding tests for all languages.

### DIFF
--- a/src/TriggerBinding/SqlTriggerBindingProvider.cs
+++ b/src/TriggerBinding/SqlTriggerBindingProvider.cs
@@ -10,6 +10,7 @@ using Microsoft.Azure.WebJobs.Host.Triggers;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Sql
 {
@@ -65,18 +66,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 return Task.FromResult<ITriggerBinding>(null);
             }
 
-            if (!IsValidTriggerParameterType(parameter.ParameterType))
+            Type parameterType = parameter.ParameterType;
+            if (!IsValidTriggerParameterType(parameterType))
             {
-                throw new InvalidOperationException($"Can't bind SqlTriggerAttribute to type {parameter.ParameterType}." +
-                    " Only IReadOnlyList<SqlChange<T>> is supported, where T is the type of user-defined POCO that" +
-                    " matches the schema of the user table");
+                throw new InvalidOperationException($"Can't bind SqlTriggerAttribute to type {parameter.ParameterType}.");
             }
 
             string connectionString = SqlBindingUtilities.GetConnectionString(attribute.ConnectionStringSetting, this._configuration);
 
-            // Extract the POCO type 'T' and use it to instantiate class 'SqlTriggerBinding<T>'.
-            Type userType = parameter.ParameterType.GetGenericArguments()[0].GetGenericArguments()[0];
-            Type bindingType = typeof(SqlTriggerBinding<>).MakeGenericType(userType);
+            Type bindingType;
+            if (parameterType == typeof(string))
+            {
+                bindingType = typeof(SqlTriggerBinding<>).MakeGenericType(typeof(JObject));
+            }
+            else
+            {
+                // Extract the POCO type 'T' and use it to instantiate class 'SqlTriggerBinding<T>'.
+                Type userType = parameter.ParameterType.GetGenericArguments()[0].GetGenericArguments()[0];
+                bindingType = typeof(SqlTriggerBinding<>).MakeGenericType(userType);
+            }
 
             var constructorParameterTypes = new Type[] { typeof(string), typeof(string), typeof(ParameterInfo), typeof(IHostIdProvider), typeof(ILogger), typeof(IConfiguration) };
             ConstructorInfo bindingConstructor = bindingType.GetConstructor(constructorParameterTypes);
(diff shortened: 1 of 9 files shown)


src/TriggerBinding/SqlTriggerValueProvider.cs
```diff
@@ -24,21 +27,37 @@ internal class SqlTriggerValueProvider : IValueProvider
...
             this._value = value;
             this._tableName = tableName;
+            this._isString = parameterType == typeof(string);
         }
 
         /// <summary>
         /// Gets the trigger argument value.
         /// </summary>
-        public Type Type { get; }
+        public Type Type
```
lucyzhang929, line 39: Followed the CosmosDB extension implementation [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Trigger/CosmosdbTriggerValueBinder.cs.). They also support JArray but the languages I tested so far (JS, PS, Python) are only using String. We can add JArray later if needed for other languages/scenarios.

src/TriggerBinding/SqlTriggerValueProvider.cs
```diff
@@ -24,21 +27,37 @@ internal class SqlTriggerValueProvider : IValueProvider
...
+        }
 
         /// <summary>
         /// Returns value of the trigger argument.
         /// </summary>
         public Task<object> GetValueAsync()
         {
+            if (this._isString)
+            {
+                return Task.FromResult<object>(JArray.FromObject(this._value).ToString(Newtonsoft.Json.Formatting.None));
```
chlafreniere, line 58: Just clarifying that the use of JArray here is expected

src/TriggerBinding/SqlTriggerBindingProvider.cs
```diff
@@ -65,18 +66,25 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
 
             string connectionString = SqlBindingUtilities.GetConnectionString(attribute.ConnectionStringSetting, this._configuration);
 
-            // Extract the POCO type 'T' and use it to instantiate class 'SqlTriggerBinding<T>'.
-            Type userType = parameter.ParameterType.GetGenericArguments()[0].GetGenericArguments()[0];
-            Type bindingType = typeof(SqlTriggerBinding<>).MakeGenericType(userType);
+            Type bindingType;
+            if (parameterType == typeof(string))
+            {
+                bindingType = typeof(SqlTriggerBinding<>).MakeGenericType(typeof(JObject));
```
Charles-Gagnon, line 80: Add comment explaining why parameter type of string maps to binding type of JObject

src/TriggerBinding/SqlTriggerValueProvider.cs
Charles-Gagnon, line 58: You can't use `JArray.FromObject` directly or we'll run into the issues I fixed previously with picking up global default settings.

https://github.com/Azure/azure-functions-sql-extension/pull/703

You should use the overload that takes in a serializer (and ideally put this with the other helpers in Utils.cs) : https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JArray.cs#L202

Charles-Gagnon (changes requested):

src/TriggerBinding/SqlTriggerValueProvider.cs
chlafreniere: do we have tests for this for input/output bindings already? (i.e. non-default global settings)

Let's create an issue either for this to create tests to ensure we don't regress this scenario -- either just for trigger (if tests for bindings already exist), or for both.

Charles-Gagnon: I added https://github.com/Azure/azure-functions-sql-extension/blob/main/test/Integration/test-csharp/Startup.cs - so we should be getting validation of that. But that only covers the functions in the test folder so we might need/want to expand it to be applied to our "sample" functions as well (but only during testing) if we find that the test folder coverage isn't enough.

Charles-Gagnon: Now that I think about it - I believe technically this doesn't need to follow the same logic since this code should only be hit for non-C# languages. But those can't override the default Newtonsoft settings - so we should be safe. But I also don't think it hurts to do this and might save us in case there's an edge case where it can happen (for example, I suppose it's possible another extension changes the default settings....).

lucyzhang929: Updated to use Utils.JsonSerializeObject instead of converting to JArray and using ToString(). I tested the different languages and the results were the same. This way the conversion from JObject to JSON string is consistent between input binding and trigger.

src/TriggerBinding/SqlTriggerBindingProvider.cs
```diff
@@ -65,18 +66,26 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
             }
 
-            if (!IsValidTriggerParameterType(parameter.ParameterType))
+            Type parameterType = parameter.ParameterType;
+            if (!IsValidTriggerParameterType(parameterType))
             {
-                throw new InvalidOperationException($"Can't bind SqlTriggerAttribute to type {parameter.ParameterType}." +
-                    " Only IReadOnlyList<SqlChange<T>> is supported, where T is the type of user-defined POCO that" +
-                    " matches the schema of the user table");
+                throw new InvalidOperationException($"Can't bind SqlTriggerAttribute to type {parameter.ParameterType}.");
```
Charles-Gagnon, line 72: I kind of liked the previous one that said very directly what was supported, but I suppose now it'd be a bit too verbose. Some more context would be good at least though I think.

```suggestion
                throw new InvalidOperationException($"Can't bind SqlTriggerAttribute to type {parameter.ParameterType}, this is not a supported type.");
```

Charles-Gagnon (commented):
The changes look good, but there are some test failures and a lint failure still so waiting for those to get fixed.

Charles-Gagnon (approved):

VasuBhog (approved):

Charles-Gagnon:
@lucyzhang929 Switched this to target release/trigger branch, should be good to go now but not a bad idea to wait for PR run to complete just to make sure

PR #725, opened 2023-03-13
GA package changes - add ReleaseType parameter 

MaddyDev:
added a ReleaseType parameter following the Dacfx pipeline definition and used that to set the nugetVersion with/without preview tag.


builds/azure-pipelines/build-release.yml
```diff
@@ -18,7 +27,7 @@ variables:
   versionMajorMinor: '$(versionMajor).$(versionMinor)'  # This variable is only used for the counter so we reset properly when either major or minor is bumped
   versionPatch: $[counter(variables['versionMajorMinor'], 0)] # This will reset when we bump minor version
   binariesVersion: '$(versionMajor).$(versionMinor).$(versionPatch)'
-  nugetVersion: '$(binariesVersion)-preview'
+  nugetVersion: $[replace(replace('True', eq(parameters.ReleaseType, 'full'), '$(binariesVersion)'),'True','$(binariesVersion)-preview')] # This will set the preview tag to the version based on the ReleaseType parameter.
```
MaddyDev, line 30: There is no ternary operator in ADO expressions, so used the [replace](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#replace) as a hack for similar effect.

Charles-Gagnon (approved):

PR #729, opened 2023-03-14
accessing the parameter syntax

MaddyDev:


MaddyDev:
merging this in since it's blocking the product build due to the yaml error.

Issue #723, opened 2023-03-13
Invalid operation. The connection is closed


rvbaast:
We are seeing the issue again as stated here : #593 
We see that the the first time the connection fails it seems to recover correctly, the second time it occurs it doesn't.
I added the logs of that second time below. ( Table name is redacted )
If more information is needed i am happy to provide it.

version used : 1.0.122-preview

logs : 

|timestamp       |message                                                                                                                                  |severityLevel|itemType|
|----------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|--------|
|11/03/2023 15:17|TID:20 BEGIN ProcessingChanges State=CheckingForChanges   |0            |trace   |
|11/03/2023 15:17|TID:20 BEGIN GetTableChanges                              |0            |trace   |
[...]

PR #731, opened 2023-03-14
Fix trigger connection not being retried when connection is closed

Charles-Gagnon:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/723

Turns out that the "Connection is closed" exception is just an InvalidOperationException so the checks I added weren't catching that. So to fix it also checking the actual connection state, since if that's closed/broken then we know something happened and can hopefully recover from it. 

While I was doing this I realized the "is broken or closed" check I had previously wasn't even working...HasFlags checks if ALL the flag bits are set that you give it (so a connection would have to be both closed AND broken to be true). And using HasFlags for Closed isn't even correct since its value is 0 so will always return true. I don't know why this is a flags enum to begin with but just checking the State directly is what we actually want to be doing here. (I didn't notice this because the connection retry was being forced if a fatal SQL exception occurred, which was the thing being fixed with the original logic)

### DIFF
--- a/src/SqlBindingUtilities.cs
+++ b/src/SqlBindingUtilities.cs
@@ -237,7 +237,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         }
 
         /// <summary>
-        /// Checks whether an exception is a fatal SqlException. It is deteremined to be fatal
+        /// Checks whether an exception is a fatal SqlException. It is determined to be fatal
         /// if the Class value of the Exception is 20 or higher, see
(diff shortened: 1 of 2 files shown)


MaddyDev (approved):

chlafreniere (approved):
Wouldn't complain if you added a couple of tests though 😄

Charles-Gagnon:
I was trying to think of a way to do that cleanly but don't have a good solution. I'll see what I can come up with as a follow up though since we definitely want to have *something* for ensuring this behavior works as expected.

lucyzhang929 (approved):

PR #732, opened 2023-03-14
Remove trigger binding for GA release

Charles-Gagnon:
Removing from here in preparation for GA release of input/output bindings.


README.md
```diff
@@ -8,7 +8,6 @@
   - [Supported SQL Server Versions](#supported-sql-server-versions)
   - [Known Issues](#known-issues)
     - [Output Bindings](#output-bindings)
-    - [Trigger Bindings](#trigger-bindings)
```
chlafreniere, line 11: My assumption is that links from docs will need to be updated. Is this correct?

@dzsquared as FYI

src/SqlBindingConfigProvider.cs
```diff
@@ -35,10 +33,9 @@ internal class SqlBindingConfigProvider : IExtensionConfigProvider
         /// <exception cref="ArgumentNullException">
         /// Thrown if either parameter is null
         /// </exception>
-        public SqlBindingConfigProvider(IConfiguration configuration, IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory)
```
chlafreniere, line 38: Just confirming this change is intentional. Unfamiliar with IHostIdProvider

chlafreniere (approved):
Went through this, just unfamiliar with one piece of code. Let's see if the tests uncover anything!

src/SqlBindingConfigProvider.cs
Charles-Gagnon: Yes, that's just used to generate the function ID for the trigger Leases table.

tafs7:
@Charles-Gagnon just following this, and wanted to make sure that removing the trigger binding just means, "remove for this 1st GA release, which includes only the input/output bindings, but the trigger will still be developed and released at a later date"?

Or did you guys backtrack and decide to NOT move forward with the SQL trigger? This is a functionality we're currently starting to adopt for functionality in an app, and if it gets pulled before we release, we'd like to know so we can pivot to another option.

dzsquared:
@tafs7 good question - being super clear, this is **just** for a GA release with input/output bindings, and the trigger is moving forward. The trigger will remain in preview for a little while longer until it reaches GA, but it is moving forward!

PR #746, opened 2023-03-16
update oop supported package version

MaddyDev:
updated the supported version similar to nuget version.

### DIFF
--- a/builds/azure-pipelines/build-release.yml
+++ b/builds/azure-pipelines/build-release.yml
@@ -4,7 +4,7 @@ parameters: # parameters are shown up in ADO UI in a build queue time
 - name: 'ReleaseType'
   displayName: 'Type of release'
   type: string
-  default: full
+  default: preview
   values:
     - full
     - preview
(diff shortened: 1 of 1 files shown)


builds/azure-pipelines/build-release.yml
```diff
@@ -28,7 +28,7 @@ variables:
   versionPatch: $[counter(variables['versionMajorMinor'], 0)] # This will reset when we bump minor version
   binariesVersion: '$(versionMajor).$(versionMinor).$(versionPatch)'
   nugetVersion: $[replace(replace('True', eq('${{ parameters.ReleaseType }}', 'full'), '$(binariesVersion)'),'True','$(binariesVersion)-preview')] # This will set the preview tag to the version based on the ReleaseType parameter.
-  supportedVersion: '$(versionMajor).*-*'
+  supportedVersion: $[replace(replace('True', eq('${{ parameters.ReleaseType }}', 'full'), '$(versionMajor).*'),'True','$(versionMajor).*-*')] # This will set the expression to pull the preview vs full version based on the ReleaseType parameter.
```
Charles-Gagnon, line 31: It would be good to verify that 3.* won't also pick up preview versions (e.g. 3.0.143-preview)

It's not something we'll be running into currently, but in the future I can see us doing preview release of the same major version and we don't want this to cause problems then.

Charles-Gagnon (approved):

builds/azure-pipelines/build-release.yml
MaddyDev: So we'll be ahead but in the minor versions, okay will look into how we can add that check then. Thanks!

MaddyDev:
#747

PR #754, opened 2023-03-22
Enable debug logging for Python and PS samples

lucyzhang929:
The SingleOperationTriggerTest has been failing intermittently for Python and PS. Temporarily enabling debug logs to help with investigation.

### DIFF
--- a/samples/samples-js/host.json
+++ b/samples/samples-js/host.json
@@ -8,7 +8,7 @@
       }
     },
     "logLevel": {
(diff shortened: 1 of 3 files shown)


MaddyDev (approved):

Charles-Gagnon (approved):

lucyzhang929:
The test failed for JS in the last PR pipeline run so enabling debug logs for JS too.

PR #758, opened 2023-03-23
Update SetupGuide_PowerShell.md

gabesmsft:
Remove the commas from the hashtable. The commas result in script parsing errors. Removing them enables the Function to run without error.

### DIFF
--- a/docs/SetupGuide_PowerShell.md
+++ b/docs/SetupGuide_PowerShell.md
@@ -159,17 +159,17 @@ Note: This tutorial requires that a SQL database is setup as shown in [Create a 
 
     $req_body = @(
         @{
-            EmployeeId=1,
-            FirstName="Hello",
-            LastName="World",
-            Company="Microsoft",
+            EmployeeId=1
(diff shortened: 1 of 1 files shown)


VasuBhog:
Hey @gabesmsft, we could use `;` instead of a comma here to indicate the end of each hashtable item. 
Each of the samples has been tested through this and this seems to not have been updated. Example sample [here](https://github.com/Azure/azure-functions-sql-extension/blob/main/samples/samples-powershell/AddProductWithIdentityColumnIncluded/run.ps1#L14)

Charles-Gagnon:
@VasuBhog Does it make a difference here? Not having the ; is cleaner looking to me so this seems fine unless there's some reason you think that might cause issues/confusion

VasuBhog (approved):

Charles-Gagnon:
Gotcha, yeah if you have some time to go through and clean the rest of the samples up that'd be great! If not then please make an issue for tracking it

Charles-Gagnon (approved):

PR #768, opened 2023-03-24
Move Database folder out of CSharp folder

lucyzhang929:
This makes it easier for users using non-csharp functions to find the .sql files for our samples.


chlafreniere:
Do any docs need to updated as well?

chlafreniere (approved):

VasuBhog:
What do you guys think about adding a doc or comments to some of the samples (such as PowerShell) to indicate this is the database we used (with this new updated link)?

VasuBhog (approved):

PR #775, opened 2023-03-29
Increase build timeout from 90 to 120 minutes

lucyzhang929:
Temporarily increasing the timeout to 120 minutes to unblock builds while I work on decreasing our test run times.


MaddyDev (approved):

chlafreniere (approved):
Let's see if 120 is sufficient 😄

PR #780, opened 2023-03-31
Add detailed trigger docs

Charles-Gagnon:
Adding more info about the inner workings of the trigger - both to help onboard new devs and to help troubleshooting for both us and the customers. Will be continuing to add more to this as I go along, but this should be enough to at least get a good grasp on the main stages of execution it goes through.

Plus this includes fancy Mermaid flowcharts! See https://github.com/Azure/azure-functions-sql-extension/blob/chgagnon/triggerDocs/docs/TriggerBinding.md for the rendered view.

### DIFF
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -12,11 +12,7 @@
       - [Columns with Default Values](#columns-with-default-values)
     - [Retry support for Output Bindings](#retry-support-for-output-bindings)
   - [Trigger Binding](#trigger-binding)
-    - [Change Tracking](#change-tracking)
-    - [Functionality Overview](#functionality-overview)
-    - [Internal State Tables](#internal-state-tables)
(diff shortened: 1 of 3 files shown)


VasuBhog (approved):
I like the mermaid flowcharts. I also saw that we could have links within the flow chart if we wanted to point to specific documents within the chart so users can easily click on them.
https://mermaid.js.org/syntax/sequenceDiagram.html?#actor-menus

Charles-Gagnon:
I tried adding links but couldn't get them to work. Clicking on them would just show an error. This seems to be because of how Github renders the charts - it's done in an iFrame which then seems to block navigation (presumably because of XSS concerns). 

I'll keep trying though, I couldn't find any issues about it so maybe I was just doing it wrong.

lucyzhang929 (approved):

PR #783, opened 2023-04-03
Add SQLTrigger annotation to java library

lucyzhang929:
Updated the azure-functions-java-library-sql version to 2.0.0-preview. Will also be releasing version 1.0.0 (with only input and output annotations) for input/output GA.

### DIFF
--- a/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
+++ b/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
@@ -30,7 +30,7 @@ public @interface SQLInput {
     /**
      * Text or Stored Procedure.
      */
-    String commandType();
+    CommandType commandType();
 
     /**
      * Parameters to the query or stored procedure. This string must follow the format
--- a/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
+++ b/java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
@@ -25,12 +25,12 @@ public @interface SQLInput {
     /**
      * Query string or name of stored procedure to be run.
      */
-    String commandText() default "";
+    String commandText();
 
     /**
      * Text or Stored Procedure.
      */
-    String commandType() default "";
+    String commandType();
 
     /**
      * Parameters to the query or stored procedure. This string must follow the format
@@ -42,5 +42,5 @@ public @interface SQLInput {
     /**
      * Setting name for SQL connection string.
      */
-    String connectionStringSetting() default "";
+    String connectionStringSetting();
 }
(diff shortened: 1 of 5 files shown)


java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLTrigger.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Name of the table to watch for changes.
+    */
+    String tableName() default "";
+
+    /**
+     * Setting name for SQL connection string.
+    */
+    String connectionStringSetting() default "";
```
Charles-Gagnon, line 33: Do we need the defaults here or something? These are required so we shouldn't make it easy for users to not fill these out.

java-library/src/main/java/com/microsoft/azure/functions/sql/annotation/SQLInput.java
```diff
@@ -25,12 +25,12 @@
...
      * Query string or name of stored procedure to be run.
      */
-    String commandText() default "";
+    String commandText();
 
     /**
      * Text or Stored Procedure.
      */
-    String commandType() default "";
+    String commandType();
```
Charles-Gagnon, line 33: This we should default to Text if possible - that's the default we do for the other langs

lucyzhang929: I tried removing commandType from non-CSharp samples that include parameters and the parameters were not getting set correctly. When I debugged the [GetProducts](https://github.com/Azure/azure-functions-sql-extension/blob/main/samples/samples-js/GetProducts/function.json#L23) function, it hits this constructor https://github.com/Azure/azure-functions-sql-extension/blob/f01ffa47e887baaf5cd33cea8103f2c8c928a0b1/src/SqlAttribute.cs#L39 and sets the parameter to null. 

I think we need to add another constructor that includes parameters. I will send a PR out for that.

Right now our docs and samples still note that commandType is required in non-CSharp languages. https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-input?tabs=in-process&pivots=programming-language-javascript#configuration

Charles-Gagnon: Gotcha, yeah we can make that improvement. But you should still be able to give this a default value here, right? It'll end up having a value - it's just one the user doesn't necessarily need to specify.

Charles-Gagnon, line 33: Can we make this an enum?

lucyzhang929: For some reason when I put a default value (CommandType.Text), the GetProducts function still uses the SqlAttribute constructor that sets the parameters to null. Once I add a new constructor that sets parameters, I'll update the java library to include a default value for CommandType.

PR #785, opened 2023-04-04
[Trigger] Merge latest from main

Charles-Gagnon:
Keeping the trigger branch up to date with latest stuff from main


MaddyDev (approved):
Some tests failed again due to the sql package copy task again. Am looking into how we can ensure the sql package is copied for tests.

PR #787, opened 2023-04-04
Add SqlAttribute constructor with parameters

lucyzhang929:
Currently if we remove the commandType in non-CSharp functions, the parameters do not get set correctly because this [SqlAttribute constructor](https://github.com/Azure/azure-functions-sql-extension/blob/b3c84ebf0726b723387c6f3a32e805e4dcbbab64/src/SqlAttribute.cs#L39) is getting called and the parameters are set to null.
Adding a new constructor that has a default value for commandType but still sets the parameters.

### DIFF
--- a/src/SqlAttribute.cs
+++ b/src/SqlAttribute.cs
@@ -39,6 +39,14 @@ namespace Microsoft.Azure.WebJobs
         public SqlAttribute(string commandText, string connectionStringSetting) : this(commandText, connectionStringSetting, CommandType.Text, null) { }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="SqlAttribute"/> class with the default value for the CommandType.
+        /// </summary>
+        /// <param name="commandText">For an input binding, either a SQL query or stored procedure that will be run in the database. For an output binding, the table name to upsert the values to.</param>
+        /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
+        /// <param name="parameters">Specifies the parameters that will be used to execute the SQL query or stored procedure. See <see cref="Parameters"/> for more details.</param>
+        public SqlAttribute(string commandText, string connectionStringSetting, string parameters) : this(commandText, connectionStringSetting, CommandType.Text, parameters) { }
+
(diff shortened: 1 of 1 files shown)


src/SqlAttribute.cs
```diff
@@ -38,6 +38,14 @@ public SqlAttribute(string commandText, string connectionStringSetting, CommandT
         /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
         public SqlAttribute(string commandText, string connectionStringSetting) : this(commandText, connectionStringSetting, CommandType.Text, null) { }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SqlAttribute"/> class with the default value for the CommandType.
+        /// </summary>
+        /// <param name="commandText">For an input binding, either a SQL query or stored procedure that will be run in the database. For an output binding, the table name to upsert the values to.</param>
+        /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
+        /// <param name="parameters">Specifies the parameters that will be used to execute the SQL query or stored procedure. See <see cref="Parameters"/> for more details.</param>
+        public SqlAttribute(string commandText, string connectionStringSetting, string parameters) : this(commandText, connectionStringSetting, CommandType.Text, parameters) { }
```
Charles-Gagnon, line 47: Do we need to do this for SqlInputAttribute as well?

lucyzhang929: The outofproc samples without commandType and with parameters are working as expected. Ex. https://github.com/Azure/azure-functions-sql-extension/blob/60d0bae609cdf59a31bb82a0a3d415eeece23d07/samples/samples-outofproc/InputBindingSamples/GetProducts.cs#L20-L22

MaddyDev (approved):

src/SqlAttribute.cs
Charles-Gagnon: Oh right, yeah that makes sense given that it's still C# and not used for other languages!

Charles-Gagnon (approved):

PR #792, opened 2023-04-06
add copy task to ensure sql package is available in local-packages folder for tests

### LINKED ISSUE #687: Pipeline builds sometimes fail
Been seeing this occasionally in adhoc runs - it'll fail the initial build with this error 

image

If you look closely it seems like the package task isn't running. Here's the output from a successful build : 

image

MaddyDev:
Added a copy task to explicitly copy the sql package to local-packages.

Fixes #687

### DIFF
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -102,6 +102,14 @@ steps:
     targetFolder: $(azureFunctionsExtensionBundlePath)/bin
     overWrite: true
 
+- task: CopyFiles@2
+  displayName: 'Copy local Sql package to local-packages folder'
+  inputs:
+    sourceFolder: $(Build.SourcesDirectory)/src/bin/${{ parameters.configuration }}
+    contents: *
+    targetFolder: $(Build.SourcesDirectory)/local-packages
+    overWrite: true
+
 - script: |
     npm install
     npm run lint
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -102,6 +102,7 @@ steps:
     targetFolder: $(azureFunctionsExtensionBundlePath)/bin
     overWrite: true
 
+  # Copy the Sql nupkg to ensure it's available for tests.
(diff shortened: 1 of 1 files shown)


builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -102,6 +102,15 @@ steps:
     targetFolder: $(azureFunctionsExtensionBundlePath)/bin
     overWrite: true
 
+  # Copy the Sql nupkg to ensure it's available for tests.
+- task: CopyFiles@2
```
Charles-Gagnon, line 106: But why does the build not do the copy? We have the task there so duplicating it here seems weird.

MaddyDev: I've looked up why occasionally the package copy task isn't working as intended, but couldn't find anything relatable. I'll keep investigating meanwhile since this has happened a couple times, wanted to have this redundancy.

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -102,6 +102,15 @@ steps:
...
+  displayName: 'Copy local Sql package to local-packages folder'
+  inputs:
+    sourceFolder: $(Build.SourcesDirectory)/src/bin/${{ parameters.configuration }}
+    contents: '*'
```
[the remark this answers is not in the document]
MaddyDev: the folder has just nupkg and snupkg, * only copies the files and not all the subfolders and it's contents.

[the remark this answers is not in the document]
Charles-Gagnon: It would still be good to specify only *.nupkg anyways (we don't need the snupkg) - that way if something else ends up there we don't unnecessarily copy stuff and potentially cause issues.

builds/azure-pipelines/template-steps-build-test.yml
Charles-Gagnon: Sounds good, could you add that to the comment so future us remembers why we have this duplicate copy?

Charles-Gagnon (approved):

lucyzhang929 (approved):

PR #795, opened 2023-04-10
Enable trace level logging for tests

Charles-Gagnon:
See if this provides any more useful information with the random trigger test failures we're seeing.

### DIFF
--- a/test/Integration/test-java/host.json
+++ b/test/Integration/test-java/host.json
@@ -8,7 +8,7 @@
       }
     },
     "logLevel": {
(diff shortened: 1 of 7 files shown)


test/Integration/test-java/host.json
```diff
@@ -8,7 +8,7 @@
       }
     },
     "logLevel": {
-      "default": "Information"
+      "default": "Trace"
```
lucyzhang929, line 11: There is also the host.json in samples-java which is currently set to "Information".

PR #799, opened 2023-04-10
Fix up performance tests & enable schedule

Charles-Gagnon:
Going to start working on setting up the performance test runs to actually start comparing to a baseline - but the first step is to fix the project so it runs and then enable the pipeline to run on a schedule so we can catch any issues that crop up.

### DIFF
--- a/builds/azure-pipelines/performance.yml
+++ b/builds/azure-pipelines/performance.yml
@@ -2,7 +2,13 @@ trigger: none
 
 pr: none
 
-schedules: [ ]
+schedules:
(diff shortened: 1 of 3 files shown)


lucyzhang929 (approved):

Charles-Gagnon:
@lucyzhang929 There's a failing test for Python with the trigger binding polling interval. You had done something in the previous PR to fix that, right? Is there still an issue you're aware of?

I'm going to merge this in since it isn't related to the changes here, but let's try to figure out if this is the same missing log message issue or a new one.

PR #801, opened 2023-04-12
enable trigger with exception test

MaddyDev:
Enabled the TriggerWithException test for OOP C# tests and Java.

### DIFF
--- a/test-outofproc/TriggerWithException.cs
+++ b/test-outofproc/TriggerWithException.cs
@@ -14,6 +14,7 @@ namespace DotnetIsolatedTests
     {
         public const string ExceptionMessage = "TriggerWithException test exception";
         private static bool threwException = false;
+        private static readonly Action<ILogger, string, Exception> _loggerMessage = LoggerMessage.Define<string>(LogLevel.Information, eventId: new EventId(0, "INFO"), formatString: "{Message}");
 
         /// <summary>
         /// Used in verification that exceptions thrown by functions cause the trigger to retry calling the function
@@ -23,14 +24,15 @@ namespace DotnetIsolatedTests
         public static void Run(
             [SqlTrigger("[dbo].[Products]", "SqlConnectionString")]
             IReadOnlyList<SqlChange<Product>> changes,
-            ILogger logger)
+            FunctionContext context)
         {
             if (!threwException)
             {
                 threwException = true;
                 throw new InvalidOperationException(ExceptionMessage);
             }
-            logger.LogInformation("SQL Changes: " + Utils.JsonSerializeObject(changes));
+            _loggerMessage(context.GetLogger("TriggerWithException"), "SQL Changes: " + Utils.JsonSerializeObject(changes), null);
+
         }
     }
 }
--- a/test/Integration/SqlTriggerBindingIntegrationTests.cs
+++ b/test/Integration/SqlTriggerBindingIntegrationTests.cs
@@ -588,11 +588,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
         /// <summary>
         /// Tests that when a user function throws an exception we'll retry executing that function once the lease timeout expires
         /// </summary>
-        [Fact]
-        public async Task FunctionExceptionsCauseRetry()
+        [Theory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.JavaScript, SupportedLanguages.Python, SupportedLanguages.PowerShell, SupportedLanguages.CSharpscript)] //Static state threwException is only valid for C# and Java.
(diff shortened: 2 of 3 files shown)


test-outofproc/TriggerWithException.cs
```diff
@@ -23,14 +24,15 @@ public static class TriggerWithException
...
-            ILogger logger)
+            FunctionContext context)
         {
             if (!threwException)
             {
                 threwException = true;
                 throw new InvalidOperationException(ExceptionMessage);
             }
-            logger.LogInformation("SQL Changes: " + Utils.JsonSerializeObject(changes));
+            _loggerMessage(context.GetLogger("TriggerWithException"), "SQL Changes: " + Utils.JsonSerializeObject(changes), null);
```
Charles-Gagnon, line 34: Is there a reason you're changing this?

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -588,11 +588,13 @@ public void UnsupportedDatabaseThrows(SupportedLanguages lang)
         /// <summary>
         /// Tests that when a user function throws an exception we'll retry executing that function once the lease timeout expires
         /// </summary>
-        [Fact]
-        public async Task FunctionExceptionsCauseRetry()
+        [Theory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.JavaScript, SupportedLanguages.Python, SupportedLanguages.PowerShell, SupportedLanguages.CSharpscript)] //Static state threwException is only valid for C# and Java.
```
Charles-Gagnon, line 593: What does this mean?

test-outofproc/TriggerWithException.cs
MaddyDev: I keep seeing the exception "logger cannot be null" when I run it the other way:
image

test/Integration/SqlTriggerBindingIntegrationTests.cs
MaddyDev: Lucy told me that she intentionally skipped this since the static variables isn't valid for those languages, now that I think about it I can add a variable and add the test.

lucyzhang929: The FunctionExceptionsCauseRetry test relies on the static [threwException](https://github.com/Azure/azure-functions-sql-extension/blob/33ee38c3f1a578093cf7612649ead737b20f8120/test/Integration/test-csharp/TriggerWithException.cs#L14) value in the function but I don't think there is a way to do this in JS, Python, PS.

Charles-Gagnon: Hm, ok this is fine for now then. Hopefully can figure out a way to do this with a bit more digging in the future but not a huge concern. @MaddyDev could you file an issue to track looking into this?

Charles-Gagnon, line 593:
```suggestion
        [UnsupportedLanguages(SupportedLanguages.JavaScript, SupportedLanguages.Python, SupportedLanguages.PowerShell, SupportedLanguages.CSharpscript)] // Keeping static state for threwException across calls is only valid for C# and Java.
```

test-outofproc/TriggerWithException.cs
Charles-Gagnon: Ok, so it looks like using the execution context is just how you do it in OOP (for some reason). 

https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#logging

Could you simplify this though? It doesn't seem necessary to have this logger delegate. (if you're getting a compiler warning that's probably the one I've disabled for the rest of our test code as well - you should do the same here in the editorconfig)

MaddyDev: shouldn't we address the complier warning so users know how to overcome it instead of ignoring it?

Charles-Gagnon: 1. This is specifically test code, so it's not something users should really be using as a guide
2. It's normally a warning, we just by default treat everything as errors. It's not breaking to have it with the simplified version and the difference only really matters for things that are doing a high rate of logging which really care about performance (if this is the same warning that I'm thinking of)

MaddyDev: Will clean up the tests in a separate PR, thanks.

chlafreniere (approved):

lucyzhang929 (approved):

Charles-Gagnon (approved):

MaddyDev:
Looking into the timeout even after all the tests pass.

MaddyDev:
Merging this in, created #805 for investigating Java failure and enabling it for other languages.

PR #806, opened 2023-04-14
add csx trigger samples

MaddyDev:
Trigger samples for CSharp Script and enabled the tests.


test/Integration/test-csx/Common/Product.csx
```diff
@@ -234,4 +242,66 @@ public class ProductUtilities
         }
         return products;
     }
+}
+
+public static class Utils
```
Charles-Gagnon, line 247: Would suggest making a separate file for this, we should generally be keeping 1 class/file. 

But also - can you not just use the one from the Utils class directly with the #load directive?

test/Integration/test-csx/MultiFunctionTrigger2/run.csx
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Net;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
+using Microsoft.Azure.WebJobs.Extensions.Sql;
+
+public static void Run(IReadOnlyList<SqlChange<Product>> changes, ILogger log)
+{
+    log.LogInformation("Trigger2 Changes: " + JsonConvert.SerializeObject(changes));
```
Charles-Gagnon, line 13: Shouldn't this be using the Utils.SerializeObject method?

test/Integration/test-csx/Common/Product.csx
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Linq;
+using System.IO;
```
chlafreniere, line 6: Can we remove these `using` statements now?

test/Integration/test-csx/Common/utils.csx
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
chlafreniere, line 2: Is this file materially different from utils.cs elsewhere in test code? Can it be reused?

test/Integration/test-csx/Common/utils.csx
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// <summary>
+    /// Serializes the specified object into a JSON string.
+    /// </summary>
+    /// <param name="obj">The object to serialize</param>
+    /// <param name="settings">The specific settings to use, uses a simple set of default settings if not specified</param>
+    /// <returns>The serialized JSON string</returns>
+    /// <remarks>This will NOT use any global settings to avoid picking up changes that may have been made by other code running in the host (such as user functions)</remarks>
+    public static string JsonSerializeObject(object obj, JsonSerializerSettings settings = null)
```
Charles-Gagnon, line 37: Is there no way to import the existing Utils.cs file and use that? Avoids us having to remember to update both if we make changes there later on.

test/Integration/test-csx/Common/utils.csx
MaddyDev: that's testutils and the functions in use aren't there, so added a new one here - but I finally figured out on how to use the utils from extension itself.

Charles-Gagnon (approved):

PR #808, opened 2023-04-14
Cleanup trigger logging

Charles-Gagnon:
Part 2 of logging cleanup, same stuff as https://github.com/Azure/azure-functions-sql-extension/pull/802

With this we now have : 

13 Debug log messages - Extra information that could be useful for debugging, mostly startup/config info
2 Information messages - Telemetry opt-out and connection re-established messages
6 Warning messages - Non-critical errors
28 Error messages - Critical errors that indicate something is wrong

I plan on doing another pass after this is in to verify that the levels we have stuff at makes sense, but this should be pretty much complete at this point. Normal execution should only have a dozen or so messages per instance on startup with the occasional status update message.

### DIFF
--- a/samples/samples-csharp/host.json
+++ b/samples/samples-csharp/host.json
@@ -8,7 +8,7 @@
             }
         },
         "logLevel": {
-          "default": "Information"
+          "default": "Debug"
         }
     }
 }
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -197,9 +197,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                 using (var connection = new SqlConnection(this._connectionString))
                 {
-                    this._logger.LogDebug("BEGIN OpenChangeConsumptionConnection");
                     await connection.OpenAsync(token);
-                    this._logger.LogDebug("END OpenChangeConsumptionConnection");
 
                     bool forceReconnect = false;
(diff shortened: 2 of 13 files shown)


Charles-Gagnon:
Ah right, yeah I'll have to fix up the tests. Putting it in draft while I work on those in case I have to refactor anything...

samples/samples-csharp/host.json
```diff
@@ -8,7 +8,7 @@
             }
         },
         "logLevel": {
-          "default": "Information"
+          "default": "Debug"
```
Charles-Gagnon, line 11: Now that we use debug logging for more stuff I had to change this for tests. We originally had it as Information so customers didn't copy the debug settings, but in this case I think it's fine to have that vs. adding some functionality to switch the log level at runtime (which I had looked into earlier and didn't see an obvious way to do)

Charles-Gagnon:
Ok, test fixes are in. One failing test is from the logging issue we're currently working on so unrelated to these changes.

lucyzhang929 (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -717,9 +685,7 @@ private long RecomputeLastSyncVersion()
             // the only version number in the set.
             // Also this LastSyncVersion is actually updated in the GlobalState table only after verifying that the changes with
```
VasuBhog, line 686: Remove / Change comment since you removed the lastSyncVersion var.

VasuBhog (approved):

PR #813, opened 2023-04-18
Refactor integration tests

lucyzhang929:
Previously we were starting and stopping the function host for each test. Now in IntegrationTestFixture we will
1. create a database to be shared by all tests (the tables, views, and stored procedures are still recreated for each test)
2. start the function hosts (one for each language folder). Some functions are skipped because they will affect all tests (ex. any non-http trigger functions) or they don't apply to all languages (ex. AddProductsCollector).


test/Integration/IntegrationTestBase.cs
```diff
@@ -427,16 +311,18 @@ protected void DisposeFunctionHosts()
             this.FunctionHostList.Clear();
         }
 
-        protected async Task<HttpResponseMessage> SendInputRequest(string functionName, string query = "")
+        protected async Task<HttpResponseMessage> SendInputRequest(string functionName, string query = "", int? port = null)
```
MaddyDev, line 314: port shouldn't be nullable here?

MaddyDev, line 314: port shouldn't be nullable here, you can directly default it to Port here

test/Integration/IntegrationTestBase.cs
```diff
@@ -60,91 +54,36 @@ public class IntegrationTestBase : IDisposable
         protected ITestOutputHelper TestOutput { get; private set; }
 
         /// <summary>
-        /// The port the Functions Host is running on. Default is 7071.
+        /// The port the Functions Host is running on. Start at 7081 since the function hosts started in
+        /// IntegrationTestFixture are already running on ports 7071 - 7080.
         /// </summary>
-        protected int Port { get; private set; } = 7071;
+        protected int Port { get; private set; } = 7081;
```
Charles-Gagnon, line 60: I'd suggest defining a const for this in TestUtils next to the port helper method so that if we end up adding more ports later on we remember to update this.

Charles-Gagnon (approved):

MaddyDev (approved):

test/Integration/IntegrationTestFixture.cs
```diff
@@ -41,8 +80,184 @@ protected void StartAzurite()
...
+
+            Console.WriteLine($"Waiting for Azure Function host in {workingDirectory} to start...");
+
+            const int FunctionHostStartupTimeoutInSeconds = 60;
+            bool isCompleted = taskCompletionSource.Task.Wait(TimeSpan.FromSeconds(FunctionHostStartupTimeoutInSeconds));
+            Assert.True(isCompleted, $"Functions host in {workingDirectory} did not start within specified time.");
+
+            // Give additional time to Functions host to setup routes for the HTTP triggers so that the HTTP requests
+            // made from the test methods do not get refused.
+            const int BufferTimeInSeconds = 5;
```
chlafreniere, line 238: nit: this would ideally live in a TestConstants.cs file or something of that nature

chlafreniere (approved):

PR #815, opened 2023-04-18
Add extra simple log statement to trigger tests

Charles-Gagnon:
See if size of the log message is potentially an issue here


chlafreniere (approved):

lucyzhang929:
We've seen failures in OutOfProc tests too - can we add the log there as well?

MaddyDev (approved):

lucyzhang929 (approved):

VasuBhog (approved):

PR #821, opened 2023-04-21
[Trigger] Fix java version in setup guide

Charles-Gagnon:
Leaving as draft until I can verify it (or @lucyzhang929 gets to it first)


lucyzhang929 (approved):
Verified that this works as expected

PR #812, opened 2023-04-18
Clean csx samples

MaddyDev:
Simplified the samples structure to follow javascript. 
Cleaned the samples to remove the [Sql()] attribute from the function code.

PR #823, opened 2023-04-21
Enable logging for all tests and rename samples-csharpscript

Charles-Gagnon:
The CSX tests are failing because they don't have the log level set to the correct level (that message is sent at the Debug level). So fixing that and added the log level to the other tests/samples that don't have it so all our tests/samples have the same settings.

I'm also renaming samples-csharpscript to samples-csx to match the test-csx folder we already have.


MaddyDev:
Is there a reason why you have created against release/trigger branch and not main? These changes are not exactly related to trigger stuff. FYI I also have a PR out #812 to address some csx clean up against main - merging can get messy.

Charles-Gagnon:
Fair - the failing tests were the trigger ones (I don't believe we have any Input/Output tests that rely on messages like the trigger tests) so that's why I did them here, but you're right about the merging so let me open up an initial PR in main and then I can port it over + make the rest of the updates as a follow up.

PR #825, opened 2023-04-21
[Trigger] Rename csharpscript and align host.json

Charles-Gagnon:
Port over changes from main and move over trigger stuff. This should fix the issue with the trigger tests failing for CSX as well due to the log level not being at a high enough level.


Charles-Gagnon:
Merging this in since the one in main was already approved and I want to kick off some runs with my other changes over the weekend...

PR #830, opened 2023-04-25
Add node v4 programming model sample

lucyzhang929:
Using the [generic input and output bindings](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=windows-setting-the-node-version&pivots=nodejs-model-v4#generic-inputs-and-outputs) while we look into adding first class support for Sql bindings.


samples/samples-js-v4/host.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version": "2.0",
```
Charles-Gagnon, line 2: I don't see a local.settings.json, you need to `git add -f` that directly since it's ignored by default.

samples/samples-js-v4/host.json
```diff
@@ -0,0 +1,15 @@
...
+  "logging": {
```
Charles-Gagnon, line 3: Copy the host.json from the other samples so they're all aligned (should have default log level of trace)

Charles-Gagnon (approved):

MaddyDev (approved):
Would be great to link these in our docs so they are visible to whoever is looking for them.

PR #836, opened 2023-05-01
Add retry for Trigger tests

MaddyDev:
I've tried to add our own implementation of the retry following the instructions by [xunit samples](https://github.com/xunit/samples.xunit/tree/main/RetryFactExample) but couldn't make it work as expected (although I had to make some adjustments since we cannot really use the Fact with our custom SqlInlineData). But after looking for more examples online, I came across the [XRetry](https://github.com/JoshKeegan/xRetry) which works as expected.
I could look into incorporating what they have but would take a bit more time, let me get started with adding that package dependency meanwhile I look into what parts we specifically need for our specific case.

Am adding the retry for only trigger tests for now.


Directory.Packages.props
```diff
@@ -15,6 +15,7 @@
     <PackageVersion Include="Moq" Version="4.18.2" />
     <PackageVersion Include="xunit" Version="2.4.2" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
+     <PackageVersion Include="xRetry" Version="1.9.0" />
```
Charles-Gagnon, line 18: nit: Fix indentation here

Charles-Gagnon (commented):
Did you investigate this part?

```
By default, you won't see whether your tests are being retried as we make this information available via the xunit diagnostic logs but test runners will hide these detailed logs by default.
To enable them you must configure your xUnit test project to have diagnosticMessages set to true in the xunit.runner.json.
```

For now it'd be good to have these in the logs, so if that isn't happening currently then it'd be good to see how easy this is to enable and how much more verbose it makes the logs.

Charles-Gagnon:
We're still occasionally getting the `Error NU1301: The local source 'D:\a\1\s\local-packages' doesn't exist.` it seems. Could you make an issue for that and follow up with investigation there as well?

Charles-Gagnon (approved):

chlafreniere (approved):

lucyzhang929 (approved):

PR #838, opened 2023-05-01
FIx perf pipeline

lucyzhang929:


chlafreniere:
Why was starting the function host an issue for perf tests?

lucyzhang929:
The fixture starts the function hosts for all languages but the perf tests currently only copy over the CSharp functions so it was failing when trying to start the function hosts for other languages.

https://github.com/Azure/azure-functions-sql-extension/blob/2d790597f84bf48f3878d702afcc1f0e91d8ccc6/performance/Microsoft.Azure.WebJobs.Extensions.Sql.Performance.csproj#L18-L24

MaddyDev (approved):

Charles-Gagnon:
Would a better fix then be to have the fixture still start up the hosts, but be able to specify which ones to start? Having all the startup logic in the same place seems beneficial.

lucyzhang929:
Converting this to draft since tests are failing witht he following error:
`Collection fixture type 'Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration.IntegrationTestFixture' had one or more unresolved constructor arguments: Boolean startFunctionHosts`
Based on [this comment](https://github.com/xunit/xunit/discussions/2459), collection fixtures cannot have arguments. Will look into a different way.

lucyzhang929:
Decided not to use the IntegrationTestFixture in the perf tests since we don't need all the function hosts to start. I moved the start azurite and create database logic from IntegrationTestFixture to TestUtils so that it can be reused in the perf tests.

performance/SqlBindingBenchmarks.cs
```diff
@@ -48,6 +50,9 @@ public static void Main(string[] args)
             {
                 BenchmarkRunner.Run<SqlTriggerBindingPerformance_ChangeRate>();
             }
+
+            TestUtils.StopAzurite(azuriteHost);
```
Charles-Gagnon, line 54: These should go in a finally block

Charles-Gagnon (approved):

PR #839, opened 2023-05-02
Fix the race condition by adding the extension reference to build first.

### LINKED ISSUE #837: The local source 'D:\a\1\s\local-packages' doesn't exist.
Build step fails with the below error:
image

MaddyDev:
Fixes #837.
I observed that the build step fails even before the copy to local-packages step in the pipeline and the `Set up local-packages Nuget source` step prior to the build shows the below result:
C:\hostedtoolcache\windows\dotnet\dotnet.exe nuget add source -n afsqlext.local **D:\a\1\s/local-packages**
Package source with Name: afsqlext.local added successfully.

If the out of proc samples or tests are built first and the sql extension is still not available in local-packages, to avoid this added the sql extension to the project dependencies, so we make sure to have the local-packages available. 

Fixing that in this PR and going to monitor if that stops the "Error NU1301: The local source 'D:\a\1\s\local-packages' doesn't exist."

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -43,11 +43,11 @@
     <ItemGroup>
       <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
     </ItemGroup>
-    <Copy SourceFiles="@(_Packages)" DestinationFolder="../local-packages" />
+    <Copy SourceFiles="@(_Packages)" DestinationFolder="..\local-packages" />
     <Message Text="Copied sql .nupkg to local-packages" Importance="high" />
   </Target>
   <Target Name="RemoveNugetPackageCache" BeforeTargets="Build">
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -43,11 +43,11 @@
     <ItemGroup>
       <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
     </ItemGroup>
-    <Copy SourceFiles="@(_Packages)" DestinationFolder="../local-packages" />
+    <Copy SourceFiles="@(_Packages)" DestinationFolder="..\local-packages" />
```
Charles-Gagnon, line 46: If this was a problem wouldn't it be completely broken all the time? Path issues are generally not intermittent things.

Charles-Gagnon:
If I remember correctly from my investigation, the issue seems to be that the nuget package is never actually built before this project tries to access it. So I'd start your investigation with ensuring that the build order is consistent - maybe there's a race condition happening here with when the "temporary" project is being built.

Charles-Gagnon (approved):
Please update the PR title & description 😄

lucyzhang929 (approved):

PR #853, opened 2023-05-05
pack and copy the sql nupkg to local-packages after build

MaddyDev:
Make sure to copy the sql package to local-packages before build
Addresses #837

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -39,7 +39,7 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-  <Target Name="CopyNupkg" AfterTargets="Pack">
+  <Target Name="CopyNupkg" BeforeTargets="Build">
     <ItemGroup>
       <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
     </ItemGroup>
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -39,7 +39,7 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-  <Target Name="CopyNupkg" BeforeTargets="Build">
+  <Target Name="CopyNupkg" AfterTargets="Build">
     <ItemGroup>
       <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
     </ItemGroup>
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -40,6 +40,7 @@
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
   <Target Name="CopyNupkg" AfterTargets="Build">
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -39,7 +39,7 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-  <Target Name="CopyNupkg" AfterTargets="Pack">
+  <Target Name="CopyNupkg" AfterTargets="Build">
```
Charles-Gagnon, line 42: I assume this won't run if you do `dotnet pack` then, right?

MaddyDev: By default, dotnet pack builds the project first. (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-pack)

Charles-Gagnon: But after it builds the package doesn't exist yet...it needs to actually complete the pack target. So this doesn't seem like the correct ordering here - if there is a nupkg it will contain old stuff since the newly built stuff hasn't been built yet.

MaddyDev: We have <GeneratePackageOnBuild>true</GeneratePackageOnBuild>, I assumed we'll have the package ready.

Charles-Gagnon: Hmm, did you verify that? Does it actually generate it as part of the build task or just kick off the pack task after the build task is done?

Charles-Gagnon: IIRC if you run with verbose logging it lists out all the targets it ran at the end, which should help understand what exactly happens.

MaddyDev: Will add the verbose logging and see the order, thanks!

MaddyDev: > Hmm, did you verify that? Does it actually generate it as part of the build task or just kick off the pack task after the build task is done?

After checking the verbose logs, GeneratePackageOnBuild will trigger pack command after build. Now am not sure why sometimes that step is missing from the logs on the pipeline (on the failed runs). I tried searching online but didn't come across anything that's helpful. Am trying to force the pack command to run explicitly after build and see if we have a consistent copy package.

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -39,7 +39,8 @@
...
+    <Exec Command="dotnet pack &quot;$(MSBuildProjectDirectory)\$(PackageId).csproj&quot; --no-build --include-symbols"/>
```
Charles-Gagnon, line 43: Can you remove GeneratePackageOnBuild now with this then?

Charles-Gagnon, line 43: Please add a comment about why this is necessary

src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
```diff
@@ -39,7 +39,8 @@
     <None Include="..\lib\Microsoft.SqlServer.TransactSql.ScriptDom.dll" Pack="true" PackagePath="lib/$(TargetFramework)" />
     <None Include="..\README.md" Pack="true" PackagePath="" />
   </ItemGroup>
-  <Target Name="CopyNupkg" AfterTargets="Pack">
+  <Target Name="CopyNupkg" AfterTargets="Build">
```
Charles-Gagnon, line 42:
```suggestion
  <Target Name="BuildAndCopyNupkg" AfterTargets="Build">
```

Charles-Gagnon (approved):

Charles-Gagnon (commented):
I'm not 100% confident this is going to actually solve the issue - are we actually guaranteed to run this target and have it complete before running Build on the worker extension project? There's nothing that's saying that the build for that project needs to wait on this target...

Guess we can try it and see - but it might be worth it to try looking around for how to insert a new target as part of the overall build target instead of after it.

PR #861, opened 2023-05-24
add timestamp to global state table for lease table cleanup

MaddyDev:
Added LastAccessTime Datetime column to the GlobalState table. For accommodating existing triggers, added the column with default getutcdate() value.

Added the clean up scripts to the BindingsOveriew doc, I could move them over to a more appropriate place if folks have better suggestions.

### DIFF
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -22,6 +22,7 @@
       - [Startup retries](#startup-retries)
       - [Broken connection retries](#broken-connection-retries)
       - [Function exception retries](#function-exception-retries)
+      - [Lease Tables clean up](#lease-tables-clean-up)
 
 ## Input Binding
 
@@ -147,7 +148,7 @@ If an exception occurs in the user function when processing changes then the bat
 
 If the function execution fails 5 times in a row for a given row then that row is completely ignored for all future changes. Because the rows in a batch are not deterministic, rows in a failed batch may end up in different batches in subsequent invocations. This means that not all rows in the failed batch will necessarily be ignored. If other rows in the batch were the ones causing the exception, the "good" rows may end up in a different batch that doesn't fail in future invocations.
 
-You can run this query to see what rows have failed 5 times and are currently ignored, see [Leases table](#az_funcleases_) documentation for how to get the correct Leases table to query for your function.
+You can run this query to see what rows have failed 5 times and are currently ignored, see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for how to get the correct Leases table to query for your function.
 
 ```sql
 SELECT * FROM [az_func].[Leases_<FunctionId>_<TableId>] WHERE _az_func_AttemptCount = 5
@@ -167,4 +168,81 @@ e.g.
 
 ```sql
 UPDATE [Products].[az_func].[Leases_<FunctionId>_<TableId>] SET _az_func_AttemptCount = 0 WHERE ProductId = 123
+```
+
+#### Lease Tables clean up
+
+Before clean up, please see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for how to get the correct Leases table.
+
+- Clean up all the tables not accessed more than N number of days, Please initialize N according to your use case:
+
+```sql
+DECLARE @table_name NVARCHAR(MAX);
+DECLARE @UserFunctionId char(16);
+DECLARE @UserTableId int;
+DECLARE @N int;
+DECLARE leasetable_cursor CURSOR FOR
+
+SELECT 'az_func.Leases_'+UserFunctionId+'_'+convert(varchar(100),UserTableID) as TABLE_NAME, UserFunctionID, UserTableID
+FROM az_func.GlobalState
+WHERE DATEDIFF(day, LastAccessTime, GETDATE()) > N
+
+OPEN leasetable_cursor;
+
+FETCH NEXT FROM leasetable_cursor INTO @table_name, @UserFunctionId, @UserTableId;
+
+WHILE @@FETCH_STATUS = 0
+BEGIN
+    EXEC ('DROP TABLE IF EXISTS ' + @table_name);
+    DELETE FROM az_func.GlobalState WHERE UserFunctionID = @UserFunctionId and UserTableID = @UserTableId
+    FETCH NEXT FROM leasetable_cursor INTO @table_name, @UserFunctionId, @UserTableId;
+END;
+
+CLOSE leasetable_cursor;
+
+DEALLOCATE leasetable_cursor;
+```
+
+- Clean up a specific lease table concerning a specific function:
+
+To find the name of the leases table associated with your function, look in the log output for a line such as this which is emitted when the trigger is started.
+
+`SQL trigger Leases table: [az_func].[Leases_84d975fca0f7441a_901578250]`
+
+This log message is at the `Information` level, so make sure your log level is set correctly.
+
+```sql
+DECLARE @table_name NVARCHAR(MAX); --@table_name would be [az_func].[Leases_84d975fca0f7441a_901578250]
+DECLARE @UserFunctionId char(16); -- @UserFunctionId would be 84d975fca0f7441a
+DECLARE @UserTableId int; -- @UserTableId would be 901578250
+
+DROP TABLE IF EXISTS @table_name
+DELETE FROM az_func.GlobalState WHERE UserFunctionID = @UserFunctionId and UserTableID = @UserTableId
+```
+
+- Clear all trigger related data for a reset:
+
+```sql
+DECLARE @table_name NVARCHAR(MAX);
+DECLARE @UserFunctionId char(16);
+DECLARE @UserTableId int;
+DECLARE leasetable_cursor CURSOR FOR
+
+SELECT 'az_func.Leases_'+UserFunctionId+'_'+convert(varchar(100),UserTableID) as TABLE_NAME, UserFunctionID, UserTableID
+FROM az_func.GlobalState
+
+OPEN leasetable_cursor;
+
+FETCH NEXT FROM leasetable_cursor INTO @table_name, @UserFunctionId, @UserTableId;
+
+WHILE @@FETCH_STATUS = 0
+BEGIN
+    EXEC ('DROP TABLE IF EXISTS ' + @table_name);
+    DELETE FROM az_func.GlobalState WHERE UserFunctionID = @UserFunctionId and UserTableID = @UserTableId
+    FETCH NEXT FROM leasetable_cursor INTO @table_name, @UserFunctionId, @UserTableId;
+END;
+
+CLOSE leasetable_cursor;
+
+DEALLOCATE leasetable_cursor;
 ```
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -185,7 +185,8 @@ The Azure SQL Trigger does not currently handle automatically cleaning up any le
 ```sql
 -- Deletes all the lease tables that haven't been accessed in @CleanupAgeDays days (set below)
 -- and removes them from the GlobalState table.
-USE <Insert DATABASE name here>;
+USE [<Insert DATABASE name here>]
+GO
 DECLARE @TableName NVARCHAR(MAX);
 DECLARE @UserFunctionId char(16);
 DECLARE @UserTableId int;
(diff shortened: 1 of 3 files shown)


docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
 ```sql
 UPDATE [Products].[az_func].[Leases_<FunctionId>_<TableId>] SET _az_func_AttemptCount = 0 WHERE ProductId = 123
 ```
+
+#### Lease Tables clean up
+
+Before clean up, please see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for how to get the correct Leases table.
+
+- Clean up all the tables not accessed more than N number of days, Please initialize N according to your use case:
```
Charles-Gagnon, line 177:
```suggestion
- Deletes all the lease tables that haven't been accessed in N days:
```

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+
+```sql
+DECLARE @table_name NVARCHAR(MAX);
```
Charles-Gagnon, line 180: Add a `USE [<DbName>]` at the top with instructions to set it to the database to be cleaned up

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+DECLARE @UserFunctionId char(16);
+DECLARE @UserTableId int;
+DECLARE @N int;
```
Charles-Gagnon, line 183: I would also suggest naming this `CleanupAgeDays` or something similar to make it clearer

```suggestion
DECLARE @N int = <Insert desired cleanup age in days here>`
```

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+
+#### Lease Tables clean up
+
+Before clean up, please see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for how to get the correct Leases table.
+
+- Clean up all the tables not accessed more than N number of days, Please initialize N according to your use case:
+
+```sql
+DECLARE @table_name NVARCHAR(MAX);
+DECLARE @UserFunctionId char(16);
```
Charles-Gagnon, line 181: Please also add a comment explaining what the user should set this value to and how to get it so the script itself has context about what it's doing.

```suggestion
DECLARE @UserFunctionId char(16) = <Insert function ID here>`
```

docs/BindingsOverview.md
Charles-Gagnon, line 180: Would suggest putting a comment at the top explaining what the script is doing as well so people can simply copy and paste it and not have to add their own context in (for example if they make it into a job)

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+DECLARE @UserTableId int;
+DECLARE @N int;
+DECLARE leasetable_cursor CURSOR FOR
```
Charles-Gagnon, line 184: nit: LeaseTable_Cursor (keep consistent capitalization in the script)

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+- Clean up a specific lease table concerning a specific function:
+
+To find the name of the leases table associated with your function, look in the log output for a line such as this which is emitted when the trigger is started.
+
+`SQL trigger Leases table: [az_func].[Leases_84d975fca0f7441a_901578250]`
+
+This log message is at the `Information` level, so make sure your log level is set correctly.
+
+```sql
+DECLARE @table_name NVARCHAR(MAX); --@table_name would be [az_func].[Leases_84d975fca0f7441a_901578250]
```
Charles-Gagnon, line 215: Same here about doing `DECLARE @... = <Insert value here>` type stuff so the user can't run it until they fill in those values. 

You may also want to change the comments you have here to explain what the values are and then say something like `, e.g. '[az_func].[Leases_84d975fca0f7441a_901578250]'` to indicate that these are sample values of what they should look like

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
...
+DECLARE @UserFunctionId char(16); -- @UserFunctionId would be 84d975fca0f7441a
+DECLARE @UserTableId int; -- @UserTableId would be 901578250
+
+DROP TABLE IF EXISTS @table_name
+DELETE FROM az_func.GlobalState WHERE UserFunctionID = @UserFunctionId and UserTableID = @UserTableId
+```
+
+- Clear all trigger related data for a reset:
```
Charles-Gagnon, line 223: Maybe just have the first script be able to take in -1 for the days value to clean up all tables instead? Help keep things simpler

docs/BindingsOverview.md
Charles-Gagnon, line 215: nit: `@TableName` to keep casing consistent

docs/BindingsOverview.md
```diff
@@ -168,3 +169,80 @@ e.g.
 ```sql
 UPDATE [Products].[az_func].[Leases_<FunctionId>_<TableId>] SET _az_func_AttemptCount = 0 WHERE ProductId = 123
 ```
+
+#### Lease Tables clean up
+
+Before clean up, please see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for how to get the correct Leases table.
```
Charles-Gagnon, line 175: Add comment here explaining why cleanup is necessary and when the user would want to do it.

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+    DELETE FROM az_func.GlobalState WHERE UserFunctionID = @UserFunctionId and UserTableID = @UserTableId
+    FETCH NEXT FROM LeaseTable_Cursor INTO @TableName, @UserFunctionId, @UserTableId;
+END;
+
+CLOSE LeaseTable_Cursor;
+
+DEALLOCATE LeaseTable_Cursor;
+```
+
+- Clean up a specific lease table concerning a specific function:
```
Charles-Gagnon, line 215:
```suggestion
- Clean up a specific lease table:
```

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+
+To find the name of the leases table associated with your function, look in the log output for a line such as this which is emitted when the trigger is started.
```
Charles-Gagnon, line 217:
```suggestion
To find the name of the lease table associated with your function, look in the log output for a line such as this which is emitted when the trigger is started.
```

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+
+`SQL trigger Leases table: [az_func].[Leases_84d975fca0f7441a_901578250]`
+
+This log message is at the `Information` level, so make sure your log level is set correctly.
+
+```sql
+-- Deletes the specified lease table and removes it from GlobalState table.
+USE <Insert DATABASE name here>;
+DECLARE @TableName NVARCHAR(MAX) = <Insert lease table name here>; -- e.g. '[az_func].[Leases_84d975fca0f7441a_901578250]
+DECLARE @UserFunctionId char(16) = <Insert function ID here>; -- e.g. '84d975fca0f7441a'
```
Charles-Gagnon, line 227: Add comments explaining how they get these values (from the leases table name)

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+
+#### Lease Tables clean up
+
+Before clean up, please see [Leases table](./TriggerBinding.md#az_funcleases_) documentation for understanding how they are created and used.
+
+Why clean up?
+1. You renamed your function/class/method name, which causes a new lease table to be created and the old one to be obsolete.
+2. You created a trigger function that you no longer need and wish to clean up its associated data.
+3. You want to reset your environment.
+The Azure SQL Trigger does not currently handle cleaning up any leftover objects, and so to assist you with that we have provided the below scripts to help guide you through doing that.
```
Charles-Gagnon, line 181:
```suggestion

The Azure SQL Trigger does not currently handle automatically cleaning up any leftover objects, and so we have provided the below scripts to help guide you through doing that.
```

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -321,8 +321,12 @@ private async Task<long> CreateGlobalStateTableAsync(SqlConnection connection, S
                         UserFunctionID char(16) NOT NULL,
                         UserTableID int NOT NULL,
                         LastSyncVersion bigint NOT NULL,
+                        LastAccessTime Datetime NULL,
```
Charles-Gagnon, line 324: Let's change this to NOT NULL and default to GETDATE. That makes it simpler, and there's no reason I can think of to have this actually be NULL

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -321,8 +321,12 @@ private async Task<long> CreateGlobalStateTableAsync(SqlConnection connection, S
...
                         PRIMARY KEY (UserFunctionID, UserTableID)
                     );
+                ELSE IF NOT EXISTS(SELECT 1 FROM sys.columns WHERE Name = N'LastAccessTime'
```
chlafreniere, line 327: I would like for us to have an explicit test for a lease table where the LastAccessTime column doesn't exist to validate that the `ELSE IF NOT EXISTS` code path works as intended.

(Is this possible, by the way, in the real world?)

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -955,7 +955,7 @@ private SqlCommand BuildUpdateTablesPostInvocation(SqlConnection connection, Sql
                 IF @unprocessed_changes = 0 AND @current_last_sync_version < {newLastSyncVersion}
                 BEGIN
                     UPDATE {GlobalStateTableName}
-                    SET LastSyncVersion = {newLastSyncVersion}
+                    SET LastSyncVersion = {newLastSyncVersion}, LastAccessTime = GETDATE()
```
chlafreniere, line 958: Should we be using GETDATE() or GETUTCDATE()?

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+FROM az_func.GlobalState
+WHERE DATEDIFF(day, LastAccessTime, GETDATE()) > @CleanupAgeDays
+
+OPEN LeaseTable_Cursor;
+
+FETCH NEXT FROM LeaseTable_Cursor INTO @TableName, @UserFunctionId, @UserTableId;
+
+WHILE @@FETCH_STATUS = 0
+BEGIN
+    EXEC ('DROP TABLE IF EXISTS ' + @TableName);
```
Charles-Gagnon, line 205: Let's add print statements to these, put the appropriate one above each statement that drops/deletes (here and below)

`PRINT 'Dropping table ' + @TableName`
`PRINT 'Removing row from GlobalState for UserFunctionID = ' + @UserFunctionId + ' and UserTableID = ' + @UserTableId'`

docs/BindingsOverview.md
```diff
@@ -168,3 +169,93 @@ e.g.
...
+This log message is at the `Information` level, so make sure your log level is set correctly.
+
+```sql
+-- Deletes the specified lease table and removes it from GlobalState table.
+USE <Insert DATABASE name here>;
+DECLARE @TableName NVARCHAR(MAX) = <Insert lease table name here>; -- e.g. '[az_func].[Leases_84d975fca0f7441a_901578250]
+DECLARE @UserFunctionId char(16) = <Insert function ID here>; -- e.g. '84d975fca0f7441a' the first section of the lease table name [Leases_84d975fca0f7441a_901578250].
+DECLARE @UserTableId int = <Insert table ID here>; -- e.g. '901578250' the second section of the lease table name [Leases_84d975fca0f7441a_901578250].
+
+DROP TABLE IF EXISTS @TableName
```
Charles-Gagnon, line 230:
```suggestion
EXEC ('DROP TABLE IF EXISTS ' + @TableName);
```

Charles-Gagnon (commented):
Make sure you run through and test all the scripts - there was one bug I found doing it myself. Didn't see any others, but it'd be best to double check yourself.

src/TriggerBinding/SqlTriggerListener.cs
Charles-Gagnon: It will be for people with existing functions that just update the extension version but everything else stays the same. But that'll drop off as we march towards GA since after this change gets in no one should have that going forward.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: It shouldn't matter I think - it's all local to the server (so unless that changes timezones or something weird). 

That being said - using UTC seems reasonable to avoid that potential edge case. And that's a good general thing to default to for stored timestamps anyways.

MaddyDev: Defaulted to GETUTCDATE() to avoid potential edge cases.

src/TriggerBinding/SqlTriggerListener.cs
MaddyDev: I agree with @Charles here, but I can look into adding a test to be sure nothing breaks in the meantime.

Charles-Gagnon (approved):
I agree a test would be good, but the rest looks fine other than that.

MaddyDev:
Merging this in, will send a separate PR for the test.

Issue #860, opened 2023-05-24
Invalid operation: the connection has been broken.


marthijn:
I see a similar problem as #723 for the SQL Trigger.
Instead of a closed connection, now the connection has been broken. It occurred after a few month since upgrading to version **3.0.143-preview**. Restarting the function app solved the problem.
I'm currently updating my solution to the latest version. I'll let you know if the problem still occurs.

See the logs below, ordered by timestamp desc. The first occurrence is at 5/23/2023, 8:09:01.577

|Timestamp|Message                      |severityLevel|itemType                                     |
|---------|-----------------------------|-------------|---------------------------------------------|
|5/23/2023, 8:09:25.729 PM|TID:6 END ProcessingChanges  |0            |trace                                        |
|5/23/2023, 8:09:25.729 PM|TID:6 END GetTableChanges    |0            |trace                                        |
|5/23/2023, 8:09:25.729 PM|Failed to check for changes in table 'MY-TABLE' due to exception: System.InvalidOperationException. Exception message: The requested operation cannot be completed because the connection has been broken.|3            |trace                                        |
|5/23/2023, 8:09:25.729 PM|TID:6 BEGIN GetTableChanges  |0            |trace                                        |
|5/23/2023, 8:09:25.729 PM|TID:6 BEGIN ProcessingChanges State=CheckingForChanges|0            |trace                                        |
|5/23/2023, 8:09:24.738 PM|TID:6 Delaying for 1000ms    |0            |trace                                        |
|5/23/2023, 8:09:24.738 PM|TID:6 END ProcessingChanges  |0            |trace                                        |
|5/23/2023, 8:09:24.738 PM|TID:6 END GetTableChanges    |0            |trace                                        |
|5/23/2023, 8:09:24.738 PM|Failed to check for changes in table 'MY-TABLE' due to exception: System.InvalidOperationException. Exception message: The requested operation cannot be completed because the connection has been broken.|3            |trace                                        |
|5/23/2023, 8:09:24.738 PM|TID:6 BEGIN GetTableChanges  |0            |trace                                        |
|5/23/2023, 8:09:24.738 PM|TID:6 BEGIN ProcessingChanges State=CheckingForChanges|0            |trace                                        |
|5/23/2023, 8:09:24.554 PM|TID:6 ReleaseRowsLock - RenewLeases|0            |trace                                        |
[...]

PR #862, opened 2023-05-26
Add additional check for closed/broken connection errors

Charles-Gagnon:
For https://github.com/Azure/azure-functions-sql-extension/issues/860

Simple check to try and catch these extra errors, there's a bunch of InvalidOperation exceptions that can be thrown apparently (see other ones in this file https://github.com/dotnet/SqlClient/blob/f478be5bf3597f5c0f0940ac471c6b481c38416f/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlUtil.cs#L1036) and trying to catch them all explicitly is likely to miss more. So for now just check the message for both "connection" and "closed" or "broken". This should be enough to not catch any other exceptions that might be thrown until we can get the SqlClient update and hopefully have that fix these types of issues.

Another option would be to use the 5.2.0 preview, but I'm a little hesitant to do that without thorough testing given issues we've had with updating before in ADS, so given that this doesn't seem to be a widespread issue I think the message checking is a safer approach.

I'll be leaving the issue open for the time being so we can track updating SqlClient when they release 5.2.0.

### DIFF
--- a/src/SqlBindingUtilities.cs
+++ b/src/SqlBindingUtilities.cs
@@ -250,8 +250,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         /// <returns>True if the exception is a fatal SqlClientException, false otherwise</returns>
         internal static bool IsFatalSqlException(this Exception e)
         {
+            string lowerMessage = e.Message.ToLowerInvariant();
             // Most SqlExceptions wrap the original error from the native driver, so make sure to check both
-            return (e as SqlException)?.Class >= 20 || (e.InnerException as SqlException)?.Class >= 20;
+            return (e as SqlException)?.Class >= 20
+                || (e.InnerException as SqlException)?.Class >= 20
+                // TEMPORARY - Not all exceptions thrown by SqlClient are SqlExceptions, and the current SqlClient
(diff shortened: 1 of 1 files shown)


Charles-Gagnon:
Hm, test failure due to this

```
[3229] [2023-05-26T18:28:59.965Z] Function loader reset. Failed compilation result will not be cached.
[3229] [2023-05-26T18:28:59.977Z] Creating function descriptor for function AddProductMissingColumns failed
[3229] [2023-05-26T18:28:59.977Z] Microsoft.Azure.WebJobs.Script: C# compilation service error: Cannot enqueue data after PromiseNotToEnqueue. Microsoft.CodeAnalysis: Cannot enqueue data after PromiseNotToEnqueue.
```

Likely unrelated, looking into what might be causing this though...

Charles-Gagnon:
womp 😢 

https://github.com/Azure/azure-functions-host/issues/999

I'll open a separate issue for tracking. Given that this is the first time we've seen it probably not worth investing time in fixing right now, but we'll keep an eye on it.

lucyzhang929 (approved):

PR #866, opened 2023-05-31
Merge latest from main into trigger

Charles-Gagnon:


MaddyDev (approved):

Charles-Gagnon:
@MaddyDev FYI, I had to fix up the stuff you added slightly so it works correctly. There were two issues: 

1. It was hardcoded to 99.99.99 for the package version, but that is going to be different in the pipelines. We should use the actual version instead
2. We need to pass the correct version to the dotnet pack call as well, otherwise it'll always use the default 99.99.99 (which then causes failures later as it tries to find the package version from the pipeline, e.g. 1.x/2.x)

I'll backport these fixes to main once the PR builds are passing and I get it merged in here.

PR #868, opened 2023-06-02
Backport fix for nuget packaging during build

Charles-Gagnon:
Fixes two issues: 

* It was hardcoded to 99.99.99 for the package version, but that is going to be different in the pipelines. We should use the actual version instead
* We need to pass the correct version to the dotnet pack call as well, otherwise it'll always use the default 99.99.99 (which then causes failures later as it tries to find the package version from the pipeline, e.g. 1.x/2.x)

See https://github.com/Azure/azure-functions-sql-extension/pull/868#issuecomment-1574172708 for full explanation of why this wasn't caught earlier and what issues it causes.

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -44,6 +44,6 @@ stages:
       parameters:
         solution: '$(solution)'
         configuration: '$(configuration)'
-        nugetVersion: '1.0.0-test'
-        binariesVersion: '1.0.0'
-        supportedVersion: '1.*-*'
+        nugetVersion: '99.99.99-test'
+        binariesVersion: '99.99.99'
+        supportedVersion: '99.99.99'
--- a/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Sql.csproj
@@ -36,7 +36,7 @@
   <Target Name="PackAndCopyNupkg" AfterTargets="Build">
     <!-- Removed the GeneratePackageOnBuild and adding this explicit Pack command to run post build
     and also adding the copy package to local-packages to be available for the worker extension project. -->
-    <Exec Command="dotnet pack &quot;$(MSBuildProjectDirectory)\$(PackageId).csproj&quot; --no-build --include-symbols" />
+    <Exec Command="dotnet pack &quot;$(MSBuildProjectDirectory)\$(PackageId).csproj&quot; --no-build --include-symbols -p:Version=$(Version)" />
     <ItemGroup>
       <_Packages Include=".\bin\$(Configuration)\*.nupkg" />
     </ItemGroup>
@@ -44,7 +44,7 @@
     <Message Text="Copied sql .nupkg to local-packages" Importance="high" />
   </Target>
   <Target Name="RemoveNugetPackageCache" BeforeTargets="Build">
-    <RemoveDir Directories="$(NugetPackageRoot)\$(PackageId.ToLower())\99.99.99"></RemoveDir>
-    <Message Text="Deleted nuget cache for $(PackageId.ToLower())\99.99.99" Importance="high" />
+    <RemoveDir Directories="$(NugetPackageRoot)\$(PackageId.ToLower())\$(Version)"></RemoveDir>
+    <Message Text="Deleted nuget cache for $(PackageId.ToLower())\$(Version)" Importance="high" />
   </Target>
 </Project>


MaddyDev:
Even without passing the version to the pack command, it is not using the hardcoded 99.99.99 version on pipeline builds ex: https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=201942&view=logs&j=ad533c73-7a95-5e12-cf66-f0b3921575e3&t=3da72317-fc95-51b8-57c3-d32106c04e6
IIRC removing the nuget package is added specifically for local builds with 99.99.99 version.

Charles-Gagnon:
Huh, good point on the version. For some reason the PR builds seem to be different - see https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=201745&view=logs&j=f9f7e0bb-717e-5ea9-5f1c-3ee3ad57f3f6&t=6a5360e1-b584-59ac-4f6e-ca10047a7989

```
    Successfully created package 'D:\a\1\s\src\bin\Debug\Microsoft.Azure.WebJobs.Extensions.Sql.99.99.99.nupkg'.
    Successfully created package 'D:\a\1\s\src\bin\Debug\Microsoft.Azure.WebJobs.Extensions.Sql.99.99.99.snupkg'.
```

Let me keep looking...

Charles-Gagnon:
Right, yeah so this was something that was changed/broken with your PR https://github.com/Azure/azure-functions-sql-extension/pull/853 where you changed the package to be generated by calling dotnet pack directly. Because of this it wasn't passing in the version - so defaulted to 99.99.99

Before PR : https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=200531&view=logs&j=ad533c73-7a95-5e12-cf66-f0b3921575e3&t=3da72317-fc95-51b8-57c3-d32106c04e66

After PR : https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=200631&view=logs&j=ad533c73-7a95-5e12-cf66-f0b3921575e3&t=3da72317-fc95-51b8-57c3-d32106c04e66

Notice that we're now getting a bunch of warnings like this : 

`##[warning]/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ltmpwdy0.3st/WorkerExtensions.csproj(0,0): Warning NU1603: Microsoft.Azure.Functions.Worker.Extensions depends on Microsoft.Azure.WebJobs.Extensions.Sql (>= 2.0.235) but Microsoft.Azure.WebJobs.Extensions.Sql 2.0.235 was not found. An approximate best match of Microsoft.Azure.WebJobs.Extensions.Sql 99.99.99 was resolved.`

For the product build pipeline this worked by coincidence - since the version we were looking for is always going to be higher than any released version it should always default to 99.99.99 and pick up the correct package.

But for the PR pipeline we hardcode it to 1.0.0 - https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/build-pr.yml#L48

This means it then resolves to one of the released versions, see https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=201745&view=logs&j=f9f7e0bb-717e-5ea9-5f1c-3ee3ad57f3f6&t=6a5360e1-b584-59ac-4f6e-ca10047a7989

`##[warning]C:\Users\VssAdministrator\AppData\Local\Temp\fes35lxx.j44\WorkerExtensions.csproj(0,0): Warning NU1603: Microsoft.Azure.Functions.Worker.Extensions depends on Microsoft.Azure.WebJobs.Extensions.Sql (>= 1.0.0) but Microsoft.Azure.WebJobs.Extensions.Sql 1.0.0 was not found. An approximate best match of Microsoft.Azure.WebJobs.Extensions.Sql 2.0.144 was resolved.`

Which is a big bug since it means we're testing against the released version of the package (for the OOP tests only though). I noticed this because it completely failed on the trigger branch since that version of the package didn't have the trigger bindings.

Charles-Gagnon:
You're right that clearing the nuget cache is only really relevant for local dev scenarios (since the agent machines shouldn't have anything cached). But it doesn't hurt to do it and is what we should have done in the first place.

Charles-Gagnon:
I'm going to change the PR builds to use 99.99.99 as well. This should prevent issues like that from silently failing in the future - and since this isn't an actual build we intend to possibly release it's better to make that clear.

MaddyDev:
> Right, yeah so this was something that was changed/broken with your PR #853 where you changed the package to be generated by calling dotnet pack directly. Because of this it wasn't passing in the version - so defaulted to 99.99.99
> 
> Before PR : https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=200531&view=logs&j=ad533c73-7a95-5e12-cf66-f0b3921575e3&t=3da72317-fc95-51b8-57c3-d32106c04e66
> 
> After PR : https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=200631&view=logs&j=ad533c73-7a95-5e12-cf66-f0b3921575e3&t=3da72317-fc95-51b8-57c3-d32106c04e66
> 
> Notice that we're now getting a bunch of warnings like this :
> 
> `##[warning]/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ltmpwdy0.3st/WorkerExtensions.csproj(0,0): Warning NU1603: Microsoft.Azure.Functions.Worker.Extensions depends on Microsoft.Azure.WebJobs.Extensions.Sql (>= 2.0.235) but Microsoft.Azure.WebJobs.Extensions.Sql 2.0.235 was not found. An approximate best match of Microsoft.Azure.WebJobs.Extensions.Sql 99.99.99 was resolved.`
> 
> For the product build pipeline this worked by coincidence - since the version we were looking for is always going to be higher than any released version it should always default to 99.99.99 and pick up the correct package.
> 
> But for the PR pipeline we hardcode it to 1.0.0 - https://github.com/Azure/azure-functions-sql-extension/blob/main/builds/azure-pipelines/build-pr.yml#L48
> 
> This means it then resolves to one of the released versions, see https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=201745&view=logs&j=f9f7e0bb-717e-5ea9-5f1c-3ee3ad57f3f6&t=6a5360e1-b584-59ac-4f6e-ca10047a7989
> 
> `##[warning]C:\Users\VssAdministrator\AppData\Local\Temp\fes35lxx.j44\WorkerExtensions.csproj(0,0): Warning NU1603: Microsoft.Azure.Functions.Worker.Extensions depends on Microsoft.Azure.WebJobs.Extensions.Sql (>= 1.0.0) but Microsoft.Azure.WebJobs.Extensions.Sql 1.0.0 was not found. An approximate best match of Microsoft.Azure.WebJobs.Extensions.Sql 2.0.144 was resolved.`
> 
> Which is a big bug since it means we're testing against the released version of the package (for the OOP tests only though). I noticed this because it completely failed on the trigger branch since that version of the package didn't have the trigger bindings.

Ohh so the version needs to be passed, thanks @charles for looking into it and for addressing the bug introduced by #853.

MaddyDev (approved):

MaddyDev (approved):

PR #871, opened 2023-06-07
add troubleshooting information to the overview doc

MaddyDev:
Added troubleshooting section to the BindingsOverview doc.

### DIFF
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -271,6 +271,6 @@ DEALLOCATE LeaseTable_Cursor;
 
 ## Troubleshooting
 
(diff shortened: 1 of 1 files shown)


Charles-Gagnon (approved):
I agree that having all those things in our docs is a bit much. It might be interesting to have a set of public TSGs that can be used both by us and users running into issues though...something to think about in the future. But for now this is a great start, thank you!

VasuBhog (commented):
Agree we could currently surface this here but also look into adding an MS Learn troubleshooting page that internal and external users could utilize and provide feedback on.

lucyzhang929 (approved):

lucyzhang929:
CosmosDb trigger has a MS doc for troubleshooting common issues: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-changefeed-functions. Maybe we can create a similar doc.

PR #876, opened 2023-06-27
Add sql trigger test for different data types

lucyzhang929:


test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -654,7 +654,91 @@ public async void LastAccessTimeColumn_Created_OnStartup()
...
+                ".1, " + // Real
+                $"CONVERT(DATE, '{datetime}'), " + // Date
+                $"CONVERT(DATETIME, '{datetime}'), " + // Datetime
+                $"CONVERT(DATETIME2, '{datetime}'), " + // Datetime2
+                $"CONVERT(DATETIMEOFFSET, '{datetime}'), " + // DatetimeOffset
+                $"CONVERT(SMALLDATETIME, '{datetime}'), " + // SmallDatetime
+                $"CONVERT(TIME, '{datetime}'), " + // Time
+                "'test', " + // CharType
+                "'test', " + // Varchar
+                "'test', " + // Nchar
```
lucyzhang929, line 728: When I used NCHAR(0xD84C) here instead of 'test' (which is what we do in the [input/output binding tests](https://github.com/Azure/azure-functions-sql-extension/blob/ff2c3096305c6d41be969b6b75c17a5c57e345a7/test/Integration/SqlInputBindingIntegrationTests.cs#L168C29-L168C29)) the test failed in the pipeline but passed locally on my machine. Will try to keep investigating why it was failing in the pipelines but using 'test' for now so I can get the test in first.

Charles-Gagnon (approved):

PR #880, opened 2023-07-07
Fix code coverage in trigger branch

lucyzhang929:
Updating the integration tests so that for the [CSharp dotnet test task](https://github.com/Azure/azure-functions-sql-extension/blob/1a0fe10c0f2d962a17c13f1b0638648c2486d8c3/builds/azure-pipelines/template-steps-build-test.yml#L181), only the CSharp function host is started. It's not clear why having multiple function hosts start caused the code coverage to not show up (I added logging and verified all hosts were being disposed).

Ad hoc build where code coverage number shows (71.67%): https://mssqltools.visualstudio.com/CrossPlatBuildScripts/_build/results?buildId=205427&view=results

### DIFF
--- a/test/Integration/IntegrationTestFixture.cs
+++ b/test/Integration/IntegrationTestFixture.cs
@@ -63,21 +63,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
                     this.StartHost(Path.Combine(binPath, "SqlExtensionSamples", "CSharp"), TestUtils.GetPort(lang), this.SampleFunctions);
                     // this.StartHost(Path.Combine(binPath), TestUtils.GetPort(lang, true), this.TestFunctions);
                 }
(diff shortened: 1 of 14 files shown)


test/Integration/IntegrationTestFixture.cs
```diff
@@ -66,7 +66,10 @@ public IntegrationTestFixture()
         private void StartFunctionHosts()
         {
             string binPath = TestUtils.GetPathToBin();
-            foreach (SupportedLanguages lang in Enum.GetValues(typeof(SupportedLanguages)))
+            string languages = Environment.GetEnvironmentVariable("LANGUAGES_TO_TEST");
```
Charles-Gagnon, line 69: Could you add a comment here explaining why we're doing this? Will help make sure we don't break it in the future 😄

Charles-Gagnon (approved):

MaddyDev (approved):

Issue #878, opened 2023-07-06
Enable dependabot for dependency updates


MaddyDev:
```[tasklist]
### Tasks
- [x] Check and update all dependency libraries to the workable latest versions.
- [ ] Integrating Dependabot
```

Charles-Gagnon:
Reopening for dependabot enabling

PR #887, opened 2023-07-20
update dependencies

MaddyDev:
Updated all the packages to the latest available versions except for xunit.
Created https://github.com/NuGet/Home/issues/12760 to track the issue we're running into.

### DIFF
--- a/test/Common/SupportedLanguagesTestAttribute.cs
+++ b/test/Common/SupportedLanguagesTestAttribute.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Common
         public override IEnumerable<object[]> GetData(MethodInfo testMethod)
         {
             if (testMethod == null) { throw new ArgumentNullException(nameof(testMethod)); }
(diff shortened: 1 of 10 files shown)


test/Common/SupportedLanguagesTestAttribute.cs
```diff
@@ -46,7 +46,7 @@ public SqlInlineDataAttribute(params object[] args)
         public override IEnumerable<object[]> GetData(MethodInfo testMethod)
         {
             if (testMethod == null) { throw new ArgumentNullException(nameof(testMethod)); }
-            if (this.testData.Count == 0) { throw new EmptyException(nameof(testMethod)); }
+            if (this.testData.Count == 0) { throw new ArgumentNullException(nameof(testMethod)); }
```
Charles-Gagnon, line 49: This isn't really accurate - could you just throw a general Exception and then add a message indicating what the actual issue is instead?

Charles-Gagnon (approved):

chlafreniere:
Have we thought about enabling dependabot?

Charles-Gagnon:
We had/have https://github.com/Azure/azure-functions-sql-extension/issues/878 for tracking, but it got closed. Reopening to follow up on that...

Issue #882, opened 2023-07-10
SQL trigger firing from stage slot when it should be disabled


grantpalmer-carvana:
We are running into some issues where the SQL table trigger is occasionally firing from the stage slot while swapping slots during a release. We have slot settings in place to disable the sql trigger function unless it is the prod slot. 

For these logs, the values used are:
``` 
var hostName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
var slotName = Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME");
var slotSettingValue = Environment.GetEnvironmentVariable("AzureWebJobs.CheckoutRequestTrigger.Disabled");
```

We had a branch deployed to test that had logging that didn't include the slot setting value as can be seen by the trace from 12:06. While swapping slots at 12:09:27, the sql trigger fired from the stage slot, had a function disabled slot setting value of false which corresponds to the expected value for the prod slot, and had msi/hostname corresponding to the prod slot of the function app.  The app insights telemetry role name for that trace is that of the stage-slot. 

The trace from 12:10:05 shows a log running off of the previous iteration of the code since it doesn't contain the slot setting portion of the log. The app insights telemetry role name for this trace is that of the prod-slot. It also would've needed the prod-slot disabled slot setting to be false in order for this log to be created. 

Taking all of this into account, it appears as if the sql trigger is currently able to fire from the stage slot while doing a slot swap during a release despite there being a slot setting to disable the sql trigger function for stage. When it does so, it uses the prod slot msi/hostname and app settings although it is in the stage-slot. When this occurs, an app lease table is created for the stage slot as well. Following the completion of performing the slot swap, the stage-slot function is disabled and does not continue to trigger from changes to the sql table that it tracks so this is only seen when performing a slot swap.

![image (1)](https://github.com/Azure/azure-functions-sql-extension/assets/131677375/29638ed2-f0d3-45f3-b1a4-3c0da3dc391d)

```[tasklist]
### Tasks
- [x] Add optional LeaseTableNameSetting to SqlTriggerAttribute
- [x] Release new extension
- [x] Release new worker extension
- [x] Release new preview extension bundle
- [x] Release new java library
- [x] Update python sql decorator
- [x] Update docs
```

lucyzhang929:
[...]

PR #893, opened 2023-07-27
Add LeasesTableName to SqlTriggerAttribute

lucyzhang929:
Addresses https://github.com/Azure/azure-functions-sql-extension/issues/882. Adding LeasesTableName option to the SqlTriggerAttribute will prevent two leases tables from being created during a slot swap.

If the LeasesTableName is not included, we default back to using the [host id](https://github.com/Azure/azure-functions-sql-extension/blob/f3eafb01c092036bad3d1e557af2dc117640d91d/src/TriggerBinding/SqlTriggerBinding.cs#L98C36-L98C58) to generate the lease table name.

### DIFF
--- a/Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
+++ b/Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
@@ -13,12 +13,12 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Sql
         /// </summary>
         /// <param name="tableName">Name of the table to watch for changes.</param>
         /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
-        /// <param name="leasesTableNameSetting">The name of the app setting where the leases table name is stored</param>
-        public SqlTriggerAttribute(string tableName, string connectionStringSetting, string leasesTableNameSetting = null)
+        /// <param name="leasesTableName">Optional - The name of the table used to store leases</param>
+        public SqlTriggerAttribute(string tableName, string connectionStringSetting, string leasesTableName = null)
         {
             this.TableName = tableName ?? throw new ArgumentNullException(nameof(tableName));
             this.ConnectionStringSetting = connectionStringSetting ?? throw new ArgumentNullException(nameof(connectionStringSetting));
-            this.LeasesTableNameSetting = leasesTableNameSetting;
+            this.LeasesTableName = leasesTableName;
         }
 
         /// <summary>
@@ -32,11 +32,11 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Sql
         public string TableName { get; }
 
         /// <summary>
-        /// Name of the app setting containing the leases table name.
+        /// Name of the table used to store leases.
         /// If not specified, the leases table name will be Leases_{FunctionId}_{TableId}
         /// More information on how this is generated can be found here
         /// https://github.com/Azure/azure-functions-sql-extension/blob/release/trigger/docs/TriggerBinding.md#az_funcleases_
         /// </summary>
-        public string LeasesTableNameSetting { get; }
+        public string LeasesTableName { get; }
     }
 }
--- a/Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
+++ b/Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
@@ -13,10 +13,12 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Sql
         /// </summary>
         /// <param name="tableName">Name of the table to watch for changes.</param>
         /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
-        public SqlTriggerAttribute(string tableName, string connectionStringSetting)
+        /// <param name="leasesTableNameSetting">The name of the app setting where the leases table name is stored</param>
+        public SqlTriggerAttribute(string tableName, string connectionStringSetting, string leasesTableNameSetting = null)
         {
             this.TableName = tableName ?? throw new ArgumentNullException(nameof(tableName));
             this.ConnectionStringSetting = connectionStringSetting ?? throw new ArgumentNullException(nameof(connectionStringSetting));
+            this.LeasesTableNameSetting = leasesTableNameSetting;
         }
 
         /// <summary>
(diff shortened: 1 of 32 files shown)


Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
```diff
@@ -13,12 +13,21 @@ public sealed class SqlTriggerAttribute : TriggerBindingAttribute
         /// </summary>
         /// <param name="tableName">Name of the table to watch for changes.</param>
         /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
-        public SqlTriggerAttribute(string tableName, string connectionStringSetting)
+        /// <param name="leasesTableName">Optional - The name of the table used to store leases</param>
```
Charles-Gagnon, line 16: Add description of default value if not specified

Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
```diff
@@ -28,5 +37,13 @@ public SqlTriggerAttribute(string tableName, string connectionStringSetting)
         /// Name of the table to watch for changes.
         /// </summary>
         public string TableName { get; }
+
+        /// <summary>
+        /// Name of the table used to store leases.
+        /// If not specified, the leases table name will be Leases_{FunctionId}_{TableId}
+        /// More information on how this is generated can be found here
+        /// https://github.com/Azure/azure-functions-sql-extension/blob/release/trigger/docs/TriggerBinding.md#az_funcleases_
```
Charles-Gagnon, line 45: If we don't already, let's make sure we have an "update docs to remove reference to trigger branch" as a post-merge-to-main task.

src/TriggerBinding/SqlTriggerAttribute.cs
```diff
@@ -18,12 +18,21 @@ public sealed class SqlTriggerAttribute : Attribute
         /// </summary>
         /// <param name="tableName">Name of the table to watch for changes.</param>
         /// <param name="connectionStringSetting">The name of the app setting where the SQL connection string is stored</param>
-        public SqlTriggerAttribute(string tableName, string connectionStringSetting)
+        /// <param name="leasesTableName">Optional - The name of the table used to store leases</param>
```
Charles-Gagnon, line 21: Include default value description

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -123,7 +126,8 @@ public async Task StartAsync(CancellationToken cancellationToken)
                     IReadOnlyList<(string name, string type)> primaryKeyColumns = await GetPrimaryKeyColumnsAsync(connection, userTableId, this._logger, this._userTable.FullName, cancellationToken);
                     IReadOnlyList<string> userTableColumns = await this.GetUserTableColumnsAsync(connection, userTableId, cancellationToken);
 
-                    string leasesTableName = string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{this._userFunctionId}_{userTableId}");
+                    string leasesTableName = String.IsNullOrEmpty(this._userDefinedLeasesTableName) ? string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{this._userFunctionId}_{userTableId}") :
+                        string.Format(CultureInfo.InvariantCulture, UserDefinedLeasesTableNameFormat, $"{this._userDefinedLeasesTableName}");
```
Charles-Gagnon, line 130: This needs to be quote-escaped

src/TriggerBinding/SqlTriggerMetricsProvider.cs
```diff
@@ -99,7 +101,8 @@ private async Task<long> GetUnprocessedChangeCountAsync()
         private SqlCommand BuildGetUnprocessedChangesCommand(SqlConnection connection, SqlTransaction transaction, IReadOnlyList<(string name, string type)> primaryKeyColumns, int userTableId)
         {
             string leasesTableJoinCondition = string.Join(" AND ", primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = l.{col.name.AsBracketQuotedString()}"));
-            string leasesTableName = string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{this._userFunctionId}_{userTableId}");
+            string leasesTableName = String.IsNullOrEmpty(this._userDefinedLeasesTableName) ? string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{this._userFunctionId}_{userTableId}") :
```
Charles-Gagnon, line 104: Could we move this up to a common location instead of having to calculate the name in each place separately? If not, then at least we should have a common helper function so the logic is guaranteed to be the same.

Worker.Extensions.Sql/src/SqlTriggerAttribute.cs
lucyzhang929: Created https://github.com/Azure/azure-functions-sql-extension/issues/894 to track.

src/TriggerBinding/SqlTriggerMetricsProvider.cs
lucyzhang929: Added a helper function since moving to a common location would require the connection to be opened in SqlTriggerListener to get the userTableId.

src/TriggerBinding/SqlTriggerUtils.cs
```diff
@@ -111,5 +113,17 @@ internal static async Task<int> GetUserTableIdAsync(SqlConnection connection, Sq
...
+        /// <summary>
+        /// Returns the formatted leases table name. If userDefinedLeasesTableName is null, the default name Leases_{FunctionId}_{TableId} is used.
+        /// </summary>
+        /// <param name="userDefinedLeasesTableName">Leases table name defined by the user</param>
+        /// <param name="userTableId">SQL object ID of the user table</param>
+        /// <param name="userFunctionId">Unique identifier for the user function</param>
+        internal static string GetLeasesTableName(string userDefinedLeasesTableName, string userFunctionId, int userTableId)
+        {
+            return string.IsNullOrEmpty(userDefinedLeasesTableName) ? string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{userFunctionId}_{userTableId}") :
+                string.Format(CultureInfo.InvariantCulture, UserDefinedLeasesTableNameFormat, $"{userDefinedLeasesTableName}");
```
Charles-Gagnon, line 126: This still needs to be quote-escaped

lucyzhang929: Do you mean to call userDefinedLeasesTableName.AsSingleQuoteEscapedString() here?

Charles-Gagnon: AsBrackQuotedString since we're quoting it with brackets here, but yes (and remove the [] from the format string since it will be added by that call)

Charles-Gagnon (approved):
Make sure you update the docs as well, https://github.com/Azure/azure-functions-sql-extension/blob/release/trigger/docs/TriggerBinding.md#az_funcleases_ for sure.

lucyzhang929:
Will update the docs in a separate PR - added it as a task to the original issue: https://github.com/Azure/azure-functions-sql-extension/issues/882.

Issue #437, opened 2022-11-02
Enhance debug logging when retrieving changed rows


Charles-Gagnon:
Currently when we get the changed rows the query skips any that are changed but have a lease on them. Having debug logging saying when this is the case can help diagnose issues since otherwise it's very difficult to tell whether there weren't any changed rows found or if there were but they're currently being locked by other instances. 

Two options to consider : 

1. Always return all changed rows but then have a column "IsLeaseLocked" that says whether it's currently held by a lease - the debug logging can then output something like `0 available changed rows (100 found with lease locks)`
2. Check at runtime if debug logging is enabled and if so run a separate query that specifically gets the rows that are held by leases. This would make it so the normal case (no debug logging enabled) is not affected by the extra overhead of returning the rows each time

PR #900, opened 2023-08-03
GetLeaseLockedRows for debug logging

MaddyDev:
Added IsLeaseLocked for #437 

Can improve it to add a check for only getting all the rows when log level is debug, will look into that but for now adding this in the normal flow.

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -306,7 +306,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                             getChangesDurationMs = commandSw.ElapsedMilliseconds;
                         }
                         // Log the number of rows
-                        //this._logger.LogDebug($"Executed GetChanges in GetTableChanges. Total changes: {rows.Count} Locked rows: {rows.FindAll(x => from n in x where (n.Value.isLeaseLocked = 1 select n.Value)}");
+                        this._logger.LogDebug($"Executed GetChanges in GetTableChanges. Total changes: {rows.Count} Locked rows: {rows.FindAll(dict => dict.ContainsKey("IsLeaseLocked") && (int)dict["IsLeaseLocked"] == 1).Count}");
 
                         // If changes were found, acquire leases on them.
                         if (rows.Count > 0)
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -305,7 +305,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                             getChangesDurationMs = commandSw.ElapsedMilliseconds;
                         }
-                        // Get the number of lease locked rows and log them if > 0.
+                        // Also get the number of rows that currently have lease locks on them
+                        // This can help with supportability by allowing a customer to see when a
+                        // trigger was processed successfully but returned fewer rows than expected
+                        // because of the rows being locked.
                         int leaseLockedRowCount = await this.GetLeaseLockedRowCount(connection, transaction);
                         if (rows.Count > 0 || leaseLockedRowCount > 0)
                         {
@@ -816,11 +819,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
         }
 
         /// <summary>
-        /// Returns the number of changes(rows) on the user's table that are actively locked by other leases.
+        /// Returns the number of changes(rows) on the user's table that are actively locked by other leases OR returns -1 on exception.
         /// </summary>
         /// <param name="connection">The connection to add to the SqlCommand</param>
         /// <param name="transaction">The transaction to add to the SqlCommand</param>
-        /// <returns>The number of rows locked by leases</returns>
+        /// <returns>The number of rows locked by leases or -1 on exception</returns>
         private async Task<int> GetLeaseLockedRowCount(SqlConnection connection, SqlTransaction transaction)
         {
             string leasesTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = l.{col.name.AsBracketQuotedString()}"));
@@ -854,7 +857,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             {
                 this._logger.LogError($"Failed to query count of lease locked changes for table '{this._userTable.FullName}' due to exception: {ex.GetType()}. Exception message: {ex.Message}");
                 TelemetryInstance.TrackException(TelemetryErrorName.GetLeaseLockedRowCount, ex, null, new Dictionary<TelemetryMeasureName, double>() { { TelemetryMeasureName.GetLockedRowCountDurationMs, getLockedRowCountDurationMs } });
-                throw;
+                leaseLockedRowsCount = -1;
             }
             return leaseLockedRowsCount;
         }
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -305,6 +305,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                             getChangesDurationMs = commandSw.ElapsedMilliseconds;
                         }
+                        // Log the number of rows
+                        //this._logger.LogDebug($"Executed GetChanges in GetTableChanges. Total changes: {rows.Count} Locked rows: {rows.FindAll(x => from n in x where (n.Value.isLeaseLocked = 1 select n.Value)}");
 
                         // If changes were found, acquire leases on them.
                         if (rows.Count > 0)
@@ -795,14 +797,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     c.SYS_CHANGE_OPERATION,
                     l.{LeasesTableChangeVersionColumnName},
                     l.{LeasesTableAttemptCountColumnName},
-                    l.{LeasesTableLeaseExpirationTimeColumnName}
+                    l.{LeasesTableLeaseExpirationTimeColumnName},
+                    CASE WHEN (l._az_func_LeaseExpirationTime < SYSDATETIME()) THEN 0 else 1 END AS IsLeaseLocked
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
                 LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
                 LEFT OUTER JOIN {this._userTable.BracketQuotedFullName} AS u ON {userTableJoinCondition}
                 WHERE
                     (l.{LeasesTableLeaseExpirationTimeColumnName} IS NULL AND
                        (l.{LeasesTableChangeVersionColumnName} IS NULL OR l.{LeasesTableChangeVersionColumnName} < c.{SysChangeVersionColumnName}) OR
-                        l.{LeasesTableLeaseExpirationTimeColumnName} < SYSDATETIME()
+                        l.{LeasesTableLeaseExpirationTimeColumnName} IS NOT NULL
                     ) AND
                     (l.{LeasesTableAttemptCountColumnName} IS NULL OR l.{LeasesTableAttemptCountColumnName} < {MaxChangeProcessAttemptCount})
                 ORDER BY c.{SysChangeVersionColumnName} ASC;";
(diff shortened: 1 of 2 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -795,14 +805,15 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                    l.{LeasesTableLeaseExpirationTimeColumnName},
+                    CASE WHEN (l.{LeasesTableLeaseExpirationTimeColumnName} IS NULL OR l.{LeasesTableLeaseExpirationTimeColumnName} < SYSDATETIME()) THEN 0 else 1 END AS IsLeaseLocked
                 FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
                 LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
                 LEFT OUTER JOIN {this._userTable.BracketQuotedFullName} AS u ON {userTableJoinCondition}
                 WHERE
                     (l.{LeasesTableLeaseExpirationTimeColumnName} IS NULL AND
                        (l.{LeasesTableChangeVersionColumnName} IS NULL OR l.{LeasesTableChangeVersionColumnName} < c.{SysChangeVersionColumnName}) OR
-                        l.{LeasesTableLeaseExpirationTimeColumnName} < SYSDATETIME()
+                        l.{LeasesTableLeaseExpirationTimeColumnName} IS NOT NULL
```
Charles-Gagnon, line 816: We only select the max batch size in this query, so you won't be getting the full number of locked rows.

We'll probably want to just do a separate query to get the locked lease count. That also means we don't have to get the row data for the locked rows (which isn't necessary and could impact performance)

MaddyDev: get only the lease locked rows count in a separate query wouldn't impact the performance, I'll go ahead and update to that.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +813,53 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
 
+        /// <summary>
+        /// Returns the query to get for number of changes(rows) on the user's table that are actively locked by other leases.
+        /// </summary>
+        /// <param name="connection">The connection to add to the returned SqlCommand</param>
+        /// <param name="transaction">The transaction to add to the returned SqlCommand</param>
+        /// <returns>The number of rows locked by leases</returns>
+        private async Task<int> GetLeaseLockedRowCount(SqlConnection connection, SqlTransaction transaction)
+        {
+            string userTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = u.{col.name.AsBracketQuotedString()}"));
```
Charles-Gagnon, line 824: We shouldn't need to join on the user table, right? Just getting the number of changes that are locked should be enough.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -305,6 +305,9 @@ private async Task GetTableChangesAsync(SqlConnection connection, CancellationTo
 
                             getChangesDurationMs = commandSw.ElapsedMilliseconds;
                         }
+                        // Get the Lease Locked rows and log the number of rows.
+                        int leaseLockedRowCount = await this.GetLeaseLockedRowCount(connection, transaction);
+                        this._logger.LogDebug($"Executed GetChangesCommand in GetTableChangesAsync. {rows.Count} available changed rows ({leaseLockedRowCount} found with lease locks).");
```
Charles-Gagnon, line 310: Let's only log this if `rows.Count > 0 || leaseLockedRowCount > 0` - otherwise we'll end up with a ton of messages where both values are 0 which isn't going to be much use.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +813,49 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                SELECT COUNT(*)
+                FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
+                LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
+                WHERE l.{LeasesTableLeaseExpirationTimeColumnName} IS NOT NULL AND l.{LeasesTableLeaseExpirationTimeColumnName} > SYSDATETIME()";
+
+            using (var getLeaseLockedRowCountCommand = new SqlCommand(getLeaseLockedrowCountQuery, connection, transaction))
+            {
+                var commandSw = Stopwatch.StartNew();
+                leaseLockedRows = (int)await getLeaseLockedRowCountCommand.ExecuteScalarAsyncWithLogging(this._logger, CancellationToken.None);
+                getLockedRowCountDurationMs = commandSw.ElapsedMilliseconds;
```
Charles-Gagnon, line 854: This doesn't seem to be used anywhere

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +813,49 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+        {
+            string leasesTableJoinCondition = string.Join(" AND ", this._primaryKeyColumns.Select(col => $"c.{col.name.AsBracketQuotedString()} = l.{col.name.AsBracketQuotedString()}"));
+            int leaseLockedRows = 0;
+            long getLockedRowCountDurationMs = 0L;
+            // Get the list of changes from CHANGETABLE that meet the following criteria:
+            // * Null LeaseExpirationTime AND (Null ChangeVersion OR ChangeVersion < Current change version for that row from CHANGETABLE)
+            //   OR
+            // * LeaseExpirationTime < Current Time
+            //
+            // The LeaseExpirationTime is only used for rows currently being processed - so if we see a
```
Charles-Gagnon, line 832: This comment needs to be updated to explain what this query is doing

Charles-Gagnon (changes requested):

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -822,18 +824,12 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+            int leaseLockedRowsCount = 0;
             long getLockedRowCountDurationMs = 0L;
-            // Get the list of changes from CHANGETABLE that meet the following criteria:
-            // * Null LeaseExpirationTime AND (Null ChangeVersion OR ChangeVersion < Current change version for that row from CHANGETABLE)
-            //   OR
-            // * LeaseExpirationTime < Current Time
+            // Get the count of changes from CHANGETABLE that meet the following criteria:
+            // * Not Null LeaseExpirationTime AND
+            // * LeaseExpirationTime > Current Time
             //
```
lucyzhang929, line 832: nit: extra line

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -305,7 +305,12 @@ private async Task GetTableChangesAsync(SqlConnection connection, CancellationTo
 
                             getChangesDurationMs = commandSw.ElapsedMilliseconds;
                         }
-
+                        // Get the number of lease locked rows and log them if > 0.
```
Charles-Gagnon, line 308:
```suggestion
                        // Also get the number of rows that currently have lease locks on them
                        // This can help with supportability by allowing a customer to see when a 
                        // trigger was processed successfully but returned fewer rows than expected
                        // because of the rows being locked.
```

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +815,50 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                    var commandSw = Stopwatch.StartNew();
+                    leaseLockedRowsCount = (int)await getLeaseLockedRowCountCommand.ExecuteScalarAsyncWithLogging(this._logger, CancellationToken.None);
+                    getLockedRowCountDurationMs = commandSw.ElapsedMilliseconds;
+                }
+            }
+            catch (Exception ex)
+            {
+                this._logger.LogError($"Failed to query count of lease locked changes for table '{this._userTable.FullName}' due to exception: {ex.GetType()}. Exception message: {ex.Message}");
+                TelemetryInstance.TrackException(TelemetryErrorName.GetLeaseLockedRowCount, ex, null, new Dictionary<TelemetryMeasureName, double>() { { TelemetryMeasureName.GetLockedRowCountDurationMs, getLockedRowCountDurationMs } });
+                throw;
```
Charles-Gagnon, line 857: Since this is just for debugging we shouldn't throw here - I'd suggest just returning -1 in that case so it's clear something went wrong but the rest of the trigger continues to run as expected.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +815,50 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                {AppLockStatements}
+
+                DECLARE @last_sync_version bigint;
+                SELECT @last_sync_version = LastSyncVersion
+                FROM {GlobalStateTableName}
+                WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {this._userTableId};
+
+                SELECT COUNT(*)
+                FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, @last_sync_version) AS c
+                LEFT OUTER JOIN {this._leasesTableName} AS l ON {leasesTableJoinCondition}
```
Charles-Gagnon, line 842: If leases table always guaranteed to be quoted here? If so let's update this var to be `_bracketedLeasesTableName`. And if not, then make sure you do that for users who have custom lease table names.

@lucyzhang929 to help verify this

Charles-Gagnon (changes requested):

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Make sure to call that out in the doc comment too in case we use this anywhere else for some reason.

src/TriggerBinding/SqlTableChangeMonitor.cs
MaddyDev: I've debugged this and saw the leasesTableName is bracketed. But I can use the _bracketedLeasesTableName just to be sure.

src/TriggerBinding/SqlTableChangeMonitor.cs
MaddyDev: return -1 instead of throwing the exception wouldn't help us right since we only log >0. I can remove the throw statement and continue the execution.

src/TriggerBinding/SqlTableChangeMonitor.cs
lucyzhang929: Yes we format the leasesTableName here: https://github.com/Azure/azure-functions-sql-extension/blob/04f64be48d0660471385be1d1407da71f701d2f6/src/TriggerBinding/SqlTriggerUtils.cs#L126.

MaddyDev: Thanks @lucyzhang929! Am going to leave it as is.

lucyzhang929: I think what Charles' comment meant is to update the variable name `_leasesTableName` to `_bracketedLeasesTableName` so it is clear that the name is correctly formatted.

MaddyDev: update _leasesTableName  to _bracketedLeasesTableName  everywhere? I can do it in a separate PR, don't want to do that change here.

Charles-Gagnon: Yup, that sounds good to me. Thanks for looking into it!

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -810,6 +818,50 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                    var commandSw = Stopwatch.StartNew();
+                    leaseLockedRowsCount = (int)await getLeaseLockedRowCountCommand.ExecuteScalarAsyncWithLogging(this._logger, CancellationToken.None);
+                    getLockedRowCountDurationMs = commandSw.ElapsedMilliseconds;
+                }
+            }
+            catch (Exception ex)
+            {
+                this._logger.LogError($"Failed to query count of lease locked changes for table '{this._userTable.FullName}' due to exception: {ex.GetType()}. Exception message: {ex.Message}");
+                TelemetryInstance.TrackException(TelemetryErrorName.GetLeaseLockedRowCount, ex, null, new Dictionary<TelemetryMeasureName, double>() { { TelemetryMeasureName.GetLockedRowCountDurationMs, getLockedRowCountDurationMs } });
+                leaseLockedRowsCount = -1;
```
Charles-Gagnon, line 860:
```suggestion
                // This is currently only used for debugging, so return a -1 instead of throwing since it isn't necessary to get the value
                leaseLockedRowsCount = -1;
```

Charles-Gagnon (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon: Oh - and to respond to this. That's fine since we'll still log if the rows returned is >= 0. And since we log the error it'll be visible in the logs regardless.

PR #904, opened 2023-08-09
replace read async with read of SqlReader for perf

### LINKED ISSUE #903: Investigate use of SqlClient ReadAsync
Turns out there's some known issues with using ReadAsync in SqlClient

https://github.com/dotnet/SqlClient/issues/593
https://github.com/dotnet/SqlClient/issues/44

We currently use ReadAsync in many places, so we should investigate whether we want to switch over to using Read as done here : https://github.com/microsoft/sqltoolsservice/pull/2161

MaddyDev:
Fixes #903 

To preemptively address the below issues, we followed the fix made in sqltools service (https://github.com/microsoft/sqltoolsservice/pull/2161)

https://github.com/dotnet/SqlClient/issues/593
https://github.com/dotnet/SqlClient/issues/44


Charles-Gagnon (approved):

Charles-Gagnon:
@cheenamalhotra This look good to you?

src/SqlAsyncEnumerable.cs
```diff
@@ -112,7 +112,7 @@ private async Task<bool> GetNextRowAsync()
                             this._reader = await command.ExecuteReaderAsync();
```
cheenamalhotra, line 112: Since ExecuteReaderAsync has issues with cancellation - I would recommend changing that to ExecuteReader too. Do you plan to do that separately?

cheenamalhotra (approved):

src/SqlAsyncEnumerable.cs
Charles-Gagnon: That goes for any Async calls from SqlClient, right? @MaddyDev Make sure you're doing a search all for anything with "Async" in its name so we aren't missing any others.

lucyzhang929 (approved):

src/SqlAsyncCollector.cs
```diff
@@ -599,7 +599,7 @@ public static async Task<TableInformation> RetrieveTableInformationAsync(SqlConn
                 {
                     string getPrimaryKeysQuery = GetPrimaryKeysQuery(table);
                     var cmd = new SqlCommand(getPrimaryKeysQuery, sqlConnection);
-                    using (SqlDataReader rdr = await cmd.ExecuteReaderAsyncWithLogging(logger, CancellationToken.None))
+                    using (SqlDataReader rdr = cmd.ExecuteReaderWithLogging(logger))
                     {
                         while (await rdr.ReadAsync())
```
Charles-Gagnon, line 604: What about this ReadAsync (and the one above)?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -297,8 +297,9 @@ private async Task GetTableChangesAsync(SqlConnection connection, CancellationTo
 
                             using (SqlDataReader reader = await getChangesCommand.ExecuteReaderAsync(token))
```
Charles-Gagnon, line 298: ExecuteReader?

src/SqlAsyncCollector.cs
```diff
@@ -199,11 +198,11 @@ private async Task UpsertRowsAsync(IList<T> rows, SqlAttribute attribute, IConfi
                     }
                 }
 
-                if (tableInfo == null)
+                if (!(cachedTables[cacheKey] is TableInformation tableInfo))
                 {
                     TelemetryInstance.TrackEvent(TelemetryEventName.TableInfoCacheMiss, props);
                     // set the columnNames for supporting T as JObject since it doesn't have columns in the member info.
-                    tableInfo = await TableInformation.RetrieveTableInformationAsync(connection, fullTableName, this._logger, GetColumnNamesFromItem(rows.First()), this._serverProperties);
+                    tableInfo = TableInformation.RetrieveTableInformationAsync(connection, fullTableName, this._logger, GetColumnNamesFromItem(rows.First()), this._serverProperties);
```
lucyzhang929, line 205: Should we also remove the Async from the method names? `RetrieveTableInformation`

Charles-Gagnon: +1 - for any methods that no longer have await in them we should clean those up

chlafreniere:
How will we measure impact after this change?

Charles-Gagnon:
That would seem to fit nicely into the work we already have planned for perf baselines. @MaddyDev maybe you hold off on getting this in for now (since it's not actively blocking anyone AFAIK) and instead focus on getting the perf stuff figured out first so we can both test the usefulness of that along with validating that this is actually making an improvement?

GeeSuth:
it's that mean the #903 will not fixed?

PR #906, opened 2023-08-16
[Trigger] Sync with main

MaddyDev:
Merge commits from main


chlafreniere (approved):

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -579,7 +579,7 @@ public void UnsupportedDatabaseThrows(SupportedLanguages lang)
         /// <summary>
         /// Tests that when a user function throws an exception we'll retry executing that function once the lease timeout expires
         /// </summary>
-        [RetryTheory]
+        [Theory]
```
Charles-Gagnon, line 582: Are you going to bring this back?

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -591,7 +591,7 @@ void TestExceptionMessageSeen(object sender, DataReceivedEventArgs e)
             {
                 if (e.Data.Contains(TriggerWithException.ExceptionMessage))
                 {
-                    taskCompletionSource.SetResult();
+                    taskCompletionSource.TrySetResult();
```
Charles-Gagnon, line 594: Do you know why this fixed the issue? (and what was the issue anyways?)

MaddyDev: I kept seeing the test failure with the error similar to "InvalidOperationException: The underlying Task is already set to final state" and when I looked up online, I learned this happens when the task is already set and if you're trying to set it again - it throws the above error. And alternatively suggested to use TrySetResult which wouldn't throw the error (in cases when more than one operation can set the result).

test/Integration/SqlTriggerBindingIntegrationTests.cs
MaddyDev: oh yeah I forgot to revert this, thanks!

test/Integration/SqlTriggerBindingIntegrationTests.cs
Charles-Gagnon: Why would we be setting the result more than once in the first place? Shouldn't the exception message only show up once?

MaddyDev: That I was not able to figure out since this error only happens on the pipeline builds and could never repro on my dev.

Charles-Gagnon (approved):

PR #910, opened 2023-08-23
Update to Latest Version of Signing Task

chlafreniere:
Update to latest version of signing task, which supports newer versions of .NET, negating the need to use older versions of .NET Core solely for signing purposes.


MaddyDev:
These should be ready now? (after resolving the merge conflicts)

Charles-Gagnon (approved):

PR #911, opened 2023-08-25
Update GeneralSetup.md

Aenas11:
Adds VIEW CHANGE TRACKING to sql command example


Charles-Gagnon:
This has already been added in the release branch docs

https://github.com/Azure/azure-functions-sql-extension/blob/release/trigger/docs/GeneralSetup.md#trigger-permissions

Issue #810, opened 2023-04-18
Add back in some diagnostic logging


Charles-Gagnon:
I recently removed most of our debug logging, but it was brought up that we likely will want to have some amount of logging to indicate that the function is running smoothly for supportability purposes. 

EventHubs can be used as a sample for what kind of logging is useful to have : https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs

We can also add in a config variable that - when enabled - will log much more verbose information if needed.

MaddyDev:
We should look into SqlClient verbose logging and follow that pattern.

Charles-Gagnon:
See https://github.com/microsoft/sqltoolsservice/pull/1544 for how I did it in STS

PR #914, opened 2023-08-30
add SqlClientListener for diagnostic logging 

MaddyDev:
To enable the sql client logging, user needs to add the VerboseLogging and set it to true in config settings.
(Followed the implementation in sql tools service https://github.com/microsoft/sqltoolsservice/pull/1544)

For #810

### DIFF
--- a/.github/label-actions.yml
+++ b/.github/label-actions.yml
@@ -27,7 +27,7 @@ Needs Connection Logs:
   comment: |
     We need more info to debug your particular issue. If you could attach your SQL Client logs to the issue, it would help us fix the issue much faster.
 
-    1. Set the AzureFunctions_SqlBindings_VerboseLogging to `true` in your `*.settings.json` file.
+    1. Set the AzureFunctions_SqlBindings_VerboseLogging to `true` in your `*.settings.json` file or your Function App settings.
 
     ```json
     {
@@ -41,14 +41,14 @@ Needs Connection Logs:
     }
     ```
 
-    2. Set the default logging level to `Debug` and enable logging to a file. To do this ensure the following entries are in your `host.json`. ([instructions](https://learn.microsoft.com/azure/azure-functions/configure-monitoring?tabs=v2#configure-log-levels))
+    2. Set the default logging level to `Trace` and enable logging to a file. To do this ensure the following entries are in your `host.json`. ([instructions](https://learn.microsoft.com/azure/azure-functions/configure-monitoring?tabs=v2#configure-log-levels))
 
     ```json
     {
       "logging": {
         "fileLoggingMode": "always",
         "logLevel": {
-          "default": "Debug"
+          "default": "Trace"
         }
       }
     }
--- a/.github/label-actions.yml
+++ b/.github/label-actions.yml
@@ -2,6 +2,7 @@
 Needs Logs:
   comment: |
     We need more info to debug your particular issue. If you could attach your logs to the issue, it would help us fix the issue much faster.
+    Note that these instructions are for local development only.
 
     1. Set the default logging level to `Debug` and enable logging to a file. To do this ensure the following entries are in your `host.json`. ([instructions](https://learn.microsoft.com/azure/azure-functions/configure-monitoring?tabs=v2#configure-log-levels))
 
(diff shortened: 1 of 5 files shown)


src/Telemetry/Telemetry.cs
```diff
@@ -122,6 +122,26 @@ private void InitializeTelemetry(string productVersion, IConfiguration config)
...
+        {
+            try
+            {
+                if (!this._initialized || !this.Enabled || this._client is null)
+                {
+                    return;
+                }
+                this._logger.LogTrace($"Sending event {eventName}");
+
+                this._client.TrackEvent($"{EventsNamespace}/{eventName}", properties, null);
```
Charles-Gagnon, line 135: You should use TrackEvent here (see other Track* methods)

MaddyDev: yes but the TrackEvent accepts *TelemetryEventName* and IDictionary<*TelemetryPropertyName*, string> properties and the data we get from SqlClientListener cannot be predefined and sent - the reason why I used the TelemetryClient's TrackEvent.

Charles-Gagnon: Oh wait, so first big question is why are you even sending this via telemetry? We shouldn't be doing that. We just want to log these events locally.

src/Telemetry/Telemetry.cs
```diff
@@ -122,6 +122,23 @@ private void InitializeTelemetry(string productVersion, IConfiguration config)
             }
         }
 
+        public void TrackSQLClientEvent(string eventName, IDictionary<string, string> properties = null)
```
Charles-Gagnon, line 125: This shouldn't go in the telemetry class at all. Just have the event listener handle doing the logging.

.github/label-actions.yml
```diff
@@ -20,4 +20,42 @@ Needs Logs:
...
     3. Get the log file and attach it to this issue. By default this will be in `%TMP%/LogFiles/Application/Functions/Host`.
 
+    **NOTE** Debug logging will include information such as Database and table names, if you do not wish to include this information you can either redact it from the logs before attaching or let us know and we will provide a way to send logs directly to us.
+
+#actions for Needs Connection Logs
+Needs Connection Logs:
+  comment: |
+    We need more info to debug your particular issue. If you could attach your SQL Client logs to the issue, it would help us fix the issue much faster.
+
+    1. Set the AzureFunctions_SqlBindings_VerboseLogging to `true` in your `*.settings.json` file.
```
Charles-Gagnon, line 30:
```suggestion
    1. Set the AzureFunctions_SqlBindings_VerboseLogging to `true` in your `*.settings.json` file or your Function App settings.
```

src/SqlClientEventListener.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    if (payload != null)
+                    {
+                        this._logger.LogTrace($"Sending event {eventData.EventName}. Properties: {Utils.JsonSerializeObject(payloadDictionary)}");
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                this._logger.LogError($"Error sending event {eventData.EventName}. Message={ex.Message}");
```
Charles-Gagnon, line 66:
```suggestion
                this._logger.LogError($"Error logging SqlClient event {eventData.EventName}. Message={ex.Message}");
```

.github/label-actions.yml
```diff
@@ -20,4 +20,42 @@ Needs Logs:
...
+        "logLevel": {
+          "default": "Trace"
+        }
+      }
+    }
+    ```
+
+    3. Restart your function and reproduce your issue
+
+    4. Get the log file and attach it to this issue. By default this will be in `%TMP%/LogFiles/Application/Functions/Host`.
```
Charles-Gagnon, line 59: Actually, looking at this it would be good to include instructions (for both this and the other logging message) for doing this in a deployed function too since they won't be the same. Not sure offhand what the easiest way to export the logs would be, so that might take some investigation.

In the meantime I think we can actually revert the change I had you add earlier about the Function App settings and instead specify that these instructions are for local development only. Then when we figure out a way for deployed functions we can come back and update both with those instructions.

Charles-Gagnon (approved):

.github/label-actions.yml
Charles-Gagnon: Thanks for adding the note! Might be worth making an actual issue to track updating this with instructions for deployed functions as well. 

(although if we do that it might actually be best to just update the instructions in our docs and then link to those from here so we don't have to remember to keep both in sync)

Charles-Gagnon (approved):

MaddyDev:
added #917 for us to track the hosted scenario docs.

PR #916, opened 2023-09-01
Update links on all docs to point to main

MaddyDev:
replaced release/trigger with main.
I'll need update the templates before replacing the Preview everywhere.

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -4,6 +4,7 @@ pr:
   branches:
     include:
     - main
+    - release/*
 
 variables:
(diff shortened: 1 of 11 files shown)


builds/azure-pipelines/build-pr.yml
```diff
@@ -4,7 +4,6 @@ pr:
   branches:
     include:
     - main
-    - release/*
```
Charles-Gagnon, line 7: Let's actually keep this, it's possible we'll want to have release branches in the future and keeping this doesn't really hurt anything.

Charles-Gagnon (approved):

PR #941, opened 2023-10-06
add baselines json

MaddyDev:
Will add instructions to the on-call wiki for the on-call person to keep tabs on the perf run times.

### DIFF
--- a/performance/PerformanceBaselines.json
+++ b/performance/PerformanceBaselines.json
@@ -1,0 +1,79 @@
+{
+    "PerformanceRuntime": {
+        "InputBinding": {
+            "TestName": "GetProductsTest",
+            "RunTimeAverage": 2.081
+        },
+        "OutputBinding": {
+            "TestName": "AddProductsArrayTest",
+            "RunTimeAverage": 2.055
+        },
+        "TriggerBinding": [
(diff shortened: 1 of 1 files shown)


lucyzhang929 (approved):

performance/PerformanceBaselines.json
```diff
@@ -0,0 +1,79 @@
+{
+    "PerformanceRuntime": {
+        "InputBinding": {
+            "TestName": "GetProductsTest",
+            "RunTimeAverage": 2.081
```
Charles-Gagnon, line 5: Please add units to these property names (e.g. `RunTimeAverageSec`)

Charles-Gagnon (approved):

MaddyDev:
Added the [monitor perf pipeline](https://msdata.visualstudio.com/SQLToolsAndLibraries/_wiki/wikis/SQLToolsAndLibraries.wiki/52049/On-Call-Duties?anchor=monitor-performance-pipeline) section in on-call wiki, I've added 100 sec ~> 1.5 min as the threshold there for us to investigate the deviation, we can discuss what's a good number and update it there in our on-call handoff sync.

Issue #920, opened 2023-09-12
Failure to start SQL trigger on .NET 7


Stxtics:
Due to project dependencies, I'm working with a function app that targets .NET 7 on an isolated process on Azure functions v4. 

On Friday, I started using this package as I wanted to make use of the SQL Triggers functionality. I had it working with a table for an initial test but on Monday I made some changes to use a new table. Since that, I have been dealing with an issue that I can't seem to figure out. So far the only fix I've found is switching to target .NET 6 (which requires commenting out a bunch of existing code in the project due to project dependencies). Unless I'm missing something I thought that .NET 7 is supported with this package, even though it targets .NET 6. 

Here is the exception output:

Failed to start SQL trigger listener for table: '', function ID: ''. Exception: System.TypeInitializationException: The type initializer for 'Microsoft.Data.SqlClient.SqlAuthenticationProviderManager' threw an exception.
System.IO.FileNotFoundException: Could not load file or assembly 'System.Threading, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'System.Threading, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'For detailed output, run func with --verbose flag.
at System.Configuration.ConfigurationManager.EnsureConfigurationSystem()
at System.Configuration.ConfigurationManager.PrepareConfigSystem()
at System.Configuration.ConfigurationManager.GetSection(String sectionName)
at Microsoft.Data.SqlClient.SqlAuthenticationProviderManager.FetchConfigurationSection[T](String name)
at Microsoft.Data.SqlClient.SqlAuthenticationProviderManager..cctor()
--- End of inner exception stack trace ---
at Microsoft.Data.SqlClient.SqlInternalConnectionTds..ctor(DbConnectionPoolIdentity identity, SqlConnectionString connectionOptions, SqlCredential credential, Object providerInfo, String newPassword, SecureString newSecurePassword, Boolean redirectedUserInstance, SqlConnectionString userConnectionOptions, SessionData reconnectSessionData, Boolean applyTransientFaultHandling, String accessToken, DbConnectionPool pool)
[...]

PR #942, opened 2023-10-09
downgrade package bumps

MaddyDev:
Following out discussion with AF team, downgrading the major package bumps back.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,8 +9,8 @@
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.0.1" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
     <PackageVersion Include="morelinq" Version="3.4.2" />
-    <PackageVersion Include="System.Drawing.Common" Version="7.0.0" />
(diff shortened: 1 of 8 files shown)


Directory.Packages.props
```diff
@@ -27,7 +27,7 @@
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Timer" Version="4.2.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Storage.Queues" Version="5.1.2" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.5" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="6.0.1" />
```
Charles-Gagnon, line 30: Did you go through and check all the transitive dependencies as well?

Also - while you're here could you add a comment to the SqlClient package entry explaining that we can't bump beyond 5.0.1 right now and link to https://github.com/Azure/azure-functions-sql-extension/issues/920?

Charles-Gagnon (approved):
Approving assuming that transitive dependency versions have been verified

PR #947, opened 2023-10-12
Update java library version to 2.1.0

lucyzhang929:
Fixes: https://github.com/Azure/azure-functions-sql-extension/issues/901

### DIFF
--- a/java-library/pom.xml
+++ b/java-library/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>com.microsoft.azure.functions</groupId>
   <artifactId>azure-functions-java-library-sql</artifactId>
-  <version>2.1.0-preview</version> <!-- Update the dependency version in samples-java/pom.xml and test-java/pom.xml if this version is updated. -->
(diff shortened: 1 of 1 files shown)


java-library/pom.xml
```diff
@@ -4,7 +4,7 @@
 
   <groupId>com.microsoft.azure.functions</groupId>
   <artifactId>azure-functions-java-library-sql</artifactId>
-  <version>2.1.0-preview</version> <!-- Update the dependency version in samples-java/pom.xml and test-java/pom.xml if this version is updated. -->
+  <version>2.1.0</version> <!-- Update the dependency version in samples-java/pom.xml and test-java/pom.xml if this version is updated. -->
```
lucyzhang929, line 7: Will update version in samples once the library is released.

MaddyDev (approved):

java-library/pom.xml
Charles-Gagnon: Do we have an issue for tracking this?

lucyzhang929: Created issue: https://github.com/Azure/azure-functions-sql-extension/issues/948 to track.

PR #949, opened 2023-10-13
Fix nuget packaging during build

Charles-Gagnon:
We've been getting these warnings in the pipelines lately : 

`##[warning]/var/folders/3s/vfzpb5r51gs6y328rmlgzm7c0000gn/T/b3fgdwow.a3k/WorkerExtensions.csproj(0,0): Warning NU1603: Microsoft.Azure.Functions.Worker.Extensions depends on Microsoft.Azure.WebJobs.Extensions.Sql (>= 3.0.447-preview) but Microsoft.Azure.WebJobs.Extensions.Sql 3.0.447-preview was not found. An approximate best match of Microsoft.Azure.WebJobs.Extensions.Sql 3.0.447 was resolved.`

Looking into it I found a couple of issues that are fixed in this PR

1. We weren't passing in the nuget version to the dotnet pack target so it was falling back to the binary version (which doesn't have -preview for preview builds)
2. We weren't passing in the configuration to the dotnet pack target so it was generating from the wrong assemblies (this was for the test package only, the release packaging was done correctly)
3. We were generating the nuget package when building for release - even though we don't want to at that point (since we have to sign the binaries before packaging)
4. The dotnet test steps were rebuilding the projects again unnecessarily so I'm making them not build to speed up the pipeline. Doing this then required moving the build step below where we build the samples so that the non .NET samples would be copied over for the tests correctly

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -46,4 +46,4 @@ stages:
         configuration: '$(configuration)'
         nugetVersion: '99.99.99-test'
         binariesVersion: '99.99.99'
-        oopWorkerSupportedExtensionVersion: '99.99.99'
+        oopWorkerSupportedExtensionVersion: '99.99.99-test'
--- a/samples/samples-outofproc/Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.csproj
+++ b/samples/samples-outofproc/Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.csproj
@@ -19,8 +19,10 @@
       Include="..\..\Worker.Extensions.Sql\src\Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj" />
     <!-- We want to build the extension project first since this project references the package
     produced by building that. But we don't want it as an actual runtime depdendency -->
-    <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Extensions.Sql.csproj"
-      ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Extensions.Sql.csproj">
+        <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+        <Private>false</Private>
+    </ProjectReference>
   </ItemGroup>
   <ItemGroup>
     <None Update="host.json">
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -91,7 +91,7 @@ steps:
   inputs:
     command: build
     projects: ${{ parameters.solution }}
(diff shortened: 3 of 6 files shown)


chlafreniere:
(holding off reviewing until checks pass)

Charles-Gagnon:
Note - I plan on following up with fixing the "project cannot be packaged" warnings after this one goes in.

Charles-Gagnon:
Converting to draft while I figure out all the failures

Charles-Gagnon:
Ok, figured out the problems. Previously since we were building the projects multiple times (once in the build step + once for each test) then it would run the "CopySamples" target multiple times as well. The first time it runs during the build step it did this before the samples projects were actually built - so the samples weren't actually copied over (the .NET ones at least, since we specifically build the non-.NET ones earlier). 

Previously this would then be "fixed" during the subsequent builds, since it would then have the built samples from the previous build step and copy those over at that point. But with my change to not build during tests this never happened.

The fix for that was to reference the .NET samples projects in the test project. But then this led to another weird issue where the tests were trying to start up the function host with dotnet-isolated instead of just dotnet for the CSharp integration tests. That was because the ProjectReference was resulting in the metadata files (such as local.settings.json) being copied over as well (from items such as https://github.com/Azure/azure-functions-sql-extension/blob/main/test-outofproc/test-outofproc.csproj#L31). This resulted in us overwriting the settings files from the integration test itself (which was set as `dotnet`) To fix this we need to add `Private=false` to the ProjectReference to cause it to not copy over any output content. See https://github.com/dotnet/msbuild/issues/7986 for more information on this very-badly-named property and how it relates to ReferenceOutputAssembly.

builds/azure-pipelines/build-pr.yml
```diff
@@ -46,4 +46,4 @@ stages:
         configuration: '$(configuration)'
         nugetVersion: '99.99.99-test'
         binariesVersion: '99.99.99'
-        oopWorkerSupportedExtensionVersion: '99.99.99'
\ No newline at end of file
+        oopWorkerSupportedExtensionVersion: '99.99.99-test'
```
Charles-Gagnon, line 49: The worker extension version is the nuget package version, not the binary version.

samples/samples-outofproc/Microsoft.Azure.WebJobs.Extensions.Sql.SamplesOutOfProc.csproj
```diff
@@ -19,8 +19,10 @@
       Include="..\..\Worker.Extensions.Sql\src\Microsoft.Azure.Functions.Worker.Extensions.Sql.csproj" />
     <!-- We want to build the extension project first since this project references the package
     produced by building that. But we don't want it as an actual runtime depdendency -->
-    <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Extensions.Sql.csproj"
-      ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Extensions.Sql.csproj">
+        <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+        <Private>false</Private>
```
Charles-Gagnon, line 24: Technically we shouldn't need to do this since the extensions project doesn't generate other output files, but adding it doesn't hurt and will prevent issues later on if we do add something.

lucyzhang929 (approved):

PR #953, opened 2023-10-16
Don't use GeneratePackageOnBuild & sign worker package

Charles-Gagnon:
By setting this to true at the global level (including as a parameter to dotnet build of the sln) we ended up with these warnings : 

`This project cannot be packaged because packaging has been disabled. Add <IsPackable>true</IsPackable> to the project file to enable producing a package from this project.`

I had added this in the previous PR, but after looking at it realized there doesn't seem to be any reason why we need two separate build steps. We can use the same binaries for tests & release, so we can start off with just building everything once. This will include the packages now (which will contain unsigned binaries) but that's fine since we'll repackage at the end.

So this allowed me to just remove the GeneratePackageOnBuild props, which meant we're only using that in one place (the worker project) and thus the warnings go away.

While I was looking into this I noticed that we were never signing the Worker assembly, so I'm fixing that as well.

### DIFF
--- a/builds/azure-pipelines/build-pr.yml
+++ b/builds/azure-pipelines/build-pr.yml
@@ -46,4 +46,4 @@ stages:
         configuration: '$(configuration)'
         nugetVersion: '99.99.99-test'
         binariesVersion: '99.99.99'
-        oopWorkerSupportedExtensionVersion: '99.99.99'
+        oopWorkerSupportedExtensionVersion: '99.99.99-test'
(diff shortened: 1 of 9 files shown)


Charles-Gagnon:
With this there's:

1. A couple serialization warnings I'll fix in a follow up PR
2. Occasional warnings from the Roslyn analyzers. I don't believe this is anything we can do anything about, but I'll look and make sure we're running things correctly. 

And then we're back to clean builds!

lucyzhang929 (approved):

PR #955, opened 2023-10-19
Bump Azure.Identity to v1.10.2

lucyzhang929:


Worker.Extensions.Sql/src/packages.lock.json
```diff
@@ -199,8 +201,11 @@
       },
       "System.Diagnostics.DiagnosticSource": {
         "type": "Transitive",
-        "resolved": "5.0.0",
-        "contentHash": "tCQTzPsGZh/A9LhhA6zrqCRV4hOHsK90/G7q3Khxmn6tnB1PuNU0cRaKANP2AWcF9bn0zsuOoZOSrHuJk6oNBA=="
+        "resolved": "6.0.1",
```
Charles-Gagnon, line 204: We can't do this, right?

Worker.Extensions.Sql/src/packages.lock.json
```diff
@@ -238,6 +243,15 @@
           "System.Threading.Tasks": "4.3.0"
         }
       },
+      "System.IO.FileSystem.AccessControl": {
```
Charles-Gagnon, line 246: Hm, so this whole versioning thing makes me wonder if we have to also be checking new dependencies...because if we're adding new references like this that don't match what is currently in the bundle couldn't that also cause problems?

@lucyzhang929 @MaddyDev Could you follow up with checking this with the AF Host team?

Worker.Extensions.Sql/src/packages.lock.json
lucyzhang929: The minor bump of Azure.Core from 1.31 to 1.35 bumped this major version. I'll reach out to the Func host team to see what we can do here.

Worker.Extensions.Sql/src/packages.lock.json
lucyzhang929: Will reach out to the AF host team.

Worker.Extensions.Sql/src/packages.lock.json
lucyzhang929: Confirmed with Fabio that this bump is okay.

Charles-Gagnon (approved):

MaddyDev (approved):

PR #960, opened 2023-10-30
Add SqlOptions for reading extra configuration

MaddyDev:
Reading our optional configuration values from app settings is not the recommended way and instead using options to read from host.json is. In this PR, I added the SqlOptions to read our [optional config values for Sql Trigger](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-trigger?tabs=isolated-process%2Cportal&pivots=programming-language-csharp#optional-configuration) from host.json to follow the pattern suggested [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-host-json#sample-hostjson-file)

I did not remove the current way of reading from app settings since that would be a breaking change and would warrant a vbump. We can at a later point remove it when we vbump (#961).

Tests - I've added a lot of plumbing code copied over from azure storage sdk for the end to end test. I would revisit and clean that up (removing the unused code) but since we're time bound for GA, I have copied over the entire classes that are needed for set up. (#962)


src/Common/SqlOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// <summary>
+    /// Represents configuration for <see cref="SqlTriggerAttribute"/>.
+    /// </summary>
+    public class SqlOptions : IOptionsFormatter
+    {
+        public const int DefaultBatchSize = 100;
+        public const int DefaultPollingIntervalMs = 1000;
+        private const int DefaultMinimumPollingIntervalMs = 100;
+
+        // NOTE: please ensure the Readme file and other public documentation are also updated if this value ever
```
lucyzhang929, line 21: This comment would apply to DefaultBatchSize and DefaultPollingIntervalMs too right?

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -124,21 +124,23 @@ internal sealed class SqlTableChangeMonitor<T> : IDisposable
...
             this._primaryKeyColumns = primaryKeyColumns ?? throw new ArgumentNullException(nameof(primaryKeyColumns));
+            this._sqlOptions = sqlOptions ?? throw new ArgumentNullException(nameof(sqlOptions));
             this._executor = executor ?? throw new ArgumentNullException(nameof(executor));
             this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
             this._userTableId = userTableId;
             this._telemetryProps = telemetryProps ?? new Dictionary<TelemetryPropertyName, string>();
 
             // Check if there's config settings to override the default max batch size/polling interval values
+            // TODO: remove reading from settings when we decide to move to reading them from func.json.
```
lucyzhang929, line 135: Did you mean host.json here?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -64,20 +60,23 @@ internal sealed class SqlTriggerListener<T> : IListener, IScaleMonitorProvider,
...
             this._connectionString = !string.IsNullOrEmpty(connectionString) ? connectionString : throw new ArgumentNullException(nameof(connectionString));
             this._userTable = !string.IsNullOrEmpty(tableName) ? new SqlObject(tableName) : throw new ArgumentNullException(nameof(tableName));
             this._userDefinedLeasesTableName = userDefinedLeasesTableName;
             this._userFunctionId = !string.IsNullOrEmpty(userFunctionId) ? userFunctionId : throw new ArgumentNullException(nameof(userFunctionId));
             this._executor = executor ?? throw new ArgumentNullException(nameof(executor));
+            this._sqlOptions = sqlOptions ?? throw new ArgumentNullException(nameof(sqlOptions));
             this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
             this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             int? configuredMaxChangesPerWorker;
+            // TODO: remove reading from settings when we decide to move to reading them from func.json.
```
lucyzhang929, line 77: host.json?

src/Common/SqlOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        public SqlOptions()
+        {
+
+        }
+
+        /// <summary>
+        /// Gets or sets the number of changes per batch to retrieve from the server.
+        /// The default is 100.
+        /// </summary>
+        public int BatchSize
```
lucyzhang929, line 47: We should be using MaxBatchSize (https://github.com/Azure/azure-functions-sql-extension/pull/668).

Charles-Gagnon:
@MaddyDev Would suggest you create follow-up issues for the two things you called out in the description (and then link them there too)

test/Integration/SqlOptionsEndToEndTests.cs
```diff
@@ -0,0 +1,32 @@
...
+                { $"{extensionPath}:PollingIntervalMs", "1000" },
+                { $"{extensionPath}:MaxChangesPerWorker", "100" },
+            };
+
+            SqlOptions options = TestHelpers.GetConfiguredOptions<SqlOptions>(b =>
+            {
+                b.AddSql();
+            }, values);
+
+            Assert.Equal(30, options.BatchSize);
```
Charles-Gagnon, line 27: It'd be good to add a test showing that when both host.json and app settings are set it will pick the app settings one.

src/TriggerBinding/SqlTableChangeMonitor.cs
Charles-Gagnon, line 135: I'd also suggest making it clear in our docs (both xml and docs) that if both are specified the app settings values will be chosen.

MaddyDev:
Added #961 and #962 as follow up issues.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -124,21 +124,23 @@ internal sealed class SqlTableChangeMonitor<T> : IDisposable
...
             this._userTableColumns = userTableColumns ?? throw new ArgumentNullException(nameof(userTableColumns));
             this._primaryKeyColumns = primaryKeyColumns ?? throw new ArgumentNullException(nameof(primaryKeyColumns));
+            this._sqlOptions = sqlOptions ?? throw new ArgumentNullException(nameof(sqlOptions));
             this._executor = executor ?? throw new ArgumentNullException(nameof(executor));
             this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
             this._userTableId = userTableId;
             this._telemetryProps = telemetryProps ?? new Dictionary<TelemetryPropertyName, string>();
 
+            // TODO: when we move to reading them exclusively from the host options, remove reading from settings.
```
Charles-Gagnon, line 134: nit: Link issue for tracking this

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -64,20 +60,23 @@ internal sealed class SqlTriggerListener<T> : IListener, IScaleMonitorProvider,
...
             this._connectionString = !string.IsNullOrEmpty(connectionString) ? connectionString : throw new ArgumentNullException(nameof(connectionString));
             this._userTable = !string.IsNullOrEmpty(tableName) ? new SqlObject(tableName) : throw new ArgumentNullException(nameof(tableName));
             this._userDefinedLeasesTableName = userDefinedLeasesTableName;
             this._userFunctionId = !string.IsNullOrEmpty(userFunctionId) ? userFunctionId : throw new ArgumentNullException(nameof(userFunctionId));
             this._executor = executor ?? throw new ArgumentNullException(nameof(executor));
+            this._sqlOptions = sqlOptions ?? throw new ArgumentNullException(nameof(sqlOptions));
             this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
             this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             int? configuredMaxChangesPerWorker;
+            // TODO: when we move to reading them exclusively from the host options, remove reading from settings.
```
Charles-Gagnon, line 77: nit: Link tracking issue

Charles-Gagnon (approved):

lucyzhang929 (approved):

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -114,6 +113,7 @@ internal sealed class SqlTableChangeMonitor<T> : IDisposable
             IReadOnlyList<string> userTableColumns,
             IReadOnlyList<(string name, string type)> primaryKeyColumns,
             ITriggeredFunctionExecutor executor,
+            SqlOptions sqlOptions,
```
mathewc, line 116: Here and below in all types you're taking a SqlOptions directly - make sure you're doing that only for types that aren't being created by DI. You want to use `IOptions<SqlOptions>` generally for DI created instances which ensures that all config is applied to the options

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -44,15 +45,17 @@ internal sealed class SqlTriggerBinding<T> : ITriggerBinding
         /// <param name="leasesTableName">Optional - Name of the leases table</param>
         /// <param name="parameter">Trigger binding parameter information</param>
         /// <param name="hostIdProvider">Provider of unique host identifier</param>
+        /// <param name="sqlOptions"></param>
         /// <param name="logger">Facilitates logging of messages</param>
         /// <param name="configuration">Provides configuration values</param>
-        public SqlTriggerBinding(string connectionString, string tableName, string leasesTableName, ParameterInfo parameter, IHostIdProvider hostIdProvider, ILogger logger, IConfiguration configuration)
+        public SqlTriggerBinding(string connectionString, string tableName, string leasesTableName, ParameterInfo parameter, SqlOptions sqlOptions, IHostIdProvider hostIdProvider, ILogger logger, IConfiguration configuration)
```
mathewc, line 51: Same comment here and elsewhere regarding `IOptions<SqlOptions>`

src/TriggerBinding/SqlTriggerBindingProvider.cs
```diff
@@ -22,18 +22,20 @@ internal sealed class SqlTriggerBindingProvider : ITriggerBindingProvider
...
         /// <param name="hostIdProvider">Provider of unique host identifier</param>
         /// <param name="loggerFactory">Used to create logger instance</param>
-        public SqlTriggerBindingProvider(IConfiguration configuration, IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory)
+        /// <param name="sqlOptions"></param>
+        public SqlTriggerBindingProvider(IConfiguration configuration, IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, Microsoft.Extensions.Options.IOptions<SqlOptions> sqlOptions)
         {
             this._configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             this._hostIdProvider = hostIdProvider ?? throw new ArgumentNullException(nameof(hostIdProvider));
-
+            this._sqlOptions = sqlOptions?.Value ?? new SqlOptions();
```
mathewc, line 38: I don't think you should ever be newing up the options yourself like this - you always want options to be hydrated by the configuration system

src/TriggerBinding/SqlTriggerListener.cs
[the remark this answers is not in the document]
MaddyDev: We'll be removing reading from config at some point (so not going to add more code w.r.t to that). For now, I did add a test for checking precedence (and reading from app settings takes precedence).

src/TriggerBinding/SqlTriggerBinding.cs
MaddyDev: Updated this to use IOptions  here and other places you commented are not created using DI but use the options from there, so we are good. Thanks!

PR #963, opened 2023-11-03
Add scaler api for consumption plan support

AmeyaRele:
**Description**
This PR adds scaler apis for supporting SQL extension on Azure Functions Consumption Plan. The TargetScaler and ScaleMonitor would be called from scale controller by including the sql package.

**Task List**

- [x] Create SqlScalerProvider class
- [x] Add extension method 'AddSqlScaleForTrigger'
- [x] Add e2e test similar to [cosmosdb](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/a390e4eee0f7bb5d2add08e12db7368b98163ea7/test/WebJobs.Extensions.CosmosDB.Tests/ScaleHostEndToEndTests.cs#L74)

**Reference PRs**

- [ServiceBus](https://github.com/Azure/azure-sdk-for-net/pull/34557)
- [CosmosDB](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/841)
- [EventHub](https://github.com/Azure/azure-sdk-for-net/pull/35344)


Charles-Gagnon:
@AmeyaRele We've already started the release process for our GA Trigger release - is this something that should be included as a part of that? Or is it ok to follow up with later on?

Also, is there any documentation you could link to with more context about what this feature is as well?

AmeyaRele:
@Charles-Gagnon we can follow up with this later on. It isn't a criteria for GA of trigger afaik.

I don't believe there is public documentation on this, but essentially the scale component is redesigned to use these methods by including the extension as a package and use reflection to call the method in the scale component. This is required in consumption(scale to zero) plan for azure functions as the scale component is not guaranteed to have a worker running which it can query for the scaling decisions.

src/SqlBindingExtension.cs
```diff
@@ -28,5 +29,22 @@ public static IWebJobsBuilder AddSql(this IWebJobsBuilder builder, Action<SqlOpt
 
             return builder;
         }
+
+        internal static IWebJobsBuilder AddSqlScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
+        {
+            IServiceProvider serviceProvider = null;
+            var scalerProvider = new Lazy<SqlScalerProvider>(() => new SqlScalerProvider(serviceProvider, triggerMetadata));
+            builder.Services.AddSingleton((Func<IServiceProvider, IScaleMonitorProvider>)delegate (IServiceProvider resolvedServiceProvider)
+            {
```
pragnagopa, line 38: is incremental scaling already used in public preview? if not is it possible to block use of incremental scaling  - @alrod ?

chlafreniere:
Hi @AmeyaRele, just curious how much more work is needed in this PR. Is it just the tests, or is there other known work that needs to go in product code-wise?

src/TriggerBinding/SqlScalerProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+        public SqlScalerProvider(IServiceProvider serviceProvider, TriggerMetadata triggerMetadata)
+        {
+            IConfiguration config = serviceProvider.GetService<IConfiguration>();
+            ILoggerFactory loggerFactory = serviceProvider.GetService<ILoggerFactory>();
+            ILogger logger = loggerFactory.CreateLogger(LogCategories.CreateTriggerCategory("Sql"));
+            SqlMetaData sqlMetadata = JsonConvert.DeserializeObject<SqlMetaData>(triggerMetadata.Metadata.ToString());
+            sqlMetadata.ResolveProperties(serviceProvider.GetService<INameResolver>());
+            var userTable = new SqlObject(sqlMetadata.TableName);
+            string connectionString = SqlBindingUtilities.GetConnectionString(sqlMetadata.ConnectionStringSetting, config);
+            int maxChangesPerWorker = config.GetValue(SqlTriggerConstants.ConfigKey_SqlTrigger_MaxChangesPerWorker, SqlOptions.DefaultMaxChangesPerWorker);
```
Charles-Gagnon, line 35: You need to be reading from the SqlOptions we added a little while back too - e.g. https://github.com/Azure/azure-functions-sql-extension/blob/main/src/TriggerBinding/SqlTriggerListener.cs#L79

PR: https://github.com/Azure/azure-functions-sql-extension/pull/960

src/TriggerBinding/SqlScalerProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+            string userDefinedLeasesTableName = sqlMetadata.LeasesTableName;
+            string userFunctionId = (string)triggerMetadata.Properties["userFunctionId"];
```
Charles-Gagnon, line 37: Are we guaranteed that this is going to have a value? Is there a reason you're not able to use the same method we use for getting the function ID elsewhere?

https://github.com/Azure/azure-functions-sql-extension/blob/main/src/TriggerBinding/SqlTriggerBinding.cs#L105

src/TriggerBinding/SqlScalerProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+            public string TableName { get; set; }
+
+            [JsonProperty]
+            public string LeasesTableName { get; set; }
+
+            public void ResolveProperties(INameResolver resolver)
+            {
+                if (resolver != null)
+                {
+                    this.TableName = resolver.ResolveWholeString(this.TableName);
```
Charles-Gagnon, line 68: Why do we need to resolve the user & leases table name?

Charles-Gagnon (changes requested):
Are you going to be adding tests as well? Especially integration tests, is that something that we're going to be able to do?

src/SqlBindingExtension.cs
alrod: Yes, it looks like we can ger rid of incremental scaling(IScaleMonitorProvider) here as we have never released it.

src/TriggerBinding/SqlScalerProvider.cs
alrod: Why don't we use `triggerMetadata.FunctionName` here?

MaddyDev: @Charles-Gagnon we can't really reuse the function ID from the SqlTriggerBinding, can I replace it with  `triggerMetadata.FunctionName` mentioned above?

Charles-Gagnon: Why aren't we able to use the same method? Do we not have access to IHostIdProvider ?

If not, then I believe we're still going to need to find another since we still will need to get the function ID somehow in order for the metrics provider to be able to get the list of unprocessed changes : https://github.com/Azure/azure-functions-sql-extension/blob/main/src/TriggerBinding/SqlTriggerMetricsProvider.cs#L103

src/SqlBindingExtension.cs
AmeyaRele: It is in the present in the public docs https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-trigger?tabs=isolated-process%2Cportal&pivots=programming-language-csharp#enable-runtime-driven-scaling

src/TriggerBinding/SqlScalerProvider.cs
AmeyaRele: the userFunctionId is gotten from HostIdProvider, and it would use this to  
https://github.com/Azure/azure-functions-host/blob/47411dfdc56bedd7f7dd20aa487fe69dbc596325/src/WebJobs.Script/Host/ScriptHostIdProvider.cs#L49

I saw that the scalecontroller also had a hostid which I thought I could use, but the values might be different, as it is using a different class namely FixedHostIdProvider please confirm @alrod 

We could go ahead and use triggerMetadata.FunctionName but we would have to use the same at places where userFunctionId is used, in order to get unprocessed changes, as Charles has mentioned.

src/TriggerBinding/SqlScalerProvider.cs
AmeyaRele: Changed to read from this section in config `$"azureWebJobs:extensions:sql:MaxChangesPerWorker"`. I believe this is the path for setting SqlOptions?

MaddyDev:
There are merge conflicts, please merge main. Please confirm with Alexey on using the hostId and we can have a discussion on getting userFunctionId.

AmeyaRele:
@Charles-Gagnon please review again

test/Microsoft.Azure.WebJobs.Extensions.Sql.Tests.csproj
```diff
@@ -10,6 +10,8 @@
     <PackageReference Include="Newtonsoft.Json" />
     <PackageReference Include="Microsoft.AspNetCore.Http" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" />
+	  <PackageReference Include="Microsoft.Extensions.Configuration.Json" />
```
Charles-Gagnon, line 14: nit: fix indentation

Charles-Gagnon (approved):

PR #964, opened 2023-11-06
Update docs with host.json options

MaddyDev:
Updated the instructions to redirect to how to use host.json options and also added note that we currently override app settings over host.json.

### DIFF
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -114,7 +114,8 @@ Azure SQL Trigger bindings utilize SQL [change tracking](https://docs.microsoft.
 
 ### Configuration for Trigger Bindings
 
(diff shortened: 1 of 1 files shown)


JetterMcTedder (changes requested):
Can we maybe rephrase this?
**NOTE:** We currently override configuration from app settings over host.json.

We are stating that we take the configuration in app settings and over write/use these instead of the ones in the host.json if present?

Maybe something like:

**NOTE:** Currently, SQL trigger bindings configurations from app settings will take precedence over settings in the host.json file.

JetterMcTedder (approved):

Charles-Gagnon (approved):

lucyzhang929 (approved):

MaddyDev:
Merging this in sine the test failure isn't related to the changes in the PR.

PR #970, opened 2023-11-27
Use CodeQL 3000 for Java build

Charles-Gagnon:
The old Semmle task is deprecated, so switching to just using the CodeQL 3000 task (which is auto-injected already)

image

I also renamed the TSA config file, since it appears to not be picked up with the other custom name. This should result in the issues being uploaded to the TSA database for easy viewing. 

Verifying that this is working correctly is difficult since CodeQL doesn't run against non-main branches by design, so I may need to do a follow up if this doesn't end up running the scan as expected.


.vscode/settings.json
```diff
@@ -8,6 +8,6 @@
     "omnisharp.enableEditorConfigSupport": true,
     "omnisharp.enableRoslynAnalyzers": true,
     "yaml.schemas": {
-         "https://raw.githubusercontent.com/microsoft/azure-pipelines-vscode/master/service-schema.json": "builds/azure-pipelines/**/*.yml"
```
Charles-Gagnon, line 11: I don't know why this keeps getting modified, but 4 spaces should be the current setting and this is what I get when running format document so I'm just changing that now since it was always showing as changed otherwise...

lucyzhang929 (approved):

PR #972, opened 2023-11-28
add dependabot.yml

MaddyDev:
adding dependabot.yml for enabling Dependabot.

There is a ignore attribute that we can use to explicitly [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-4) certain package updates (ignoring the major version updates for following semver). I haven't included them in this PR since we still have to approve to merge it in. I'll update the yml depending on if there are too many that we encounter.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


Charles-Gagnon (approved):

MaddyDev:
We have dependabot alerts enabled on the repo for vulnerabilities and I think that's why we saw the PRs that you mentioned:
image

But for receiving package update PRs, we need the yml that am adding. Let me get this in and check if we do need anything else or it works as expected. (I couldn't find any useful articles on how to test the dependabot alerts)

Issue #973, opened 2023-11-28
Verify CodeQL scans


Charles-Gagnon:
Make sure that the scans are working correctly for all projects/languages and bugs/issues are being filed in a way that's discoverable to us

lucyzhang929:
CodeQL task is working as expected for all languages except PowerShell and TSQL which are both still in beta support. Following up with the CodeQL team on this.

PR #976, opened 2023-11-29
Add codeql language for each stage

lucyzhang929:
This PR addresses https://github.com/Azure/azure-functions-sql-extension/issues/973.

Since our pipeline has two stages, we need to specify the language in each stage separately so that CodeQL will scan only the relevant files. In the case of the java-library build stage, we only build the java-library and we want CodeQL to scan that.
More information:
https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/codeql/codeql-codeql3000-faq#there-are-multiple-pipelines-or-stagesjobs-how-do-we-onboard

### DIFF
--- a/builds/azure-pipelines/build-release.yml
+++ b/builds/azure-pipelines/build-release.yml
@@ -39,7 +39,9 @@ variables:
 stages:
 - stage: BuildPublish
   displayName: 'Release Build and Publish'
-
+  variables:
+    Codeql.BuildIdentifier: extension
+    Codeql.Language: csharp,java,javascript,powershell,python,tsql
   jobs:
   - job: BuildTest
     displayName: 'Build and Test on '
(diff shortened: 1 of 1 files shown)


builds/azure-pipelines/build-release.yml
```diff
@@ -32,12 +32,16 @@ variables:
   LGTM.UploadSnapshot: true
   Codeql.Enabled: true
   Codeql.TSAEnabled: true
+  Codeql.PublishDatabaseLog: true
+  Codeql.Cadence: 0
```
lucyzhang929, line 36: The default cadence is 72 hours (meaning codeql will only run every 72 hours) so I am temporarily updating the cadence to 0 to check if codeql is running as expected.

Charles-Gagnon (approved):

PR #990, opened 2023-12-19
add traceroute against api.esrp.microsoft.com 

MaddyDev:
As suggested in the [ICM](https://portal.microsofticm.com/imp/v3/incidents/incident/447617542/summary), added the tracert step for the mac build. 
Following up with the on-call engineer on the Icm if this is a required step and sending this PR for temporarily unblocking the build.


lucyzhang929 (approved):

VasuBhog (approved):

Charles-Gagnon (approved):

Charles-Gagnon:
@MaddyDev A couple of suggestions : 

1. Make a task to track fixing the code signing
2. Add a comment/description in the title of this task to make it clear what it's for (and ideally mention it in the issue so we can track removing it once the issue has been determined and resolved)

PR #999, opened 2024-01-02
Clarify exception retries section

Charles-Gagnon:
Clarifying that the 60sec delay also applies to exceptions that occur in the binding code itself, such as deadlocks causing the a query to fail. If that happens we clear the rows currently being processed and don't try again until 60sec has elapsed.


lucyzhang929 (approved):

Charles-Gagnon:
Build failure is unrelated, since this is just doc update it's safe to merge

PR #1015, opened 2024-01-17
Don't cache query type

Charles-Gagnon:
Found this when updating the xUnit packages. Previously, the tests ran slow enough that the cached data timed out and so the later tests weren't affected by it. But with the update they all ran together in <10min, which meant that we used the same table info cached for the first test ran (which happened to be AddProductIncorrectCasing.

The issue is this - when we cached the table info we were also caching the query type. But the query type is based off of the incoming object for an output request, which means that since we were caching the info only at the table level (the cache key is connection string + table name) that we'd reuse that query type for ALL queries after that first one within that 10min span. For most cases this probably isn't an issue - we generally prefer to do a merge anyways and only fall back to insert in special circumstances.

But it's a valid scenario that there might be two separate functions which operate differently on the same table - in which case they could easily run into problems where the first one to run causes the other one to start using the wrong query type (like we had here).

In the case that I found this showed up as this error : 

`[3435] [2024-01-17T02:34:50.108Z] System.Private.CoreLib: Exception while executing function: Functions.AddProductsArray. Microsoft.Azure.WebJobs.Host: Error while handling parameter _binder after function returned:. Microsoft.Azure.WebJobs.Extensions.Sql: Unexpected error upserting rows. Core Microsoft SqlClient Data Provider: Violation of PRIMARY KEY constraint 'PK__Products__B40CC6CDD4C38CC3'. Cannot insert duplicate key in object 'dbo.Products'. The duplicate key value is (2).`

Which was because it was trying to do an INSERT instead of a MERGE (where it would have updated the row instead of trying to insert a new one)

This is a follow up to https://github.com/Azure/azure-functions-sql-extension/pull/628 where we did something similar

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -202,7 +202,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 {
                     TelemetryInstance.TrackEvent(TelemetryEventName.TableInfoCacheMiss, props);
                     // set the columnNames for supporting T as JObject since it doesn't have columns in the member info.
-                    tableInfo = TableInformation.RetrieveTableInformation(connection, fullTableName, this._logger, GetColumnNamesFromItem(rows.First()), this._serverProperties);
+                    tableInfo = TableInformation.RetrieveTableInformation(connection, fullTableName, this._logger, this._serverProperties);
                     var policy = new CacheItemPolicy
                     {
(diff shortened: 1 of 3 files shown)


Charles-Gagnon:
The failures are due to CG issues being fixed in another PR. I won't merge this in until that's in and the runs here are green, but in the meantime the tests are passing in the build so feel free to review the changes since there shouldn't be anything else needed.

lucyzhang929 (approved):

PR #1033, opened 2024-01-29
invalid operator used to compare with NULL values in get primary keys query

sumitkumams:
Problem: should not use assignment '=' operator comparing with NULL 
Fix: best to use 'IS NULL' here to get expected value of 'has_default'

Impact Area:  **I have no idea on impact area of this change**. So, I am leaving this on your team's responsibility to make sure, if this change fixes any existing problems in your knowledge.

NOTE to reviewers: The reason of change that I found the existing query as inappropriate. So, **please consider this change after proper testing**.

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -474,7 +474,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                         ccu.{ColumnName},
                         c.is_identity,
                         case
-                            when isc.COLUMN_DEFAULT = NULL then 'false'
+                            when isc.COLUMN_DEFAULT IS NULL then 'false'
                             else 'true'
                         end as {HasDefault}
                     FROM


Charles-Gagnon:
Thanks for submitting this! We'll investigate - I would have expected the test added in https://github.com/Azure/azure-functions-sql-extension/pull/388 to not have worked with the check being broken like this, but maybe I'm missing something.

Did you encounter a specific issue with a repro you can file an issue for to help validate this change? Or did you just notice the incorrect syntax while looking through at the code/queries directly?

MaddyDev:
@Charles-Gagnon Sumit reached out to me via teams and he said he didn't encounter any issue but noticed the incorrect syntax and sent the PR. Am looking into it and will see why the test didn't catch it and update accordingly.

MaddyDev:
Closing this, added the fix in #1039

PR #1035, opened 2024-02-02
Bump the versions of Xunit and ScriptDom by dependabot

MaddyDev:
Bumping the dependencies to address the below 3 PRs:

https://github.com/Azure/azure-functions-sql-extension/pull/1030
https://github.com/Azure/azure-functions-sql-extension/pull/1032

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "Azure.Identity"
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,8 @@ updates:
     schedule:
       interval: "weekly"
     ignore:
+        # Please do not bump Azure.Identity from 1.10.2, even a minor bump to 1.10.3 is causing a major bump in it's trasitive dependecny Microsoft.Identity.Client.Extensions.Msal which will fail the bundle test.
       - dependency-name: "Azure.Identity"
+        # Ignore all major version bumps for semver.
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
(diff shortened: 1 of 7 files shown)


src/packages.lock.json
```diff
@@ -364,18 +364,18 @@
...
+        "contentHash": "rr4zbidvHy9r4NvOAs5hdd964Ao2A0pAeFBJKR95u1CJAVzbd1p6tPTXUZ+5ld0cfThiVSGvz6UHwY6JjraTpA==",
         "dependencies": {
           "Microsoft.IdentityModel.Abstractions": "6.22.0"
         }
       },
       "Microsoft.Identity.Client.Extensions.Msal": {
         "type": "Transitive",
-        "resolved": "2.31.0",
-        "contentHash": "IhGSqN0szneKC5Qk3/okJQJbDpQfLW/+mvslhzJPox4t2UuIkA2ZHe4w/z62ASye46G9sQWF9qqLXTgNacE2xQ==",
+        "resolved": "4.56.0",
```
Charles-Gagnon, line 375: Isn't this going to fail the bundle test?

MaddyDev: It depends on the fact that if they have Microsoft.Identity.Client.Extensions.Msal in their deps.json. If it is not present it will not fail. But I can run and check this.

Charles-Gagnon: Any dependencies of ours should be transitive dependencies of theirs too, so yeah I expect this will fail. But let's see what the test says - maybe it's doing something differently.

.github/dependabot.yml
```diff
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "Azure.Identity"
```
Charles-Gagnon, line 8: It would be good to add a comment here too explaining why this is ignored

Charles-Gagnon (approved):
Would suggest updating the PR title and description to include mentioning that Azure.Identity isn't actually being updated

PR #1042, opened 2024-02-15
Fix output binding when using column names with slashes

### LINKED ISSUE #1026: Output binding is unable to update column that contains a slash in their name.
- Azure Functions SQL Extension or Extension Bundle Version:
```json
  "extensionBundle": {
      "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[4.*, 5.0.0)"
  }
```
- Is this a deployed or local function:

Issue is present on both deployed and local

- What type of Database are you using? (Run `SELECT @@VERSION as Version, SERVERPROPERTY('EngineEdition') as EngineEdition` on your database):

| Version                                                                                                    | EngineEdition |
| ---------------------------------------------------------------------------------------------------------- | ------------- |
| Microsoft SQL Azure (RTM) - 12.0.2000.8  	Jan  3 2024 16:32:18  	Copyright (C) 2022 Microsoft Corporation | 5             |

- List any custom settings for your function app. This could be a custom time-out defined specifically for your database server or optional configuration that can be customized for the app defined [here](https://learn.microsoft.com/azure/azure-functions/functions-bindings-azure-sql-trigger?tabs=isolated-process%2Cportal&pivots=programming-language-csharp#optional-configuration).

The contents of function.json:
```json
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "myblob",
      "type": "blobTrigger",
      "direction": "in",
      "path": "my_blob_path",
      "connection": "my_connection_string"
    },
    {
      "name": "tableCols",
      "type": "sql",
      "direction": "in",
      "commandText": "SELECT TABLE_NAME, COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = 'my_table';",
      "commandType": "Text",
      "connectionStringSetting" : "SqlConnectionString"
    },
    {
      "name": "outputTable",
      "type": "sql",
      "direction": "out",
      "commandText": "dbo.my_table",
      "connectionStringSetting": "SqlConnectionString"
    }
  ]
}

```

Contents of `__init__.py`:
```python
def main(
    myblob: func.InputStream,
    tableCols: func.SqlRowList,
    outputTable: func.Out[func.SqlRow],
  ):
  """Main function"""
  blob_contents = myblob.read()
  blob_data = create_dataframe(blob_contents)

  database_columns = {sql_row['COLUMN_NAME'] for sql_row in tableCols}
  
  cols_in_dataset = set(blob_data.columns)
  columns_to_insert = list(cols_in_dataset & cols_in_db)

  records = (
      blob_data
      [columns_to_insert]
      .dropna(how='all', axis=1)
      .to_dict('records')
  )

  output_rowlist = func.SqlRowList(func.SqlRow.from_dict(row) for row in records)

  outputTable.set(output_rowlist)
```

SQL create statement for the target table:
```SQL
SET ANSI_NULLS ON
GO
SET QUOTED_IDENTIFIER ON
GO
CREATE TABLE [dbo].[my_table](
	[ID1] [varchar](4) NOT NULL,
	[Foo1] [varchar](255) NULL,
	[Foo2] [tinyint] NULL,
	[Foo/Bar3] [tinyint] NULL,
) ON [PRIMARY]
GO
SET ANSI_PADDING ON
GO
ALTER TABLE [dbo].[my_table] ADD  CONSTRAINT [PK_my_table] PRIMARY KEY CLUSTERED 
(
	[ID1] ASC

)WITH (STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ONLINE = OFF, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
GO
EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'This is my database table' , @level0type=N'SCHEMA',@level0name=N'dbo', @level1type=N'TABLE',@level1name=N'my_table'
GO
```

In the above Azure Function, the output_rowlist which is used to set the output binding consists of SqlRows which look like this:
{'ID1': '9999', 'Foo1': None, 'Foo2': '1', 'Foo/Bar3': '1'}

When the function runs, all of the columns are updated properly except for 'Foo/Bar3', which is always set to NULL regardless of the value present in output_rowlist.
When I update my_table using an SQL statement in Azure Data Studio, the column 'Foo/Bar3' updates without issue.
I have since changed the name of the column in the database to 'FooBar3' and the output binding updates the column correctly. I have not seen similar issues when updating columns that contain non-alphanumeric characters ie %, #.

MaddyDev:
Replace the / slash while building the query for inserting the items.

Fixes #1026

### DIFF
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -412,6 +412,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             rowData = Utils.JsonSerializeObject(rowsToUpsert, table.JsonSerializerSettings);
             IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
             IEnumerable<string> bracketColumnDefinitionsFromItem = columnNamesFromItem.Select(c => $"{c.AsBracketQuotedString()} {table.Columns[c]}");
+            // Escape any forward and backward slashes in the column names of rowData using REPLACE so the OPENJSON can read from those columns.
             newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON(REPLACE(REPLACE({RowDataParameter}, N'\\', N'\\\\'), N'/', N'\\/')) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
         }
 
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -412,7 +412,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             rowData = Utils.JsonSerializeObject(rowsToUpsert, table.JsonSerializerSettings);
             IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
             IEnumerable<string> bracketColumnDefinitionsFromItem = columnNamesFromItem.Select(c => $"{c.AsBracketQuotedString()} {table.Columns[c]}");
-            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON({RowDataParameter}) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
+            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON(REPLACE(REPLACE({RowDataParameter}, N'\\', N'\\\\'), N'/', N'\\/')) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
         }
 
         public class TableInformation
@@ -633,7 +633,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                     throw new InvalidOperationException(message, ex);
                 }
 
-                if (!primaryKeys.Any())
+                if (primaryKeys.Count == 0)
                 {
                     string message = $"Did not retrieve any primary keys for {table}. Cannot generate upsert command without them.";
                     var ex = new InvalidOperationException(message);
--- a/src/SqlAsyncCollector.cs
+++ b/src/SqlAsyncCollector.cs
@@ -413,7 +413,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
             IEnumerable<string> bracketColumnDefinitionsFromItem = columnNamesFromItem.Select(c => $"{c.AsBracketQuotedString()} {table.Columns[c]}");
             // Escape any forward and backward slashes in the column names of rowData using REPLACE so the OPENJSON can read from those columns.
-            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON(REPLACE(REPLACE({RowDataParameter}, N'\\', N'\\\\'), N'/', N'\\/')) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
+            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON(REPLACE({RowDataParameter}, N'/', N'\\/')) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
         }
 
         public class TableInformation
--- a/test/Database/Tables/ProductsWithSlashInColumnNames.sql
+++ b/test/Database/Tables/ProductsWithSlashInColumnNames.sql
@@ -3,5 +3,5 @@ DROP TABLE IF EXISTS [ProductsWithSlashInColumnNames];
 CREATE TABLE [ProductsWithSlashInColumnNames] (
     [ProductId] [int] NOT NULL PRIMARY KEY,
     [Name/Test] [nchar](100) NOT NULL,
-	[Cost/Test] [int] NOT NULL
+    [Cost\Test] [int] NOT NULL
 )
--- a/test/Database/Tables/ProductsWithSlashInColumnNames.sql
+++ b/test/Database/Tables/ProductsWithSlashInColumnNames.sql
@@ -1,0 +1,7 @@
+DROP TABLE IF EXISTS [ProductsWithSlashInColumnNames];
+
+CREATE TABLE [ProductsWithSlashInColumnNames] (
+    [ProductId] [int] NOT NULL PRIMARY KEY,
+    [Name/Test] [nchar](100) NOT NULL,
+	[Cost/Test] [int] NOT NULL
+)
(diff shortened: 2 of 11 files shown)


lucyzhang929 (approved):

lucyzhang929:
@MaddyDev Since this new test is not being ran for CSharp, you'll have to add the test to the filter here: https://github.com/Azure/azure-functions-sql-extension/blob/d3c48a4e3d49eec7d94714245de30fada2b07a13/builds/azure-pipelines/template-steps-build-test.yml#L185C93-L185C111.
That should fix the failure in the pipeline.

src/SqlAsyncCollector.cs
```diff
@@ -412,7 +412,7 @@ private static void GenerateDataQueryForMerge(TableInformation table, IEnumerabl
             rowData = Utils.JsonSerializeObject(rowsToUpsert, table.JsonSerializerSettings);
             IEnumerable<string> columnNamesFromItem = GetColumnNamesFromItem(rows.First());
             IEnumerable<string> bracketColumnDefinitionsFromItem = columnNamesFromItem.Select(c => $"{c.AsBracketQuotedString()} {table.Columns[c]}");
-            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON({RowDataParameter}) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
+            newDataQuery = $"WITH {CteName} AS ( SELECT * FROM OPENJSON(REPLACE(REPLACE({RowDataParameter}, N'\\', N'\\\\'), N'/', N'\\/')) WITH ({string.Join(",", bracketColumnDefinitionsFromItem)}) )";
```
Charles-Gagnon, line 415: Add comment about why we're escaping

src/SqlAsyncCollector.cs
```diff
@@ -633,7 +633,7 @@ public static TableInformation RetrieveTableInformation(SqlConnection sqlConnect
                     throw new InvalidOperationException(message, ex);
                 }
 
-                if (!primaryKeys.Any())
+                if (primaryKeys.Count == 0)
```
Charles-Gagnon, line 636: Was there a specific reason you changed this?

test/Database/Tables/ProductsWithSlashInColumnNames.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE IF EXISTS [ProductsWithSlashInColumnNames];
+
+CREATE TABLE [ProductsWithSlashInColumnNames] (
+    [ProductId] [int] NOT NULL PRIMARY KEY,
+    [Name/Test] [nchar](100) NOT NULL,
+	[Cost/Test] [int] NOT NULL
```
Charles-Gagnon, line 6: nit: Fix indentation
```suggestion
    [ProductId] [int] NOT NULL PRIMARY KEY,
    [Name/Test] [nchar](100) NOT NULL,
    [Cost/Test] [int] NOT NULL
```

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -551,5 +551,22 @@ public async Task AddProductDefaultPKAndDifferentColumnOrderTest(SupportedLangua
             await this.SendOutputGetRequest("addproductdefaultpkanddifferentcolumnorder", null, TestUtils.GetPort(lang, true));
             Assert.Equal(1, this.ExecuteScalar("SELECT COUNT(*) FROM dbo.ProductsWithDefaultPK"));
         }
+
+        /// <summary>
+        /// Tests that when using a table with column names having slash (/ or \) in them 
+        /// we can insert data using output binding.
+        /// Excluding C#, JAVA and CSX languages since / or \ are reserved characters and are discouraged to use in variable names
```
Charles-Gagnon, line 558: nit: It's not just discouraged, it's specifically not allowed. Would suggest making that clear to avoid confusion

Charles-Gagnon (approved):

src/SqlAsyncCollector.cs
Charles-Gagnon, line 415: Can you add tests for backslashes in the names as well?

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -551,5 +551,22 @@ public async Task AddProductDefaultPKAndDifferentColumnOrderTest(SupportedLangua
...
+        [Theory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.CSharp, SupportedLanguages.OutOfProc, SupportedLanguages.Java, SupportedLanguages.Csx)]
+        public async Task AddProductWithSlashInColumnName(SupportedLanguages lang)
+        {
+            Assert.Equal(0, this.ExecuteScalar("SELECT COUNT(*) FROM dbo.ProductsWithSlashInColumnNames"));
+            this.StartFunctionHost("AddProductWithSlashInColumnName", lang);
+            await this.SendOutputPostRequest("addproduct-slashcolumns", "");
+            // Check that a product should have been inserted
+            Assert.Equal(1, this.ExecuteScalar("SELECT COUNT(*) FROM dbo.ProductsWithSlashInColumnNames"));
```
Charles-Gagnon, line 569: Shouldn't this test be verifying that the value in the columns is valid? If I'm reading the issue correctly - it inserted a row but the values in the columns with slashes were just NULL

test/Integration/SqlOutputBindingIntegrationTests.cs
```diff
@@ -551,5 +551,22 @@ public async Task AddProductDefaultPKAndDifferentColumnOrderTest(SupportedLangua
...
+        /// Excluding C#, JAVA and CSX languages since / or \ are reserved characters and are discouraged to use in variable names
+        /// </summary>
+        [Theory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.CSharp, SupportedLanguages.OutOfProc, SupportedLanguages.Java, SupportedLanguages.Csx)]
+        public async Task AddProductWithSlashInColumnName(SupportedLanguages lang)
+        {
+            Assert.Equal(0, this.ExecuteScalar("SELECT COUNT(*) FROM dbo.ProductsWithSlashInColumnNames"));
+            this.StartFunctionHost("AddProductWithSlashInColumnName", lang);
+            await this.SendOutputPostRequest("addproduct-slashcolumns", "");
```
Charles-Gagnon, line 567: Normally we create the object here and pass it in the POST request (which is typically how people will be using output bindings). Is there a reason you chose to do it in the test functions themselves directly instead?

MaddyDev: I just followed the pattern we had for all the negative scenario tests. All of those have the data hardcoded in the function (not sure why we did that to begin with). But I agree with what you said, will add one more insert and test both.

PR #1071, opened 2024-04-25
Migrating data to newUserFunctionId tables as part of Consumption plan support

MaddyDev:
**New UserFunctionId**:
We have used the [WEBSITE_SITE_NAME](https://learn.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#app-environment) instead of the host id for creating the new **_userFunctionId_**

**Migration** steps followed:
1. While creating the lease table entry in the global state table, check if there a value with the oldUserFunctionId and use that value while inserting a new entry with the new userFunctionId to the global state table. And delete the oldUserFunctionId  entry.
2. Create new leases table with new hash and if old leases table exists, copy over all the data from it and delete the table.

**Test**: To Be Done


src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -389,16 +392,30 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
                 }
             }
 
-            string insertRowGlobalStateTableQuery = $@"
+            bool migrateOldUserFunctionIdData = await this.OldUserFunctionIdRecordExists(connection, transaction, userTableId, cancellationToken);
```
Charles-Gagnon, line 395: Do we need to do this check? Couldn't you just make each of the below queries do the check regardless and migrate as needed? Would save a bit of execution time and duplication of the query.

MaddyDev: The reason I added the method is to add a telemetry event to track migration which I totally missed to add here. Will look into if there is a way to not have this check and still track migration.

Charles-Gagnon: Ok, although given how little we look at this stuff currently I'd question whether we actually care about tracking this. Is that something you see being valuable?

Anyways it's fine if you want to keep this, not a huge deal (and we do other similar things elsewhere anyways).

MaddyDev: I don't really think that tracking is required here but definitely useful when deciding on deprecation path, we can look if there are any still using the migration process and take a call. Plus Chris asked me to add a telemetry point :)

Charles-Gagnon: We could just look at version numbers for that, right? It's not like people can opt out of this so as long as everyone is on the version with this change or higher we should be able to assume they've migrated

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -95,14 +96,37 @@ public ParameterDescriptor ToParameterDescriptor()
         /// <summary>
         /// Returns an ID that uniquely identifies the user function.
         ///
+        /// We call the WEBSITE_SITE_NAME from the configuration and use that to create the hash of the
+        /// user function id. Appending another hash of class+method in here ensures that if there
+        /// are multiple user functions within the same process and tracking the same SQL table, then each one of them
+        /// gets a separate view of the table changes.
+        /// </summary>
+        private string GetUserFunctionIdAsync()
```
alrod, line 104: Remove Async from the name

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -426,18 +443,41 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
         {
             string primaryKeysWithTypes = string.Join(", ", primaryKeyColumns.Select(col => $"{col.name.AsBracketQuotedString()} {col.type}"));
             string primaryKeys = string.Join(", ", primaryKeyColumns.Select(col => col.name.AsBracketQuotedString()));
+            string OldLeasesTableName = leasesTableName.Contains(this._userFunctionId) ? leasesTableName.Replace(this._userFunctionId, this._oldUserFunctionId) : string.Empty;
```
alrod, line 446: oldLeasesTableName

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -391,19 +394,33 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
 
             string insertRowGlobalStateTableQuery = $@"
                 {AppLockStatements}
-
+                
                 IF NOT EXISTS (
                     SELECT * FROM {GlobalStateTableName}
                     WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {userTableId}
                 )
+                BEGIN
```
alrod, line 402: Could you please add a comment explaining why do we need _oldUserFunctionId and whats is GlobalStateTableName

MaddyDev: Yes, this for back compatibility. We have the version at which the function is in the GlobalState table, this is to copy that version from the `oldUserFunctionId` to the new `userFunctionId` and continue the function.

src/SqlBindingConstants.cs
```diff
@@ -7,6 +7,8 @@ internal static class SqlBindingConstants
     {
         public const string ISO_8061_DATETIME_FORMAT = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffZ";
 
+        public const string WebsiteName = "WEBSITE_SITE_NAME";
```
chlafreniere, line 10: nit: all uppercase variable name for const

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -426,18 +443,41 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
...
+                        {LeasesTableChangeVersionColumnName} bigint NOT NULL,
+                        {LeasesTableAttemptCountColumnName} int NOT NULL,
+                        {LeasesTableLeaseExpirationTimeColumnName} datetime2,
+                        PRIMARY KEY ({primaryKeys})
+                    );
+            " : $@"
                 {AppLockStatements}
 
                 IF OBJECT_ID(N'{leasesTableName}', 'U') IS NULL
+                Begin
```
chlafreniere, line 463: Let's remain consistent and use all uppercase for TSQL

src/SqlBindingUtilities.cs
```diff
@@ -54,6 +54,15 @@ public static string GetConnectionString(string connectionStringSetting, IConfig
             return connectionString;
         }
 
+        public static string GetWebSiteName(IConfiguration configuration)
+        {
+            if (configuration == null)
```
chlafreniere, line 59: What happens for local development? Is website site name set?

src/Telemetry/Telemetry.cs
```diff
@@ -348,7 +348,8 @@ public enum TelemetryEventName
         TableInfoCacheMiss,
         TriggerFunction,
         TriggerMonitorStart,
-        Upsert
+        Upsert,
+        InsertGlobalStateTableRow,
```
chlafreniere, line 352: nit: unnecessary extra `,`

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -78,8 +78,9 @@ public async Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
         {
             _ = context ?? throw new ArgumentNullException(nameof(context), "Missing listener context");
 
-            string userFunctionId = await this.GetUserFunctionIdAsync();
-            return new SqlTriggerListener<T>(this._connectionString, this._tableName, this._leasesTableName, userFunctionId, context.Executor, this._sqlOptions, this._logger, this._configuration);
+            string userFunctionId = this.GetUserFunctionId();
```
chlafreniere, line 81: Even after a migration is done once, are we still calculating these old + new values? Is that really needed?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -391,19 +394,33 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
...
             string insertRowGlobalStateTableQuery = $@"
                 {AppLockStatements}
-
+                -- For back compatibility copy the lastSyncVersion from _oldUserFunctionId if it exists.
                 IF NOT EXISTS (
                     SELECT * FROM {GlobalStateTableName}
                     WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {userTableId}
                 )
+                BEGIN
+                    -- Migrate LastSyncVersion from oldUserFunctionId if it exists and delete the record
```
chlafreniere, line 403: Please use all caps for T-SQL

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -391,19 +394,33 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
...
+                    Declare @lastSyncVersion bigint;
+                    Select @lastSyncVersion = LastSyncVersion from az_func.GlobalState where UserFunctionID = '{this._oldUserFunctionId}' AND UserTableID = {userTableId}
```
chlafreniere, line 405: Plz check for sql injection possibilities

chlafreniere:
@Charles-Gagnon can you take another look?

src/TriggerBinding/SqlTriggerListener.cs
```diff
@@ -391,19 +394,33 @@ private async Task<long> InsertGlobalStateTableRowAsync(SqlConnection connection
...
+                    IF @lastSyncVersion is NULL
+                        SET @lastSyncVersion = {(long)minValidVersion};
+                    ELSE
+                        Delete from az_func.GlobalState where UserFunctionID = '{this._oldUserFunctionId}' AND UserTableID = {userTableId}
+                    
                     INSERT INTO {GlobalStateTableName}
                     VALUES ('{this._userFunctionId}', {userTableId}, {(long)minValidVersion}, GETUTCDATE());
```
Charles-Gagnon, line 412: Shouldn't this be using @lastSyncVersion? That'll be the old value is we're migrating over.

Charles-Gagnon (changes requested):
There's one potential bug to look at, and then please make sure to respond to Chris' comments as well. Otherwise looks good though!

src/SqlBindingUtilities.cs
MaddyDev: WEBSITE_SITE_NAME value is null for local development.

src/TriggerBinding/SqlTriggerBinding.cs
MaddyDev: Yes, we decided to have the queries such that if old values are present, migrate them and clean up else continue with new values.

Charles-Gagnon (approved):
I would definitely recommend updating the [README](https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/TriggerBinding.md#az_funcleasestablename)/MS docs to make it clear how the app name relates to the table and the consequences of changing that. 

One thing I'm not sure we tested before is whether changing the app name previously would cause the host ID to change (and thus "lose" the history). I believe that would have been the case even previously, but now it's much more explicit so it'd be good to add that information in.

chlafreniere (approved):

MaddyDev:
> I would definitely recommend updating the [README](https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/TriggerBinding.md#az_funcleasestablename)/MS docs to make it clear how the app name relates to the table and the consequences of changing that.
> 
> One thing I'm not sure we tested before is whether changing the app name previously would cause the host ID to change (and thus "lose" the history). I believe that would have been the case even previously, but now it's much more explicit so it'd be good to add that information in.

Will be adding the docs along with an end to end test in a separate PR. Thanks for calling this out.

PR #1091, opened 2024-06-21
UserfunctionId change and integration test.

MaddyDev:
For enabling the SQL on Scale controller, Ameya has a [PR](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController/pullrequest/9035947) which requires further changes in userfunctionId since the current namespace+functionName is not available on scale controller.
I updated the userfucntionId to just use functionName. The implications of doing that is: For local functions if the user doesn't provide a WEBSITE_SITE_NAME setting and uses the same functionName across apps, they use the same leases table. Deployed functions read the WEBSITE_SITE_NAME which is a read-only environment variable and does not run into this issue. Below are the changes addressed in this PR:
1. I made the WEBSITE_SITE_NAME setting required and added a documentation section.
2. Updated the samples to have the WEBSITE_SITE_NAME setting in local.settings.json
3. Added an integration test which runs a trigger function before and after migration.

### DIFF
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -280,7 +280,7 @@ steps:
 - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
   displayName: 'Component Detection'
   inputs:
-    failOnAlert: false
+    failOnAlert: true
 
 - task: securedevelopmentteam.vss-secure-development-tools.build-task-postanalysis.PostAnalysis@2
   displayName: 'Post Analysis'
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -280,7 +280,7 @@ steps:
 - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
   displayName: 'Component Detection'
   inputs:
-    failOnAlert: true
+    failOnAlert: false
 
 - task: securedevelopmentteam.vss-secure-development-tools.build-task-postanalysis.PostAnalysis@2
   displayName: 'Post Analysis'
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -156,6 +156,7 @@ The upper limit on the number of pending changes in the user table that are allo
 
 The unique name used in creating the lease tables. The local apps depend on this setting for creating unique leases tables, please give a unique name for each app.
     > **NOTE:** If the setting is re-used accross apps, having the same function name could cause the functions to use the same lease tables and the function runs to not work as expected.
+    > **NOTE:** If you have 2 different SQL trigger functions with same functionName locally, not having WEBSITE_SITE_NAME would mean that the same leasees table would be used for both triggers resulting in only one of the functions being triggered.
     > **NOTE:** This is a read-only variable that is provided by the azure environment variables for deployed functions and the user provided value will be overridden. Refer to [Environment variables](https://learn.microsoft.com/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#app-environment) for apps.
 
 ### Scaling for Trigger Bindings
--- a/docs/BindingsOverview.md
+++ b/docs/BindingsOverview.md
@@ -152,6 +152,12 @@ The delay in milliseconds between processing each batch of changes.
 
 The upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#scaling-for-trigger-bindings) section for more information.
 
+#### WEBSITE_SITE_NAME
+
+The unique name used in creating the lease tables. The local apps depend on this setting for creating unique leases tables, please give a unique name for each app.
+    > **NOTE:** If the setting is re-used accross apps, having the same function name could cause the functions to use the same lease tables and the function runs to not work as expected.
+    > **NOTE:** This is a read-only variable that is provided by the azure environment variables for deployed functions and the user provided value will be overridden. Refer to [Environment variables](https://learn.microsoft.com/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#app-environment) for apps.
+
 ### Scaling for Trigger Bindings
 
 If your application containing functions with SQL trigger bindings is running as an Azure function app, it will be scaled automatically based on the amount of changes that are pending to be processed in the user table. As of today, we only support scaling of function apps running in Elastic Premium plan with 'Runtime Scale Monitoring' enabled. To enable scaling, you will need to go the function app resource's page on Azure Portal, then to Configuration > 'Function runtime settings' and turn on 'Runtime Scale Monitoring'.
--- a/src/SqlBindingUtilities.cs
+++ b/src/SqlBindingUtilities.cs
@@ -64,8 +64,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
             // We require a WEBSITE_SITE_NAME for avoiding duplicates if users use the same function name accross apps.
             if (string.IsNullOrEmpty(websitename))
             {
-                throw new ArgumentException(websitename == null ? $"WEBSITE_SITE_NAME setting is missing in your function app settings, please add the setting with a string value. Please refer to https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsOverview.md#website_site_name for more information." :
-                $"WEBSITE_SITE_NAME setting is empty in your function app settings, please update the setting with a string value.");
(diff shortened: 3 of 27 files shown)


src/SqlBindingUtilities.cs
```diff
@@ -60,7 +60,14 @@ public static string GetWebSiteName(IConfiguration configuration)
             {
                 throw new ArgumentNullException(nameof(configuration));
             }
-            return configuration.GetConnectionStringOrSetting(SqlBindingConstants.WEBSITENAME);
+            string websitename = configuration.GetConnectionStringOrSetting(SqlBindingConstants.WEBSITENAME);
+            // We require a WEBSITE_SITE_NAME for avoiding duplicates if users use the same function name accross apps.
+            if (string.IsNullOrEmpty(websitename))
+            {
+                throw new ArgumentException(websitename == null ? $"WEBSITE_SITE_NAME setting is missing in your function app settings, please add the setting with a string value. Please refer to https://github.com/Azure/azure-functions-sql-extension/blob/main/docs/BindingsOverview.md#website_site_name for more information." :
```
Charles-Gagnon, line 67: Would suggest putting the link in both. And in fact, I'd just simplify this into one and say something like "WEBSITE_SITE_NAME can't be null or empty in your function app settings, please add..."

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -107,7 +107,7 @@ private string GetUserFunctionId()
             string websiteName = SqlBindingUtilities.GetWebSiteName(this._configuration);
 
             var methodInfo = (MethodInfo)this._parameter.Member;
-            string functionName = $"{methodInfo.DeclaringType.FullName}.{methodInfo.Name}";
+            string functionName = $"{methodInfo.Name}";
```
Charles-Gagnon, line 110: If someone has two functions in their app that share a name this means they'll share a lease table now.

What is special about the scale controller that you can't get the FullName of the type it's declared in?

MaddyDev: Build will fail if user tries to create different functions with same name in an app. So that's covered. But if the user creates same function name across different apps - with this change, they could end up using the same lease table.
@AmeyaRele said he can only access the function name on the Scale controller [PR](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController/pullrequest/9035947).

Charles-Gagnon: You can have multiple methods with the same name being used for functions as long as they : 

1. Are in different classes
2. Have different FunctionName attribute values

(note this is for C# - other languages may not allow this but since C# does it doesn't matter if others do or not)

e.g. 

```csharp
namespace Company.Function
{
    public static class AddProduct
    {
        [FunctionName("AddProduct")]
        public static IActionResult Run(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "addproduct")]
            [FromBody] Product prod,
            [Sql("dbo.Products", "SqlConnectionString")] out Product product)
        {
            product = prod;
        }
    }
}
```

and 

```csharp
namespace Company.Function.Othernamespace
{
    public static class AddProduct
    {
        [FunctionName("AddProduct2")]
        public static IActionResult Run(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "addproduct")]
            [FromBody] Product prod,
            [Sql("dbo.Products", "SqlConnectionString")] out Product product)
        {
            product = prod;
        }
    }
}
```

Charles-Gagnon: Most of our samples all have the same method name `Run` - and I'm going to guess this is pretty common for users too so would be pretty impactful to make this change.

MaddyDev: methodInfo.Name gives me AddProduct for the first one and AddProduct2 for the second one. I'll have to test for non dotnet languages and confirm the same.

Charles-Gagnon: Well that's odd - can you check if it's pulling it from the class name or the FunctionName?

Might be worth asking the core team where this value is populated from. Because for non-C# languages I'm curious how they would even get a "MethodInfo" given that that's specifically for reflection. But I suppose it would make sense if they decided to shim in their own version of it based on other stuff...

If it using the FunctionName then that should be fine though.

MaddyDev: I couldn't debug and see what's the value returned for methodInfo.DeclaringType.Name on javascript functions (not able to debug even after replacing the local dll in the bundle path), but I did run two different functions under the same app and it created two different leases table so definitely not taking the method name which is same for both the functions.

MaddyDev:
@alrod @fabiocav I've added a new setting in this PR and made it required. This only effects the local functions since the setting is a read-only environment variable for deployed apps. Would this be considered a breaking change?

docs/BindingsOverview.md
```diff
@@ -152,6 +152,12 @@ The delay in milliseconds between processing each batch of changes.
 
 The upper limit on the number of pending changes in the user table that are allowed per application-worker. If the count of changes exceeds this limit, it may result in a scale out. The setting only applies for Azure Function Apps with runtime driven scaling enabled. See the [Scaling](#scaling-for-trigger-bindings) section for more information.
 
+#### WEBSITE_SITE_NAME
+
```
alrod, line 156: Can we also add this note:

"If they have 2 sql triggers with same functionName locally, not having WEBSITE_SITE_NAME would mean that the same checkpointing tables would be used for both triggers. So only one of the functions would be triggered. "

alrod (approved):
Approved with minor comment

test/Integration/SqlTriggerBindingIntegrationTests.cs
```diff
@@ -872,5 +872,68 @@ public async Task NewUserFunctionId_Migration_Test()
             Assert.True(1 == (int)this.ExecuteScalar($@"SELECT 1 FROM {GlobalStateTableName} WHERE UserFunctionID = N'{userFunctionId}'"), $"{GlobalStateTableName} should have {userFunctionId} row on successful migration");
             Assert.True(lastSyncVersion == (long)this.ExecuteScalar($@"SELECT LastSyncVersion FROM {GlobalStateTableName} WHERE UserFunctionID = N'{userFunctionId}'"), $"{GlobalStateTableName} should have {userFunctionId} row woth on successful migration");
         }
+
+        /// <summary>
+        /// Ensures that the user function gets invoked for each of the insert, update and delete operation after migration seamlessly.
+        /// </summary>
+        [RetryTheory]
+        [SqlInlineData()]
+        [UnsupportedLanguages(SupportedLanguages.Java)] // test timing out for Java
```
Charles-Gagnon, line 881: Have you verified manually that this works for Java? If the test is failing because of a timeout that's a bit worrisome to me

Charles-Gagnon (approved):

test/Integration/SqlTriggerBindingIntegrationTests.cs
MaddyDev: tested and it creates separate leases table.

src/TriggerBinding/SqlTriggerBinding.cs
```diff
@@ -107,7 +107,7 @@ private string GetUserFunctionId()
             string websiteName = SqlBindingUtilities.GetWebSiteName(this._configuration);
 
             var methodInfo = (MethodInfo)this._parameter.Member;
-            string functionName = $"{methodInfo.DeclaringType.FullName}.{methodInfo.Name}";
+            string functionName = $"{methodInfo.DeclaringType.Name}.{methodInfo.Name}";
```
vivekjilla, line 110: Based on the discussion in other comments, shouldn't this be just `methodInfo.DeclaringType.Name`. Why're we appending `methodInfo.Name` as well now? I think the ScaleController wouldn't have the info around `methodInfo.Name`

MaddyDev: `methodInfo.DeclaringType.Name` is the className of the function and not the function name as I assumed, so if a class has two functions against the same table then that would only trigger one function (one of our tests was set up that way and that's how I caught it). @AmeyaRele Could you confirm if you can access method name on ScaleController end?

MaddyDev: @vivekjilla I just synced with @AmeyaRele  and updated the code to read the function name from `FunctionName` attribute. Will merge mu changes in once the tests pass and get on with the releasing the preview package tomorrow.

vivekjilla: Thank you!

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -280,7 +280,7 @@ steps:
 - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
   displayName: 'Component Detection'
   inputs:
-    failOnAlert: true
+    failOnAlert: false
```
Charles-Gagnon, line 283: I'd suggest doing this as a separate PR so we can get this fixed asap

PR #1098, opened 2024-07-16
Fix the Component Governance issue on System.Text.Json

MaddyDev:
https://msdata.visualstudio.com/SQLToolsAndLibraries/_componentGovernance/azure-functions-sql-extension/alert/11673675?typeId=23606141&pipelinesTrackingFilter=0

Pin System.Text.Json to 8.0.4 adding it as an explicit dependency since all the root dependencies are up to date.


Charles-Gagnon:
@MaddyDev The CG alert still firing should be fixed - you needed to add the package reference to the projects themselves.

But I don't think we can even do this currently - it's a major version bump for the extension dependencies. Will need to consult with the host team on whether this one will be allowed.

MaddyDev:
Confirmed with Component Governance support (cg-support@microsoft.com) that the version we have is not vulnerable and they removed the alert. Closing this as no longer required.

PR #1103, opened 2024-07-23
Add additional trigger debug logging

Charles-Gagnon:
This will help us and users diagnose issues with the trigger function - especially for being able to track when the change version we're looking at changes so we can see what change version the function is looking at for a given point in time.

New messages:

* PreInvocation updated LastSyncVersion
`[PreInvocation] Updated LastSyncVersion from 0 to 121`
* PostInvocation updated LastSyncVersion
`[PostInvocation] Updated LastSyncVersion from 153 to 155 MaxAttemptRowsToBeDeleted=1`
* When leases are renewed for lease locked rows
`Renewed leases for 1 rows`
* When the internal state is cleared (usually due to a previous error)
`Clearing internal state for 1 rows`

Updated messages:

* Added max attempts reached count to the lease locked row count check
`Executed GetChangesCommand in GetTableChangesAsync. 0 available changed rows. (0 found with lease locks and 1 ignored because they've reached the max attempt limit)`

### DIFF
--- a/src/Telemetry/Telemetry.cs
+++ b/src/Telemetry/Telemetry.cs
@@ -394,7 +394,6 @@ To learn more about our Privacy Statement visit this link: https://go.microsoft.
         GetColumnDefinitionsDurationMs,
         GetPrimaryKeysDurationMs,
         GetUnprocessedChangesDurationMs,
-        GetLockedRowCountDurationMs,
         InsertGlobalStateTableRowDurationMs,
         MaxBatchSize,
         MaxChangesPerWorker,
@@ -441,7 +440,7 @@ To learn more about our Privacy Statement visit this link: https://go.microsoft.
         Upsert,
         UpsertRollback,
         GetServerTelemetryProperties,
-        GetLeaseLockedRowCount,
+        GetLeaseLockedOrMaxAttemptRowCount,
     }
 
     internal class ServerProperties
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -289,7 +289,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                         using (SqlCommand updateTablesPreInvocationCommand = this.BuildUpdateTablesPreInvocation(connection, transaction))
                         {
                             var commandSw = Stopwatch.StartNew();
-                            await updateTablesPreInvocationCommand.ExecuteNonQueryAsyncWithLogging(this._logger, token, true);
+                            object result = await updateTablesPreInvocationCommand.ExecuteScalarAsyncWithLogging(this._logger, token, true);
+                            if (result != null)
+                            {
+                                // If we updated the LastSyncVersion we'll get a message back from the query, so log it here
+                                this._logger.LogDebug($"[PreInvocation] {result}");
+                            }
                             setLastSyncVersionDurationMs = commandSw.ElapsedMilliseconds;
                         }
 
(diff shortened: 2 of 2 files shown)


src/Telemetry/Telemetry.cs
```diff
@@ -394,7 +394,6 @@ public enum TelemetryMeasureName
         GetColumnDefinitionsDurationMs,
         GetPrimaryKeysDurationMs,
         GetUnprocessedChangesDurationMs,
-        GetLockedRowCountDurationMs,
```
Charles-Gagnon, line 397: The duration wasn't being calculated correctly so I just removed to keep things simpler. We haven't really used these durations for anything useful so far anyways - we can always add it back in if needed.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -842,44 +860,53 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+            string getLeaseLockedOrMaxAttemptRowCountQuery = $@"
                 {AppLockStatements}
 
                 DECLARE @last_sync_version bigint;
                 SELECT @last_sync_version = LastSyncVersion
                 FROM {GlobalStateTableName}
                 WHERE UserFunctionID = '{this._userFunctionId}' AND UserTableID = {this._userTableId};
-
-                SELECT COUNT(*)
+                DECLARE @lease_locked_count int;
```
Charles-Gagnon, line 888: Refactored this to just create the message in the query itself. I could have returned a table with the numeric values, but having to read the data table and do all that extra stuff didn't seem worth it when we can just do it directly in the T-SQL.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -842,44 +860,53 @@ private SqlCommand BuildGetChangesCommand(SqlConnection connection, SqlTransacti
...
+                BEGIN
+                    SELECT '(' + CAST(@lease_locked_count AS NVARCHAR) + ' found with lease locks and ' + CAST(@max_attempts_count AS NVARCHAR) + ' ignored because they''ve reached the max attempt limit)';
+                END";
             try
             {
-                using (var getLeaseLockedRowCountCommand = new SqlCommand(getLeaseLockedrowCountQuery, connection, transaction))
+                using (var getLeaseLockedOrMaxAttemptsRowCountCommand = new SqlCommand(getLeaseLockedOrMaxAttemptRowCountQuery, connection, transaction))
                 {
-                    var commandSw = Stopwatch.StartNew();
-                    leaseLockedRowsCount = (int)await getLeaseLockedRowCountCommand.ExecuteScalarAsyncWithLogging(this._logger, CancellationToken.None);
```
Charles-Gagnon, line 875: Note I fixed an issue here with not passing in the cancellation token. We should be doing that for all queries we execute so that if the function execution is cancelled we'll properly cancel the queries we're running right away.

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -952,7 +978,10 @@ UPDATE SET
         private SqlCommand BuildRenewLeasesCommand(SqlConnection connection, SqlTransaction transaction)
         {
             string matchCondition = string.Join(" OR ", this._rowMatchConditions.Take(this._rowsToProcess.Count));
-
+            if (string.IsNullOrEmpty(matchCondition))
+            {
+                this._logger.LogError($"MatchCondition resolved to empty with '{this._rowsToProcess.Count}' rowsToProcess.");
```
Charles-Gagnon, line 983: @MaddyDev It wouldn't be a bad idea to just skip executing the query in this case as well - we know it's going to fail so no reason to add extra load to the server in that scenario.

If you'd rather not do that in this PR then I'd suggest filing a tracking issue for follow up

MaddyDev: Created https://github.com/Azure/azure-functions-sql-extension/issues/1108

MaddyDev (approved):

PR #1110, opened 2024-07-24
Skip empty where clause execution on trigger

### LINKED ISSUE #1108: Failed to renew leases due to exception: Exception message: Incorrect syntax near ';'.
The SQL trigger generates invalid T-SQL intermittently while renewing leases, causing false alerts in the monitoring system. The issue does not affect the functionality of the SQL Trigger. 
SQL trigger raising Microsoft.Data.SqlClient.SqlException. Exception message: Incorrect syntax near \';\'

Below are the logs from the system:

Exception executing query. Message=Incorrect syntax near ';'.
Query=
DECLARE @result int;
EXEC @result = sp_getapplock @Resource = '_az_func_Trigger',
@LockMode = 'Exclusive',
@LockTimeout = 30000
IF @result < 0
BEGIN
RAISERROR('Unable to acquire exclusive lock on _az_func_Trigger. Result = %d', 16, 1, @result)
END;

UPDATE [az_func].[Leases_cacf2bceb5090ff5_1845581613]
SET _az_func_LeaseExpirationTime = DATEADD(second, 60, SYSDATETIME())
WHERE ;

Refer [Icm](https://portal.microsofticm.com/imp/v5/incidents/details/517530853/summary) for more information.

MaddyDev:
Return null when match condition resolves to empty instead of sending in the empty where clause.

Fixes #1108

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -525,28 +525,30 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
                 {
                     try
                     {
-                        using (SqlCommand renewLeasesCommand = this.BuildRenewLeasesCommand(connection, transaction))
+                        SqlCommand renewLeasesCommand = this.BuildRenewLeasesCommand(connection, transaction);
+                        if (renewLeasesCommand != null)
                         {
-                            var stopwatch = Stopwatch.StartNew();
-
-                            int rowsAffected = await renewLeasesCommand.ExecuteNonQueryAsyncWithLogging(this._logger, token, true);
-
-                            long durationMs = stopwatch.ElapsedMilliseconds;
-
-                            if (rowsAffected > 0)
+                            using (renewLeasesCommand)
                             {
-                                this._logger.LogDebug($"Renewed leases for {rowsAffected} rows");
-                                // Only send an event if we actually updated rows to reduce the overall number of events we send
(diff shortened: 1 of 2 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -981,6 +983,7 @@ private SqlCommand BuildRenewLeasesCommand(SqlConnection connection, SqlTransact
             if (string.IsNullOrEmpty(matchCondition))
             {
                 this._logger.LogError($"MatchCondition resolved to empty with '{this._rowsToProcess.Count}' rowsToProcess.");
+                return null;
```
Charles-Gagnon, line 986: suggestion: Add comment explaining why we want to return null in this case

src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -981,6 +983,7 @@ private SqlCommand BuildRenewLeasesCommand(SqlConnection connection, SqlTransact
             if (string.IsNullOrEmpty(matchCondition))
             {
                 this._logger.LogError($"MatchCondition resolved to empty with '{this._rowsToProcess.Count}' rowsToProcess.");
```
Charles-Gagnon, line 985: Do we want to add a telemetry event for when this occurs?

Charles-Gagnon (approved):

chlafreniere (approved):

PR #1146, opened 2024-11-13
Update Npm installs to use local feed

MaddyDev:
Following TSG: https://eng.ms/docs/cloud-ai-platform/azure-edge-platform-aep/aep-engineering-systems/productivity-and-experiences/network-isolation/oss/npm

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,3 @@
+registry=https://msdata.pkgs.visualstudio.com/SQLToolsAndLibraries/_packaging/SQLToolsAndLibraries_PublicPackages/npm/registry/
+
+always-auth=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
-registry=https://msdata.pkgs.visualstudio.com/SQLToolsAndLibraries/_packaging/SQLToolsAndLibraries_PublicPackages/npm/registry/
+registry=https://registry.npmjs.org
 
 always-auth=true
--- a/builds/azure-pipelines/template-steps-build-test.yml
+++ b/builds/azure-pipelines/template-steps-build-test.yml
@@ -19,10 +19,15 @@ steps:
     packageType: sdk
     version: '3.1.x'
 
-- script: npm install -g azure-functions-core-tools
-  displayName: 'Install Azure Functions Core Tools'
+- task: npmAuthenticate@0
+  displayName: 'npm Authenticate for project'
+  inputs:
+    workingFile: $(Build.SourcesDirectory)/.npmrc
 
-- script: npm install -g azurite
+- script: npm install azure-functions-core-tools --global --globalconfig $(Build.SourcesDirectory)/.npmrc --loglevel verbose
+  isplayName: 'Install Azure Functions Core Tools'
+
+- script: npm install azurite --global --globalconfig $(Build.SourcesDirectory)/.npmrc --loglevel verbose
   displayName: 'Install Azurite Local Storage Emulator'
 
 # This step is necessary because npm installs to a non-traditional location on Windows hosted agents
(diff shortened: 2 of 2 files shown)


builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -60,6 +65,11 @@ steps:
   displayName: Set logging level
   condition: and(succeeded(), ne(variables['AFSQLEXT_TEST_LOGLEVEL'], ''))
 
+- task: npmAuthenticate@0
+  displayName: 'npm Authenticate for project'
+  inputs:
+  workingFile: $(Build.SourcesDirectory)/.npmrc
+  
 - script: |
     npm install
```
Charles-Gagnon, line 74: What about this? (and other ones below)

builds/azure-pipelines/template-steps-build-test.yml
```diff
@@ -19,10 +19,15 @@ steps:
     packageType: sdk
     version: '3.1.x'
 
-- script: npm install -g azure-functions-core-tools
+- task: npmAuthenticate@0
```
Charles-Gagnon, line 22: Make sure you check with a buddy build that these are pulling from the right npmrc file in our ADO pipelines.

builds/azure-pipelines/template-steps-build-test.yml
Charles-Gagnon: Oh right, nevermind. Local installs pick up the local config. It's only the global ones that need to specify the config directly.

MaddyDev: added this las week and forgot to clean up, I just cleaned up and added userconfig for the below ones.

builds/azure-pipelines/template-steps-build-test.yml
MaddyDev: yes, am going to test with a buddy build.

Charles-Gagnon:
FYI - for the PR builds to work you're going to need to add a default npmrc file here pointing at npmjs

builds/azure-pipelines/template-steps-build-test.yml
MaddyDev: What I understood is, when using as a per project config, it needs to be a sibling of the package.json but when you're specifying it as userconfig, I think you can give the path to the file. But am not 100% sure either. Do you think if I should be adding a .npmrc file here as well?

Charles-Gagnon: So the issue we have is that we have multiple projects to build (e.g. samples-js and test-js). So in order for it to pick up the npmrc directly we'd have to have one in each of those folders - it doesn't seem to contain any logic to automatically look in parent folders unfortunately.

What you have seems fine - just setting it at the user level will cause it to use that instead. But yes, you will need to add an npmrc file in order for this to successfully build.

.npmrc
```diff
@@ -0,0 +1,3 @@
+registry=https://msdata.pkgs.visualstudio.com/SQLToolsAndLibraries/_packaging/SQLToolsAndLibraries_PublicPackages/npm/registry/
```
Charles-Gagnon, line 1: This needs to point to npmjs.org, NOT our private feed

Charles-Gagnon, line 1: This needs to be pointing to the public registry, NOT our private feed (we shouldn't have anything about our private feeds in this repo)

```suggestion
registry=https://registry.npmjs.org
```

Charles-Gagnon: It's fine (and expected) for our PR build to use the public registries. It's only the ADO pipelines that need to use the private stuff.

Charles-Gagnon (approved):

PR #1160, opened 2025-01-14
Bump Microsoft.Data.SqlClient, Azure.Core, Azure.Identity and *.Storage packages

### LINKED ISSUE #1007: Bump Microsoft.Azure.WebJobs.Extensions.Storage and Azure.Identity packages to latest
Dependabot PR : https://github.com/Azure/azure-functions-sql-extension/pull/994

Doing this results in a major version update of the Microsoft.Identity.Client.Extensions.Msal package (2.31.0 -> 4.56.0) so we need to verify that such an update is safe and won't conflict with the host before taking this in.

### LINKED ISSUE #1066: Bump Microsoft.Data.SqlClient and Azure.Identity to latest
Dependabot PR : https://github.com/Azure/azure-functions-sql-extension/pull/1065

Doing this results in a major version update of the Microsoft.Identity.Client.Extensions.Msal package (2.31.0 -> 4.56.0) so we need to verify that such an update is safe and won't conflict with the host before taking this in.

Charles-Gagnon:
Closes #1066 
Closes #1007

Azure.Core 1.35.0 (transitive) -> 1.41.0
Azure.Identity 1.10.2 -> 1.12.1
Microsoft.Data.SqlClient 5.1.3 -> 5.2.2
Microsoft.Azure.WebJobs.Extensions.Storage 5.1.3 -> 5.3.2
Microsoft.Azure.WebJobs.Host.Storage 5.0.0 -> 5.0.1

This update includes a major-version update :

Microsoft.Identity.Client.Extensions.Msal 2.31.0 -> 4.66.1

After discussing with the Azure Functions core team we've decided that in this case this is safe to do - the package is marked as private in the [host runtime assemblies](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/runtimeassemblies.json), and the update itself is low risk of causing other issues with existing functions.

Note that Azure.Identity was NOT bumped to the very latest (1.13.1) since that includes a number of other transitive major version bumps that will need to be investigated further.

Azure.Core was updated to address a security issue in that library until Azure.Identity can be updated to the latest.


Directory.Packages.props
```diff
@@ -37,8 +37,8 @@
             1. Verifying with the core Azure Functions host team that there will not be any issues
             2. Running the bundle dependency test and verifying it passes
      -->
-
-    <PackageVersion Include="Azure.Identity" Version="1.10.2"/>
+    <PackageVersion Include="Azure.Core" Version="1.41.0"/> <!-- Updated to address security issue, should be removed once Azure.Identity is updated-->
```
MaddyDev, line 40: Why are we adding a new library to address security issue? Will it be okay if we remove it after Azure.Identity is updated?

Charles-Gagnon: We are not adding it (it was already brought in as a transitive dependency)

This is just ensuring that we get the correct version with the fix for the issue, since the transitive dependency brings in an older version.

MaddyDev (approved):

chlafreniere (approved):

PR #1165, opened 2025-01-28
fix #pragma warning CA2327

MaddyDev:
Fix the type safe handling security issue created: 
image

### DIFF
--- a/test/Integration/test-csharp/Startup.cs
+++ b/test/Integration/test-csharp/Startup.cs
@@ -13,17 +13,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql.Tests.Integration
     {
         public override void Configure(IFunctionsHostBuilder builder)
         {
-#pragma warning disable CA2327 // Test code, safe to ignore
+            // Test code, safe to ignore
             // Set default settings for JsonConvert to simulate a user doing the same in their function.
             // This will cause test failures if serialization/deserialization isn't done correctly
             // (using the helper methods in Utils.cs)
-            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
+            var settings = new JsonSerializerSettings
             {
                 Formatting = Formatting.Indented,
                 TypeNameHandling = TypeNameHandling.Objects,
                 ContractResolver = new CamelCasePropertyNamesContractResolver()
             };
-#pragma warning restore CA2327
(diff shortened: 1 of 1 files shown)


test/Integration/test-csharp/Startup.cs
```diff
@@ -13,17 +13,17 @@ public class Startup : FunctionsStartup
     {
         public override void Configure(IFunctionsHostBuilder builder)
         {
-#pragma warning disable CA2327 // Test code, safe to ignore
+            // Test code, safe to ignore
             // Set default settings for JsonConvert to simulate a user doing the same in their function.
             // This will cause test failures if serialization/deserialization isn't done correctly
             // (using the helper methods in Utils.cs)
-            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
+            var settings = new JsonSerializerSettings
```
Charles-Gagnon, line 20: How is this fixing the issue? You're only slightly changing how it's initialized (creating one instance that's always returned vs creating new instances each time), but none of the properties that https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2327 says are an issue seem to be changed.

MaddyDev: I might have mis worded by saying fix CA2327, what am trying to address is here is the type safe handling which incurred us a S360 Codeql violation. Here the JsonSerializerSettings instance is created and assigned to a variable setting, which is then used to set JsonConvert.DefaultSettings. This approach ensures type safety by explicitly defining the type of the settings variable.

MaddyDev:
I ran the codeql locally and did not see any issues prop up, we're good here.

test/Integration/test-csharp/Startup.cs
```diff
@@ -13,17 +13,17 @@ public class Startup : FunctionsStartup
     {
         public override void Configure(IFunctionsHostBuilder builder)
         {
-#pragma warning disable CA2327 // Test code, safe to ignore
+            // Test code, safe to ignore
```
Charles-Gagnon, line 16: Delete this - it was added as a reason for why the warning was disabled

```suggestion
```

VasuBhog (approved):

Charles-Gagnon (approved):

chlafreniere (approved):

Issue #1171, opened 2025-03-27
Current version is incompatible with latest Microsoft.Azure.Functions.Worker.Sdk 2.0.2


phelm1-landolakes:
- Azure Functions SQL Extension or Extension Bundle Version: 3.1.376
- Is this a deployed or local function: na
- What type of Database are you using? na

Steps to Reproduce:

1. Upgrade Microsoft.Azure.Functions.Worker.Sdk 2.0.1 => 2.0.2
2. Attempt to Build 

The build will fail as Microsoft.Azure.Functions.Worker.Sdk is now Net8.0. 

`obj\Debug\net8.0\WorkerExtensions\WorkerExtensions.csproj : error NU1605: Warning As Error: Detected package downgrade: System.Runtime.Caching from 8.0.0 to 6.0.1. Reference the package directly from the project to select a different version. 

obj\Debug\net8.0\WorkerExtensions\WorkerExtensions.csproj : error NU1605:  Microsoft.Azure.Functions.Worker.Extensions -> Microsoft.Azure.WebJobs.Extensions.Sql 3.1.376 -> Microsoft.Data.SqlClient 5.2.2 -> System.Runtime.Caching (>= 8.0.0) 

obj\Debug\net8.0\WorkerExtensions\WorkerExtensions.csproj : error NU1605:  Microsoft.Azure.Functions.Worker.Extensions -> Microsoft.Azure.WebJobs.Extensions.Sql 3.1.376 -> System.Runtime.Caching (>= 6.0.1)
`

zmilleson:
I am also getting this when I try to build locally or using an Azure Pipelines agent.

"Microsoft.Azure.Functions.Worker.Extensions.Sql" Version="3.1.376"
Function Core Tools version: 4.0.7030


\obj\Debug\net8.0\WorkerExtensions\WorkerExtensions.csproj : error NU1605: Warning As Error: Detected package downgrade:
 System.Runtime.Caching from 8.0.0 to 6.0.1. Reference the package directly from the project to select a different version. 

\obj\Debug\net8.0\WorkerExtensions\WorkerExtensions.csproj : error NU1605:  Microsoft.Azure.Functions.Worker.Extensions  
-> Microsoft.Azure.WebJobs.Extensions.Sql 3.1.376 -> Microsoft.Data.SqlClient 5.2.2 -> System.Runtime.Caching (>= 8.0.0)

[...]

PR #1173, opened 2025-04-01
Update Microsoft.Azure.Functions.Worker.Sdk package

VasuBhog:
This change began due to #1171. The user wants to update to 2.0.2 (but that is on .Net 8) which we currently do not support.

Updating to https://github.com/Azure/azure-functions-dotnet-worker/releases/tag/sdk-2.0.1

If we did 2.0.2 then we get errors below:
```
C:\Program Files\dotnet\sdk\6.0.411\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(144,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 8.0.  Either target .NET 6.0 or lower, or use  
a version of the .NET SDK that supports .NET 8.0. [C:\Git\azfnsql\azure-functions-sql-extension\samples\samples-outofproc\obj\Debug\net6\WorkerExtensions\WorkerExtensions.csproj]
C:\Program Files\dotnet\sdk\6.0.411\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(144,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 8.0.  Either target .NET 6.0 or lower, or use  
a version of the .NET SDK that supports .NET 8.0. [C:\Git\azfnsql\azure-functions-sql-extension\test-outofproc\obj\Debug\net6\WorkerExtensions\WorkerExtensions.csproj]
```

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -11,7 +11,7 @@
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.3.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Storage.Queues" Version="5.5.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Timer" Version="4.3.1" />
-    <PackageVersion Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.18.1" />
+    <PackageVersion Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.1" />
     <PackageVersion Include="Microsoft.Azure.WebJobs" Version="3.0.41" />
     <PackageVersion Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.3.2" />
     <PackageVersion Include="Microsoft.Azure.WebJobs.Host.Storage" Version="5.0.1" />
--- a/samples/samples-outofproc/packages.lock.json
+++ b/samples/samples-outofproc/packages.lock.json
@@ -71,11 +71,11 @@
       },
       "Microsoft.Azure.Functions.Worker.Sdk": {
         "type": "Direct",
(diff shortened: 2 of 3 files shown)


Charles-Gagnon:
How is this addressing the linked issue at all? This package is only used by our samples/tests, so has no impact on our released package.

As your comment and the issue say - the problem there is that we currently [pin System.Runtime.Caching](https://github.com/Azure/azure-functions-sql-extension/blob/9d7be1551cd3a5300f0aef5c887f6da5d9eda4ea/Directory.Packages.props#L33) to 6.0.1, which isn't compatible with the net8 version of SqlClient. So we'd need to update that in order to resolve the linked issue.

Not that we shouldn't do this of course, but just want to make sure I'm not missing something here.

Charles-Gagnon:
Also - I'm curious why dependabot didn't inform us of this update...it has in the past. Might be worth digging in to so we validate that we aren't missing any other updates.

Charles-Gagnon (approved):

VasuBhog:
> How is this addressing the linked issue at all? This package is only used by our samples/tests, so has no impact on our released package.
> 
> As your comment and the issue show - the problem there is that we currently [pin System.Runtime.Caching](https://github.com/Azure/azure-functions-sql-extension/blob/9d7be1551cd3a5300f0aef5c887f6da5d9eda4ea/Directory.Packages.props#L33) to 6.0.1, which isn't compatible with the [net8 version of SqlClient](https://www.nuget.org/packages/Microsoft.Data.SqlClient/5.2.2#dependencies-body-tab). So we'd need to update that in order to resolve the linked issue.
> 
> Not that we shouldn't do this of course, but just want to make sure I'm not missing something here.

No you are right, I started the investigation only because of that issue so updated the description to reflect that. I can look at updating the `System.Runtime.Caching` for fix the linked issue.

PR #1174, opened 2025-04-08
Update packages for .NET8

Charles-Gagnon:
Mostly a fix for #1171 

We were pinning the 6.* versions of System.Runtime.Caching, but with the host and SDK moving to .NET 8, we need to update to the 8.* versions to avoid build/runtime errors due to SqlClient needing at least 8.* when targeting net8

Doing this also caused a few other transitive dependencies to be updated : 

System.Configuration.ConfigurationManager: 6.0.2 -> 8.0.1
System.Memory 4.5.4 -> 4.5.5
System.Security.Cryptography.ProtectedData: 6.0.0 -> 8.0.0

This update also caused System.Security.AccessControl to revert back to 5.* because the transitive dependency that was bringing in 6.* was removed (and other transitive dependencies only ask for 5.*). So to avoid any regressions there I'm also pinning that to 6.0.1 to keep it the same as it was previously.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -29,6 +29,7 @@
     <PackageVersion Include="Microsoft.SqlServer.TransactSql.ScriptDom" Version="161.9135.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="System.Runtime.Caching" Version="8.0.1" />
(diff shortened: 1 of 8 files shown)


chlafreniere:
@Charles-Gagnon looks like there are some build failures

MaddyDev (approved):

PR #1180, opened 2025-05-06
Remove morelinq

### LINKED ISSUE #74: Investigate Morelinq Dependency
There's a dependency on Morelinq, but it's only leveraged once on one line: https://github.com/Azure/azure-functions-sql-extension/blob/ad352485aad33a10d457d550fc80a0475d3a93fb/src/SqlBinding/SqlAsyncCollector.cs#L148

(batch docs can be found here: https://morelinq.github.io/3.1/ref/api/html/Overload_MoreLinq_MoreEnumerable_Batch.htm)

If it significantly decreases the amount of code that we'd need to write (and we couldn't leverage another library), then it's not a huge deal to keep it, but just wanted to file an issue to keep track of this dependency.

### LINKED ISSUE #1177: Bundle Version 4.23.0 breaks SQL Output bindings
If you have a non-.NET function with a SQL output binding, and the bundle version used is 4.23.0 then you will see failures similar to this 

```
Exception while executing function
 /Functions.surewx_hot_persist_trigger ---> System.InvalidOperationException 
 /Error while handling parameter _binder after function returned
 /---> System.MissingMethodException 
 /Method not found
 /'System.Collections.Generic.IEnumerable`1<System.Collections.Generic.IEnumerable`1<!!0>> MoreLinq.MoreEnumerable.Batch(System.Collections.Generic.IEnumerable`1<!!0>, Int32)'.   
```

This is due to a bumped MoreLinq dependency in the bundle, which contains a [breaking change](https://github.com/morelinq/MoreLINQ/releases/tag/v4.0.0) to the Batch method used in the output binding.

**If you hit this, the current workaround is to pin your bundle version to 4.22.0**

In host.json

```json
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[4.22.0, 4.23.0)"
  }
```

Charles-Gagnon:
Closes #74 
Fixes #1177 

The functionality used from Morelinq is pretty simplistic, so just writing our own versions to avoid compat issues with the MySql binding.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -25,7 +25,6 @@
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.11.0" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
     <PackageVersion Include="Moq" Version="4.20.72" />
-    <PackageVersion Include="morelinq" Version="3.4.2" />
     <PackageVersion Include="Microsoft.SqlServer.TransactSql.ScriptDom" Version="161.9135.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="System.Runtime.Caching" Version="8.0.1" />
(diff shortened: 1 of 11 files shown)


Charles-Gagnon:
Lots of test failures....will investigate

MaddyDev (approved):

chlafreniere (approved):

PR #1186, opened 2025-06-17
Add the trigger tables creation on scale provider initialization

MaddyDev:
# Description

## Added Async Initialization Support for SQL Scaler Provider

This pull request enhances the `SqlScalerProvider` by adding proper asynchronous initialization capabilities to ensure trigger metadata tables are available for any scaling SKU, even when the host does not start the listener.

### Changes

1. **Added `InitializeAsync` method to `SqlScalerProvider`**
   - Properly separates asynchronous initialization logic from constructor
   - Ensures trigger tables are created before scaling operations occur
   - Maintains clean separation of concerns for resource initialization

2. **Implemented `CreateAsync` static factory method**
   - Provides a standard pattern for asynchronous object creation
   - Ensures all resources are properly initialized before the provider is used

3. **Modified service registration in `SqlBindingExtension`**
   - Updated `AddSqlScaleForTrigger` to use the new async initialization pattern
   - Used `Lazy<Task<SqlScalerProvider>>` to ensure initialization happens only once

## Benefits

- **Improved reliability**: Ensures trigger metadata tables exist before scaling decisions are made
- **Better scaling support**: Works with all hosting SKUs regardless of listener initialization state
- **Reduced race conditions**: Properly initializes resources before they are accessed

This implementation ensures the SQL binding's scaling capabilities work reliably across all Azure Functions hosting scenarios.

# Code Changes

- [x] [Unit tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Unit) are added, if possible
- [x] [Integration tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Integration)  are added if the change is modifying existing behavior of one or more of the bindings

# Dependencies

# Documentation

- [x] Add [samples](https://github.com/Azure/azure-functions-sql-extension/tree/main/samples) if the change is modifying or adding functionality
- [x] Update relevant documentation in the [docs](https://github.com/Azure/azure-functions-sql-extension/tree/main/docs)


chlafreniere:
@MaddyDev plz resolve merge conflicts

src/TriggerBinding/SqlScalerProvider.cs
```diff
@@ -55,6 +75,59 @@ public ITargetScaler GetTargetScaler()
...
+        public async Task InitializeAsync(CancellationToken cancellationToken = default)
+        {
+            using (var connection = new SqlConnection(this._connectionString))
+            {
+                await connection.OpenAsyncWithSqlErrorHandling(cancellationToken);
+                ServerProperties serverProperties = await GetServerTelemetryProperties(connection, this._logger, cancellationToken);
+                this._telemetryProps.AddConnectionProps(connection, serverProperties);
+                await VerifyDatabaseSupported(connection, this._logger, cancellationToken);
+
+                int userTableId = await GetUserTableIdAsync(connection, this._userTable, this._logger, cancellationToken);
```
Charles-Gagnon, line 95: We had discussed NOT doing this here and instead having the scaler return the total count of all changes if the tables don't exist. This would cause it to scale up, which would create the actual instance that would then create the tables as normal.

Is there a reason you decided not to do that? I'm not thrilled about having table creation happen in multiple places in our code. It makes it more likely that if we ever make any changes to how they're created that we'll miss updating one or the other and cause problems.

alrod: Yeah, it would also work.

You need to catch an exception here (schema does not exist, GlobalStateTable does not exists or GlobalStateTableRow does not exists):
https://github.com/Azure/azure-functions-sql-extension/blob/main/src/TriggerBinding/SqlTriggerTargetScaler.cs#L29

and return:
            return new TargetScalerResult
            {
                TargetWorkerCount = 1
            };
Then a worker will be assigned, and listener will create required stuff.

alrod: I believe it’s better for the SC to attempt creating the GlobalState, since we only try once.
Otherwise, we might end up with worker 1 being permanently assigned, as the SC will keep voting for it if GlobalState creation fails for any reason.

Charles-Gagnon: If the worker gets in a bad state (e.g. if global state table creation fails in the worker it will throw an exception and exit) will the scaler make a new one? If so then keeping 1 worker around seems like the behavior we want then - since it'll just keep retrying to make the table until it succeeds - which we need for the trigger to work correctly.

alrod: we do not want to keep a worker around just waiting until the issue with creation is fixed as it would affect customer's bill

src/TriggerBinding/SqlTriggerUtils.cs
```diff
@@ -126,5 +129,263 @@ internal static string GetBracketedLeasesTableName(string userDefinedLeasesTable
             return string.IsNullOrEmpty(userDefinedLeasesTableName) ? string.Format(CultureInfo.InvariantCulture, LeasesTableNameFormat, $"{userFunctionId}_{userTableId}") :
                 string.Format(CultureInfo.InvariantCulture, UserDefinedLeasesTableNameFormat, $"{userDefinedLeasesTableName.AsBracketQuotedString()}");
         }
+
```
alrod, line 132: Do we also want to add a method to `SqlTriggerUtils` which will create everything in once?

                    using (SqlTransaction transaction = connection.BeginTransaction(System.Data.IsolationLevel.RepeatableRead))
                    {
                        createdSchemaDurationMs = await this.CreateSchemaAsync(connection, transaction, cancellationToken);
                        createGlobalStateTableDurationMs = await this.CreateGlobalStateTableAsync(connection, transaction, cancellationToken);
                        insertGlobalStateTableRowDurationMs = await this.InsertGlobalStateTableRowAsync(connection, transaction, userTableId, cancellationToken);
                        createLeasesTableDurationMs = await this.CreateLeasesTableAsync(connection, transaction, bracketedLeasesTableName, primaryKeyColumns, cancellationToken);
                        transaction.Commit();
                    }

src/TriggerBinding/SqlScalerProvider.cs
alrod: If we want to scale out only once right after monitoring started for MSSQL trigger - yes, the bad worker will exit and will never add again until customer explicitly changed the function app. 

EventHubs have the same scenario - creating checkpoint blob to store position client position in the stream. The blob is created on SDK level by `EventProcessorClient`. There is no such thing as SQL SDK for monitoring changes, so the code lives in MSSQL azure functions extension.

alrod: In general, I am ok with scale out 1 instance but only once

MaddyDev:
Closing this since we synced and decided to check the changes and spin up a worker (which creates the tables) instead.

PR #1189, opened 2025-06-26
Spin worker on exception

### LINKED ISSUE #1184: SQL Trigger does not work when deployed to Flex Consumption
- Azure Functions SQL Extension or Extension Bundle Version: 3.1.490
- Is this a deployed or local function: deployed
- What type of Database are you using? (Run `SELECT @@VERSION as Version, SERVERPROPERTY('EngineEdition') as EngineEdition` on your database): Microsoft SQL Azure (RTM) - 12.0.2000.8 Apr 9 2025 07:13:50 Copyright (C) 2024 Microsoft Corporation | 5
- List any custom settings for your function app. This could be a custom time-out defined specifically for your database server or optional configuration that can be customized for the app defined [here](https://learn.microsoft.com/azure/azure-functions/functions-bindings-azure-sql-trigger?tabs=isolated-process%2Cportal&pivots=programming-language-csharp#optional-configuration).: none

Steps to Reproduce:

1. Create new sql trigger function from Visual Studio
2. Deploy to Flex Consumption
3. Modify the data in the target table
4. SqlTrigger function not triggered

I deployed a Function App that implements SqlTrigger to Flex Consumption and SqlTrigger is not triggered on table changes. I have confirmed that it works correctly when running Consumption or locally, but only does not work when deployed to Flex Consumption.

When deployed to Flex Consumption, lease-related tables were not automatically created. This is a possible cause of this issue.

MaddyDev:
# Description
Fixes #1184 
## Changes in GetScaleResultAsync in SqlTriggerTargetScaler

The changes to the `GetScaleResultAsync` method in SqlTriggerTargetScaler.cs involve checking if the absence of trigger tables (global state table and leases table) is causing trigger functions on Flex consumption plan to fail and to address them adding error handling logic, particularly for cases where SQL objects (tables) don't exist yet. Here's a breakdown of the changes:

### Main Changes

1. **More Robust SQL Exception Handling**:
   - The code now specifically checks for "Invalid object name" errors in both the exception message and its inner exception
   - This allows the scaler to detect when required SQL tables (global state or leases tables) haven't been created yet

2. **Added Table Check Tracking**:
   - A static `_lastTableCheck` DateTime variable tracks when the code last attempted to check tables
   - This prevents continuous retry loops and provides a way to throttle checks

3. **Proactive Change Detection**:
   - When tables are missing, the code now checks if there are changes in the user's table using `GetNumberOfChangesAsync`
   - This allows the system to determine if a worker is actually needed despite missing metadata tables

4. **Intelligent Worker Allocation**:
   - If changes exist in the user table but metadata tables are missing, it requests a worker (returns 1)
   - This worker will start the `SqlTriggerListener` that will create the necessary metadata tables
   - If no changes exist or checks are too frequent, it returns 0 workers

5. **Improved Logging**:
   - More descriptive log messages help diagnose the state of the system
   - Clear warnings when tables are missing or when falling back to default behavior

## Benefits of These Changes

1. **Prevents Scaling Loops**: Avoids continuously spinning up workers when tables don't exist
2. **Reduces Unnecessary Resources**: Returns 0 workers when no action is needed
3. **Self-Healing**: Intelligently allocates a worker when it detects changes that need processing
4. **Better Diagnostics**: Clearer logging about what's happening and why

These changes make the scaler more robust when dealing with edge cases like missing tables or partial initialization, improving the overall reliability of the SQL trigger scaling system.

# Code Changes

- [ ] [Unit tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Unit) are added, if possible
- [x] [Integration tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Integration)  are added if the change is modifying existing behavior of one or more of the bindings

# Dependencies

# Documentation

- [x] Add [samples](https://github.com/Azure/azure-functions-sql-extension/tree/main/samples) if the change is modifying or adding functionality
- [x] Update relevant documentation in the [docs](https://github.com/Azure/azure-functions-sql-extension/tree/main/docs)


src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,71 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                // Instance concurrency value is set by the functions host when dynamic concurrency is enabled. See https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency for more details.
+                int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
 
-            // Instance concurrency value is set by the functions host when dynamic concurrency is enabled. See https://learn.microsoft.com/en-us/azure/azure-functions/functions-concurrency for more details.
-            int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
+                return GetScaleResultInternal(concurrency, metrics.UnprocessedChangeCount);
+            }
+            catch (Exception ex)
+            {
+                if (ex is SqlException sqlEx && (sqlEx.Message.Contains("Invalid object name") ||
```
Charles-Gagnon, line 51: 1. Exception text may be localized, so we should never be relying on that
2. I think it would be far better to have the [T-SQL used to fetch the unprocessed changes ](https://github.com/Azure/azure-functions-sql-extension/blob/35ad03153a1e557d3a61f4a938b551ca4f083fb7/src/TriggerBinding/SqlTriggerMetricsProvider.cs#L100) work like this instead : 

```sql
IF GlobalStateTable NOT EXISTS OR LeasesTable NOT EXISTS
THEN
    SELECT COUNT_BIG(*) FROM CHANGETABLE(CHANGES {this._userTable.BracketQuotedFullName}, null)
ELSE
    -- the query as it is currently
END
```

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,71 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                        // Check if there are any changes in the user table.
+                        int changes = await GetNumberOfChangesAsync(this._connectionString, this._userTable, this._logger, CancellationToken.None);
+                        // If there are changes in the user table, we need to spin up a worker to create the global state and leases table.
+                        if (changes > 0)
+                        {
+                            this._logger.LogWarning("There are changes in the change-tracking table for the user table, but the global state and leases table are not created. Spinning up a worker to create those tables.");
+                            _lastTableCheck = DateTime.UtcNow;
+                            return new TargetScalerResult
+                            {
+                                TargetWorkerCount = 1
```
Charles-Gagnon, line 71: If the scale controller is checking every 10sec that means that we'll tell it to scale up once, but then 10sec later immediately tell it to scale down to 0 again. It's very unlikely that the tables will have been created by then, which could cause problems.

We discussed returning a target count of 1 for 1-2min to give more time for the function to be started and create the tables, so I think you'll want to modify the logic to account for that.

@alrod do you agree with that?

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,71 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                    }
 
-            return GetScaleResultInternal(concurrency, metrics.UnprocessedChangeCount);
+                }
+                // If the exception is not related to the invalid object name Or if there are no changes in the change tracking table for the user table yet.
+                // We can safely return 0 as the target worker count, indicating that no workers are needed at this time.
+                this._logger.LogWarning("An error occurred while getting the scale result for SQL trigger. Returning 0 as the target worker count. Exception: {ExceptionMessage}", ex.Message);
+                return new TargetScalerResult
+                {
+                    TargetWorkerCount = 0
```
Charles-Gagnon, line 82: @alrod If an exception occurs while getting the metrics count would it be better to let the exception flow back to the scale controller? Or is having it scale down to 0 until the issue is resolved expected behavior?

src/TriggerBinding/SqlTriggerTargetScaler.cs
MaddyDev: good catch! Thanks will look for the error code.

src/TriggerBinding/SqlTriggerTargetScaler.cs
MaddyDev: I remember it being every minute and I thought that's time enough - but I could add threshold time to the check.

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,80 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
-            int concurrency = context.InstanceConcurrency ?? this._maxChangesPerWorker;
+                return GetScaleResultInternal(concurrency, metrics.UnprocessedChangeCount);
+            }
+            catch (Exception ex)
+            {
+                // If the exception is SQL exception and indicates that the object name is invalid, it means that the global state and leases table are not created
+                // Check for the error number 208 or the error message https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/mssqlserver-208-database-engine-error?view=sql-server-ver17
+                if (ex is SqlException sqlEx && (sqlEx.Number == 208 || sqlEx.Message.Contains("Invalid object name") ||
+                                   (sqlEx.InnerException != null &&
+                                    sqlEx.InnerException.Message.Contains("Invalid object name"))))
```
Charles-Gagnon, line 55: Is there a reason to keep the message checking? That seems risky - we have no idea what the errors from the engine could contain - it's possible (albeit unlikely) that some other random error or object name could happen to contain the string "invalid object name" and cause problems.

src/TriggerBinding/SqlTriggerTargetScaler.cs
alrod: yeah, we should throw the exception back to SC here.

Charles-Gagnon (changes requested):

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,75 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                    }
+                    else
+                    {
+                        // Check if there are any changes in the user table.
+                        int changes = await GetNumberOfChangesAsync(this._connectionString, this._userTable, this._logger, CancellationToken.None);
+                        // If there are changes in the user table, we need to spin up a worker to create the global state and leases table.
+                        if (changes > 0)
+                        {
+                            this._logger.LogWarning("There are changes in the change-tracking table for the user table, but the global state and leases table are not created. Spinning up a worker to create those tables.");
+                            _lastTableCheck = DateTime.UtcNow;
```
Charles-Gagnon, line 75: We should only set this once, and consider renaming it something like `_firstTableCreationWarmupAttempt`

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,75 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                        this._logger.LogWarning("Returning 0 as the target worker count since we have already checked for the SQL trigger tables within the last 2 minutes.");
+                        return new TargetScalerResult
+                        {
+                            TargetWorkerCount = 0
+                        };
+                    }
+                    else
+                    {
+                        // Check if there are any changes in the user table.
+                        int changes = await GetNumberOfChangesAsync(this._connectionString, this._userTable, this._logger, CancellationToken.None);
```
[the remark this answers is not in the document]
MaddyDev: No

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,74 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                        // If we have already checked within the last 2 minutes, we will return 0 as the target worker count.
+                        this._logger.LogWarning("Returning 0 as the target worker count since the GetMetrics query threw an 'Invalid object name detected' error and we've exceeded the warmup period for scaling up a new instance to create the required state tables.");
+                        return new TargetScalerResult
+                        {
+                            TargetWorkerCount = 0
+                        };
+                    }
+                    else
+                    {
+                        // Check if there are any changes in the user table.
```
Charles-Gagnon, line 69:
```suggestion
                        // Check if there are any changes in the user table. Since we don't have a leases table that means
                        // we haven't processed any of the changes yet so we can just check if there's any changes
                        // for the table at all (no last sync point)
```

src/TriggerBinding/SqlTriggerTargetScaler.cs
```diff
@@ -14,24 +17,74 @@ internal sealed class SqlTriggerTargetScaler : ITargetScaler
...
+                        int changes = await GetChangeCountFromChangeTrackingAsync(this._connectionString, this._userTable, this._logger, CancellationToken.None);
+                        // If there are changes in the user table, we need to spin up a worker to create the global state and leases table.
+                        if (changes > 0)
+                        {
+                            this._logger.LogWarning("There are changes in the change-tracking table for the user table, but the global state and leases table are not created. Spinning up a worker to create those tables.");
```
Charles-Gagnon, line 74:
```suggestion
                            this._logger.LogWarning("There are changes in the change-tracking table for the user table, but the global state and leases table are not created. Spinning up a worker to create those tables and start processing changes.");
```

src/TriggerBinding/SqlTriggerUtils.cs
```diff
@@ -126,5 +126,35 @@ internal static string GetBracketedLeasesTableName(string userDefinedLeasesTable
...
+        /// Returns the number of changes in the change tracking table of ID of the user table.
+        /// </summary>
+        /// <param name="connectionString">SQL connection string used to connect to user database</param>
+        /// <param name="userTable">SqlObject user table</param>
+        /// <param name="logger">Facilitates logging of messages</param>
+        /// <param name="cancellationToken">Cancellation token to pass to the command</param>
+        /// <exception cref="InvalidOperationException">Thrown in case of error when querying the object ID for the user table</exception>
+        internal static async Task<int> GetChangeCountFromChangeTrackingAsync(string connectionString, SqlObject userTable, ILogger logger, CancellationToken cancellationToken)
+        {
+            string getChangeCountCommand = $"SELECT COUNT_BIG(*) FROM CHANGETABLE(CHANGES {userTable.FullName}, null) AS ChTbl;";
```
Charles-Gagnon, line 140:
```suggestion
            string getChangeCountCommand = $"SELECT COUNT_BIG(*) FROM CHANGETABLE(CHANGES {userTable.BracketQuotedFullName}, null) AS ChTbl;";
```

MaddyDev: how will the BracketQuotedFullName look? Like ex: `[dbo].[Products]`

Charles-Gagnon: We need to bracket (or single quote depending on where it's used) ALL user object names, since otherwise we leave this open to a SQL injection attack (and it just plain won't work for tables with spaces in them) otherwise

And yes - it'll bracket quote the schema + table name. https://github.com/Azure/azure-functions-sql-extension/blob/35ad03153a1e557d3a61f4a938b551ca4f083fb7/src/SqlObject.cs#L77C131-L77C152

Charles-Gagnon (approved):

PR #1221, opened 2026-04-07
Address Semaphore leak in SQL trigger renew leases sync

VasuBhog:
# Description

## Problem

When a SQL trigger's lease renewal connection breaks (transient SQL availability issue), `RenewLeasesAsync` can permanently leak the `_rowsToProcessLock` semaphore. `BeginTransaction()` throws after the semaphore is acquired, but `Release()` is not in a `finally` block — so it's never called. Both the change consumption loop and lease renewal loop then deadlock permanently on the next semaphore acquisition. The trigger silently stops processing with no error logs, while the host appears healthy.

**Customer impact:** Trigger on table `ReceiptImport` was dead for ~3.5 days (ICM 51000000966606). Only a host restart resolved it.

## Root Cause

In `RenewLeasesAsync`, `_rowsToProcessLock.Release()` was placed after the `if` block rather than in a `finally` — meaning any exception between `WaitAsync` and `Release` (e.g., `BeginTransaction` throwing on a broken connection) permanently leaks the semaphore.

The same pattern existed in 4 other sites: `GetTableChangesAsync` (2 sites), `ProcessTableChangesAsync`, and `ClearRowsAsync`. Only `ProcessChanges` correctly used `try/finally`.

## Fix

Wrap all `_rowsToProcessLock` acquire/release pairs in `try/finally`:

- `RenewLeasesAsync` — **primary fix** (the site that caused the customer incident)
- `GetTableChangesAsync` — commit path and error-handling path
- `ProcessTableChangesAsync` — success path
- `ClearRowsAsync` — state reset

No behavioral changes — just ensures the semaphore is always released.

In addition, go through the checklist below and check each item as you validate it is either handled or not applicable to this change.

# Code Changes

- [ ] [Unit tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Unit) are added, if possible
- [ ] [Integration tests](https://github.com/Azure/azure-functions-sql-extension/tree/main/test/Integration)  are added if the change is modifying existing behavior of one or more of the bindings

# Dependencies

# Documentation

- [ ] Add [samples](https://github.com/Azure/azure-functions-sql-extension/tree/main/samples) if the change is modifying or adding functionality
- [ ] Update relevant documentation in the [docs](https://github.com/Azure/azure-functions-sql-extension/tree/main/docs)

### DIFF
--- a/src/TriggerBinding/SqlTableChangeMonitor.cs
+++ b/src/TriggerBinding/SqlTableChangeMonitor.cs
@@ -362,9 +362,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.Sql
 
                         // Set the rows for processing, now since the leases are acquired.
                         await this._rowsToProcessLock.WaitAsync(token);
-                        this._rowsToProcess = rows;
-                        this._state = State.ProcessingChanges;
-                        this._rowsToProcessLock.Release();
+                        try
+                        {
+                            this._rowsToProcess = rows;
+                            this._state = State.ProcessingChanges;
+                        }
+                        finally
+                        {
+                            this._rowsToProcessLock.Release();
+                        }
                     }
                     catch (Exception)
                     {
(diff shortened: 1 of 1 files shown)


src/TriggerBinding/SqlTableChangeMonitor.cs
```diff
@@ -527,84 +545,87 @@ private async void RunLeaseRenewalLoopAsync()
...
-                    {
-                        // Do we want to update this count even in the case of a failure to renew the leases? Probably,
-                        // because the count is simply meant to indicate how much time the other thread has spent processing
-                        // changes essentially.
-                        this._leaseRenewalCount += 1;
-
-                        // If this thread has been cancelled, then the _cancellationTokenSourceExecutor could have already
-                        // been disposed so shouldn't cancel it.
-                        if (this._leaseRenewalCount == MaxLeaseRenewalCount && !token.IsCancellationRequested)
+                        finally
```
Charles-Gagnon, line 602: We should add a catch here that logs the exception trying to begin the transaction

Charles-Gagnon (changes requested):

Charles-Gagnon (approved):

chlafreniere (approved):

